PR #10, opened 2026-04-21
feat: basic request tracing with jaeger ui for dev

illegalprime:
basic visualization with jaeger for request timings, nothing deeper at the moment:

Screenshot 2026-04-21 at 2 01 17 PM


server/cmd/fleetd/main.go
```diff
@@ -412,7 +424,7 @@ func start(config *Config) error {
 		}()
 	}
 
-	var handler http.Handler = mux
+	var handler http.Handler = interceptors.TracingMiddleware(mux)
 	for _, m := range middlewares {
 		handler = m.Wrap(handler)
```
ankitgoswami, line 430: isn't this excluding the work done in the middleware? We should wrap the handler with TracingMiddleware after this loop.

ankitgoswami (dismissed):

ankitgoswami:
maybe rename tracing to observability since we'd also want to use some of these same configs for metrics.

server/internal/infrastructure/fleet-telemetry/telemetry.go
```diff
@@ -0,0 +1,66 @@
...
+		return func(context.Context) error { return nil }, nil
+	}
+
+	exporter, err := otlptracehttp.New(ctx, otlptracehttp.WithEndpointURL(cfg.Endpoint))
+	if err != nil {
+		return nil, fmt.Errorf("create OTLP HTTP exporter: %w", err)
+	}
+
+	res, err := resource.New(ctx,
+		resource.WithAttributes(semconv.ServiceName(cfg.ServiceName)),
```
ankitgoswami, line 37: we probably also want the version of fleet here. can change in a followup Pr though

illegalprime: do you know a good way to get the current fleet version?

server/cmd/fleetd/main.go
illegalprime: It's hard to say weather we want the middleware or not. Sometimes middleware has big performance hits that we'd want to include in metrics (DB lookups). Other times, like the single middleware that exists now, it's just a CORS wrapper. I only worry that placing it after the loop would make the request spans brittle (less consistent with real timings) if someone adds heavy middleware in the future.

ankitgoswami: but request times should include any processing happening in the middleware right to more realistically represent client-visible time. I think if we care about certain hot code paths and want to isolate those then we can use sub-spans.
A span that grows when someone adds a heavy middleware would be a good signal to have.

server/internal/infrastructure/fleet-telemetry/telemetry.go
ankitgoswami: won't be too hard but we'd have to modify our release job to pass in some flags to the go build

server/cmd/fleetd/main.go
illegalprime: Ah I see, I got that backwards.

server/internal/infrastructure/fleet-telemetry/telemetry.go
ankitgoswami: this should work: https://github.com/block/proto-fleet/pull/28/changes

ankitgoswami (approved):

PR #16, opened 2026-04-21
chore: simplify tooltip for single-series line charts

negarn:
## Summary
- hide the aggregate label and status dot for single-series ProtoFleet line chart tooltips while keeping the shared tooltip behavior configurable
- add focused coverage for the shared tooltip and the ProtoFleet wrapper so the aggregate-only behavior stays explicit

## Test plan
- [ ] Open a ProtoFleet chart with a single aggregate series and confirm the tooltip shows only the value
- [ ] Run the shared tooltip and ProtoFleet LineChart Vitest coverage
- [ ] Run both client builds


flesher (approved):
lgtm

PR #17, opened 2026-04-21
feat(uptime): 3-state uptime chart

### LINKED ISSUE #12: Bug: Uptime chart "not hashing" miner count is inaccurate
> _Migrated from btc-mining/proto-fleet#1588_

## Description

The Uptime chart legend shows an incorrect count for "Not hashing" miners. Clicking through to filter by that segment shows a different (lower) number of miners, and the miner(s) shown may actually be hashing.

## Steps to Reproduce

1. Open the fleet dashboard and view the Uptime chart
2. Observe the "Not hashing" legend entry — it shows **2 miners** (14% of fleet)
3. Click the "Not hashing" segment to filter the miner table
4. Only **1 miner** appears in the filtered view, and it has a hashrate of 16.1 TH/s (i.e. it *is* hashing)

## Expected Behavior

The "Not hashing" count in the legend should match the number of miners returned when filtering by that state, and those miners should genuinely not be hashing.

## Actual Behavior

- Legend reports 4 miners not hashing
- Filter returns only 1 miner
- That miner has a non-zero hashrate ("Needs attention" status, 16.1 TH/s) — it is hashing

## Screenshots

**Uptime chart showing 4 miners not hashing:**

Image

> Legend: "Not hashing — 14% of fleet — 2 miners". ALSO - 2 miners is not 14%

**Miner table after clicking "Not hashing":**
> Only 1 miner shown (M60SVL20, PSU failure, 16.1 TH/s, Needs attention)

ankitgoswami:
image

## Summary

- New `miner_state_snapshots` TimescaleDB hypertable storing **one row per paired device per tick** with a 4-state code (`offline/sleeping/broken/hashing`). Classifier CASE mirrors `CountMinersByState`, so the chart and the live FleetHealth legend agree.
- Uptime chart rendered as **3 segments**: Hashing / Needs attention / Not hashing. Drill-through buttons deep-link to the matching miner-list filters via the existing `encodeFilterToURL` helper.
- Writer is a single `INSERT ... SELECT` per 60 s tick — one round-trip, no per-org loop, no Go-side array packing.
- Read query (`GetMinerStateSnapshots`) does `DISTINCT ON (bucket, device_identifier)` + SUM-by-state and applies the device filter at the CTE level, so `uptime_status_counts` honors the full `device_selector` (fleet, group, rack, arbitrary list). Drops in cleanly for future group/rack overview pages without server changes.

fixes https://github.com/block/proto-fleet/issues/12

## Test plan

- [x] `go build`, `go vet`, `golangci-lint run` clean on `server/internal/...` and `server/cmd/...`
- [x] `go test ./internal/domain/telemetry/... ./internal/handlers/telemetry/...` green
- [x] Unit tests for `writeFleetStateSnapshot` (happy path + log-on-error) and lifecycle tests (`Start`/`Stop`) cover the tick fires
- [x] `cd client && npx tsc --noEmit`, `npm run lint`, `vitest run src/protoFleet/features/dashboard` green
- [x] Manual: bring up local stack, verify `miner_state_snapshots` grows by ~fleet-size rows/min, open dashboard, confirm all 3 segments render, confirm drill-throughs land on the right filter URLs, confirm totals match FleetHealth legend


client/src/protoFleet/features/dashboard/components/UptimePanel/UptimePanel.tsx
```diff
@@ -1,13 +1,33 @@
...
+      DeviceStatus.NEEDS_MINING_POOL,
+      DeviceStatus.UPDATING,
+      DeviceStatus.REBOOT_REQUIRED,
+    ],
+  }),
+).toString()}`;
+
+const notHashingRoute = `/miners?${encodeFilterToURL(
+  create(MinerListFilterSchema, {
+    deviceStatus: [DeviceStatus.OFFLINE, DeviceStatus.INACTIVE],
```
flesher: on the dashboard notHashing is offline + sleeping. The queries infer sleeping by MAINTENANCE | INACTIVE.  The drill down link here does not account for MAINTENANCE.  Not sure when miners actually return a deviceStatus of MAINTENANCE, but theoretically the count of mines on the filtered list page could omit miners reflected in the uptimePanel

flesher (approved):
One small thing I noticed, otherwise looks good

client/src/protoFleet/features/dashboard/components/UptimePanel/UptimePanel.tsx
ankitgoswami: good catch, that's not used right now but should definitely be handled

PR #37, opened 2026-04-22
fix: speed up GetLatestDeviceMetrics for large groups (#11)

### LINKED ISSUE #11: Group stats timeout with 10,000 miners
## Summary

`GetDeviceSetStats` times out when loading group-level performance aggregates (hashrate, temperature, efficiency, power) for a group containing 10,000 miners.

## Steps to Reproduce

1. Configure virtual plugin with 10,000 miners
2. Navigate to the "all miners" group overview page
3. Observe that the Performance section charts (Hashrate, Temperature, Avg Efficiency, Total Power) and status summary cards fail to load

## Error

```
fleet-api-1  | time=2026-04-17T19:47:19.834Z level=ERROR msg="incoming unary request failed" procedure=/device_set.v1.DeviceSetService/GetDeviceSetStats took=1m0.032350902s error="FleetError: internal (Common: 0) failed to fetch telemetry: failed to query latest metrics: timeout: context deadline exceeded"
```

The query takes >60s and hits the context deadline.

## Expected Behavior

Group stats should load within a reasonable time regardless of group size.

## Screenshots

Group stats failing to load

## Notes

- Works fine at 1,000 miners
- The telemetry query for latest metrics is the bottleneck — likely needs query optimization or pre-aggregation for large groups

ankitgoswami:
## Summary

- Fixes #11 — `GetDeviceSetStats` timing out at >60s on a group of 10,000 miners.
- Rewrites `GetLatestDeviceMetrics` from `DISTINCT ON` + `WHERE device_identifier = ANY($array)` to `unnest($array)` + `CROSS JOIN LATERAL (... LIMIT 1)`, turning a scan-and-sort of every metrics row in the 24h window into one index seek per device on the existing `idx_device_metrics_device_identifier(device_identifier, time DESC)` index.
- No schema, migration, Go consumer, or generated-type-signature changes — the param struct, return type, and `Scan` order are unchanged; only the SQL shape is.

gotta go fast

## Why the old query was slow

The previous query shape forced PostgreSQL to read every row in the 24h `maxAge` window for every device in the array and sort the whole set by `(device_identifier, time DESC)` before `DISTINCT ON` could pick one row per device. With 10k miners reporting every ~10s, that's tens of millions of rows per request. With 1k miners it fits under the 60s query timeout; with 10k it doesn't.

The `ANY(array)` + `DISTINCT ON` pattern can't be rewritten into a skip-scan by the planner, even with the ideal `(device_identifier, time DESC)` index already in place. Lateralizing the lookup forces one index seek per device, which is exactly what that index was built for.

## Test plan

- [x] `go build ./...` clean.
- [x] `TestService_GetCollectionStats_*` (full `GetCollectionStats` suite, 9 tests) pass.
- [x] `TestTelemetryStore_GetLatestDeviceMetricsBatch` passes against a real Postgres/Timescale container, confirming the rewritten SQL still returns the correct latest-per-device rows within `maxAge`.
- [x] End-to-end repro against the virtual plugin: set `plugin/virtual/config.json` `count: 10000`, pair all miners, create an all-miners group, open the group overview. Expect Performance charts + status cards to populate in seconds; server log should show `GetDeviceSetStats` returning quickly rather than erroring out at 60s.
- [ ] Optional: `EXPLAIN (ANALYZE, BUFFERS)` on the new query against a populated `device_metrics` table should show `Nested Loop` with `Index Scan using idx_device_metrics_device_identifier` + `Limit 1` on the inner side.


mcharles-square (approved):
neat

PR #44, opened 2026-04-22
ci(e2e): share admin auth via Playwright setup project [2/3]

ankitgoswami:
## Summary
Part 2 of a 3-PR CI speedup series. Replaces the per-shard shell loop over setup specs with Playwright's native project `dependencies` + `storageState`, so most target tests start already authenticated instead of running the login flow in every single `it()`.

## What changed
- **Two viewport-matched setup projects** in [playwright.config.ts](client/e2eTests/protoFleet/playwright.config.ts): `setup-desktop` and `setup-mobile` each run `spec/00-*`, `spec/01-*`, and the new `02-saveAuthState.spec.ts` under their own viewport. `desktop` and `mobile` target projects declare `dependencies: ['setup-desktop']` / `dependencies: ['setup-mobile']` respectively.
- **New [02-saveAuthState.spec.ts](client/e2eTests/protoFleet/spec/02-saveAuthState.spec.ts)** signs in as admin once and writes `playwright/.auth/admin.json` (already gitignored; `fs.mkdir` ensures the directory exists on fresh checkouts).
- **`desktop` / `mobile`** projects load that file via `use.storageState`, so most target tests boot pre-authenticated.
- **`commonSteps.loginAsAdmin()` is idempotent** via a new `authPage.isAlreadyLoggedIn()` helper that waits for either the authed UI marker or the login form to appear, then returns `true` only when positively confirming the authed state. Only Playwright timeouts are swallowed; other errors propagate.
- **Specs that manipulate sessions opt out of `storageState`**: `auth.spec.ts` validates login itself; `teamAccounts.spec.ts` and `securitySettings.spec.ts` exercise `authPage.logout()` + re-login flows that would otherwise leave later tests holding a revoked cookie while the UI still shows authed chrome.
- **`globalSetup.ts`** enforces exactly one `--project` flag (throws with guidance otherwise). This keeps a bare `npx playwright test` or multi-project invocation from scheduling both setup projects against the same backend; the numbered setup specs are not idempotent on shared state. `--list` is exempted (dry-run, no backend).
- **CI workflow** drops the shell loop that ran setup specs one-by-one. Each shard now runs a single `npx playwright test --project=${{ matrix.project }}` call and Playwright handles setup ordering via `dependencies`.

## Expected impact
- ~30s saved per target test that skips the login flow. `minersSleepWake.spec.ts mobile` (critical path, 3 tests) should drop ~90s just from this.
- Plus one-time savings of 2 extra browser cold starts per shard (setup specs used to be separate Playwright invocations).
- Combined with #43 (prebuild artifacts), total estimated critical-path reduction is ~4-5 minutes.

## Test plan
- [ ] Local: `cd client/e2eTests/protoFleet && E2E_TARGET=fake npx playwright test --project=desktop spec/minersSleepWake.spec.ts` — confirm the setup project runs first, target spec skips login step, all tests pass.
- [ ] Local: `npx playwright test --project=desktop spec/auth.spec.ts` — confirm `auth.spec.ts` still sees the login form (storageState opted out).
- [ ] Local: `npx playwright test` (no project), `npx playwright test --project=desktop --project=mobile` — both should fail at globalSetup with the guidance message.
- [ ] CI: Open PR, confirm each shard does one Playwright invocation and the blob report still merges correctly.
- [ ] CI: Check `minersSleepWake.spec.ts mobile` runtime vs main.

Stacking note: #43 is merged; this PR targets `main` cleanly.

### DIFF
--- a/.github/workflows/protofleet-e2e-tests.yml
+++ b/.github/workflows/protofleet-e2e-tests.yml
@@ -148,8 +148,9 @@ jobs:
           rm -rf blob-report playwright-report test-results
           mkdir -p blob-report
 
-          # The Playwright `setup` project (onboarding + pools + auth storageState)
-          # runs automatically before the target via project `dependencies`.
+          # The viewport-matched setup project (setup-${{ matrix.project }}:
+          # onboarding + pools + auth storageState) runs automatically before
+          # the target via project `dependencies`.
           PLAYWRIGHT_BLOB_OUTPUT_FILE="blob-report/${{ matrix.project }}-${{ steps.artifact.outputs.spec_name }}.zip" \
           PWTEST_BLOB_DO_NOT_REMOVE=1 \
           npx playwright test --project=${{ matrix.project }} --reporter=blob "spec/${{ matrix.spec }}"
--- a/.github/workflows/protofleet-e2e-tests.yml
+++ b/.github/workflows/protofleet-e2e-tests.yml
@@ -144,32 +144,15 @@ jobs:
         working-directory: client/e2eTests/protoFleet
         env:
           CI: true
-          SETUP_SPECS_JSON: ${{ needs.detect-specs.outputs.setup_specs }}
         run: |
-          run_spec() {
-            local spec_path="$1"
-            local spec_slug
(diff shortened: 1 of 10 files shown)


flesher (approved):
Nice!

PR #48, opened 2026-04-22
ci(e2e): prebuild+cache docker images and fix node cache collision

ankitgoswami:
## Summary

Two related E2E CI follow-ups to [#43](https://github.com/block/proto-fleet/pull/43):

### 1. Prebuild and cache docker-compose images

Wave 1 moved the client and plugin builds into a shared job, but each e2e shard was still running `docker compose up -d` against raw build contexts \u2014 rebuilding `fleet-api`, `timescaledb`, 6\u00d7 `fake-antminer`, and 2\u00d7 `fake-proto-rig` from scratch every time. On shards that lost the Docker layer-cache lottery that added 3\u20135 min (the `teamAccounts mobile` shard on [#43](https://github.com/block/proto-fleet/pull/43) spent 3m 5s compiling `fake-antminer-hw-errors` alone).

Moved into the `build` job:

- Restore a cross-run image tarball via `actions/cache`, keyed on every input that changes image contents (`Dockerfile`, `go.sum`/`go.mod`, `server/**/*.go`, fake-miner contexts, `timescaledb/**`, compose files).
- On cache miss, `docker compose build` \u2192 enumerate images via `docker compose config --images` \u2192 `docker save | gzip` to `/tmp/docker-images/e2e-images.tar.gz`.
- Upload as the `protofleet-e2e-docker-images` artifact (retention 1 day).

Each shard now:

- Downloads the tarball, `docker load`s it.
- Runs `docker compose up -d --no-build`. The `--no-build` is intentional \u2014 if tags ever mismatch we'd rather see a loud failure than a silent 3\u20135 min rebuild.

Expected impact: warm-cache runs, the `build` job restores the tarball and skips compose build entirely; cold-cache runs pay one compose build (in `build`) instead of ~32 in parallel shards.

### 2. Non-Playwright node cache jobs become restore-only

Fixes a cache-key collision introduced in [#43](https://github.com/block/proto-fleet/pull/43). The `build` and `merge-reports` jobs shared the cache key `${{ runner.os }}-node-${{ hashFiles('client/package-lock.json') }}` with the E2E shard but cached a smaller path set. Cache entries are **immutable per key**, so whichever job saved first won; if `build` or `merge-reports` saved first, shards could never upload a cache that included `~/.cache/ms-playwright` \u2014 Playwright browsers got re-downloaded on every shard run.

Switched the two non-Playwright jobs to `actions/cache/restore` (restore-only). The E2E shard remains the sole writer and keeps saving the full `client/node_modules + ~/.cache/ms-playwright` pair.

stay in your lane

## Test plan

**Image cache**
- [ ] First CI run after this merges: confirm the `build` job logs `docker compose build` (cache miss) and uploads `protofleet-e2e-docker-images`.
- [ ] Spot-check one shard: it should log `Load Docker images` (no build) and `docker compose up -d --no-build` should succeed.
- [ ] Subsequent run with no server/Dockerfile/compose changes: confirm the `build` job logs `Cache restored from key: ...` and skips the build step entirely.
- [ ] Compare slowest-shard duration vs a recent main run; previously the teamAccounts-mobile outlier was 12m 35s, expected closer to ~7\u20138 min.
- [ ] Confirm `--no-build` surfaces a clear error if the artifact/tarball is ever missing (i.e. no silent fallback to per-shard rebuild).

**Node cache restore-only**
- [ ] Confirm the `build` job logs \"Cache restored from key: ...\" but does **not** log a save action.
- [ ] Confirm the `e2e-tests` shard job still saves both paths (`client/node_modules` and `~/.cache/ms-playwright`).
- [ ] On a subsequent PR run, confirm the `build` job restores the cache saved by a prior shard.

\ud83e\udd16 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/protofleet-e2e-tests.yml
+++ b/.github/workflows/protofleet-e2e-tests.yml
@@ -109,6 +109,12 @@ jobs:
         working-directory: server
         run: |
           mkdir -p /tmp/docker-images
+          # `--ignore-buildable` pulls only services with `image:` and no
+          # `build:` (otel-collector, jaeger). `docker save` below otherwise
+          # fails on a cold runner because those images don't exist locally.
+          echo "::group::docker compose pull"
+          docker compose pull --ignore-buildable
+          echo "::endgroup::"
           echo "::group::docker compose build"
           docker compose build
           echo "::endgroup::"
--- a/.github/workflows/protofleet-e2e-tests.yml
+++ b/.github/workflows/protofleet-e2e-tests.yml
@@ -55,8 +55,12 @@ jobs:
       - name: Configure Go cache
         uses: ./.github/actions/go-cache-setup
 
-      - name: Cache Node modules
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      # Restore-only: the E2E shard job saves this key with an additional
+      # `~/.cache/ms-playwright` path. Cache keys are immutable, so if this job
+      # saved first it would permanently lock the shards out of caching their
+      # Playwright browsers. Restore here, save there.
+      - name: Restore Node modules cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: |
             client/node_modules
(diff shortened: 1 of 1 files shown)


flesher (approved):
Looks good! robots seem to have some consensus on using server/** to generate the cache key since ignored binaries live in that directory. Seems valid.

PR #64, opened 2026-04-23
Add Fleet activity E2E tests

edgars-avotins:
## What changed
- added a ProtoFleet Activity page object and fixture wiring
- added Activity E2E coverage for login activity, Blink LEDs activity, and schedule search/filter behavior
- reused existing Fleet helpers and filtered Blink LEDs to Proto rig miners for stable supported-action coverage

## Why
The Activity screen did not have end-to-end coverage for core user-visible audit flows after the repo move.

## Validation
- `npx eslint e2eTests/protoFleet/spec/activity.spec.ts`
- `npx playwright test spec/activity.spec.ts --project=desktop --no-deps`
- `npx playwright test spec/activity.spec.ts --project=mobile --no-deps`

### DIFF
--- a/client/e2eTests/protoFleet/spec/activity.spec.ts
+++ b/client/e2eTests/protoFleet/spec/activity.spec.ts
@@ -1,0 +1,167 @@
+/* eslint-disable playwright/expect-expect */
+import { testConfig } from "../config/test.config";
+import { test } from "../fixtures/pageFixtures";
+import { CommonSteps } from "../helpers/commonSteps";
+import { generateRandomText } from "../helpers/testDataHelper";
+import { AuthPage } from "../pages/auth";
+import { MinersPage } from "../pages/miners";
+import { SettingsSchedulesPage } from "../pages/settingsSchedules";
+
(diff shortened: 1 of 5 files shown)


client/e2eTests/protoFleet/spec/activity.spec.ts
```diff
@@ -0,0 +1,167 @@
+/* eslint-disable playwright/expect-expect */
```
mcharles-square, line 1: If the pattern for these e2e specs is to not have asserts directly in the steps it could be worth disabling in the eslint conf i.e.:

```
...
overrides: [
    {
      files: ["e2eTests/**/*.spec.ts"],
      rules: {
        "playwright/expect-expect": "off"
      }
    }
  ]
...
```

mcharles-square (approved):

client/e2eTests/protoFleet/spec/activity.spec.ts
edgars-avotins: I'll try that in next one, don't want to go through all checks again

Issue #22, opened 2026-04-21
Activity log should capture per-miner success/failure detail for bulk actions


rongxin-liu:
## Summary

When a bulk action partially fails, there is no record in the activity log of which miners succeeded and which failed. The activity log should capture this detail so operators have a historical record of bulk action outcomes.

## Current behavior

Bulk actions appear in the activity log as a single entry (e.g. "Reboot — 14 miners") with no breakdown of success vs. failure.

## Expected behavior

The activity log should record:
- Which miners succeeded and which failed
- The failure reason per miner (if available)
- The overall success/failure count

This gives operators an audit trail for diagnosing recurring issues with specific miners.

PR #71, opened 2026-04-24
patch(activity backend): record device name/IP/MAC on per-miner bulk action detail

rongxin-liu:
## Problem

PR #25 shipped the per-miner batch detail RPC, but each row returns only the device UUID — a value the app surfaces nowhere else. Operators auditing a past bulk action have no way to recognize which miners were affected without cross-referencing UUIDs against another system. Frontend PR #60 renders whatever identifiers the backend hands it; today that's just the UUID.

## What ships

Backend only. Follows the write-time capture convention already used by `ActivityEntry.scope_label`. Closes nothing; companion to #22 / #60.

- **Record raw device-identity fields at first write.** Five new nullable columns on `command_on_device_log` (`custom_name`, `manufacturer`, `model`, `ip_address`, `mac_address`), populated by `UpsertCommandOnDeviceLog` joining `device` + `discovered_device`. The read path composes the display name via `fleetmanagement.ComposeDeviceName` — the same helper the live fleet read path uses, so the audit log and live UI never disagree on what a miner is called.
- **Immutable after first write.** `ON CONFLICT` leaves the captured columns untouched so retries and the reaper-after-success path preserve the original audit values — even if the device was renamed or moved IPs between the first and second Upsert.
- **Optional proto fields with UUID fallback.** `CommandBatchDeviceResult.{device_name, ip_address, mac_address}` are `optional string`; historical rows (pre-migration) surface as unset so the frontend can fall back to UUID.

### Schema

- `server/migrations/000037_add_device_snapshot_to_command_on_device_log.up.sql` — five nullable `TEXT` columns. Nullable so historical rows stay NULL and the frontend can detect them.

### SQL

- `server/sqlc/queries/command.sql` — `UpsertCommandOnDeviceLog` has a `dev` CTE that copies `device.custom_name` / `device.mac_address` / `discovered_device.{manufacturer, model, ip_address}` verbatim; no composition or rendering happens in SQL. `ON CONFLICT DO UPDATE` intentionally excludes the captured columns. `ListBatchDeviceResults` selects the five new columns.

### Shared helper

- `server/internal/domain/fleetmanagement/device_name.go` — `ComposeDeviceName(customName, manufacturer, model string) string` is the single source of truth for the display-name rule. Both `fleetmanagement/service.go` (live fleet read) and `command/service.go` (audit log read) call it; the live-fleet call site was refactored to use it too so there's nowhere for the rule to drift.

### Proto

- `proto/minercommand/v1/command.proto` — three `optional string` fields on `CommandBatchDeviceResult` (tags 5/6/7). `optional` is deliberate so the frontend can distinguish "never captured" (historical) from "captured empty." The field doc describes the consumer contract, not the composition recipe.

### Service

- `server/internal/domain/command/service.go` — mapper reads the five raw sqlc fields, composes `device_name` via `fleetmanagement.ComposeDeviceName`, and passes `ip_address` / `mac_address` through using the existing `ErrorMessage` NullString-to-pointer pattern.

### Tests

- `server/internal/domain/fleetmanagement/device_name_test.go` — table-driven unit test for `ComposeDeviceName` covering the five relevant cases (custom set, custom empty with both mfr/model, only mfr, only model, all empty).
- `server/internal/domain/command/results_integration_test.go` — `TestGetCommandBatchDeviceResults_DeviceSnapshot` covers the full invariant: first Upsert captures the fields; a rename + IP change; second Upsert (reaper-style) flips status/error_info but leaves the captured values alone. `TestGetCommandBatchDeviceResults_HistoricalRowsOmitSnapshot` covers backward compatibility by inserting a codl row directly with NULL captured columns and asserting the proto optionals surface as unset.

## Test plan

- [ ] `cd server && just test` passes (the two new integration tests run against the TimescaleDB container and should both pass)
- [ ] Run a bulk action (e.g. "Blink LEDs") on a rack. Expand the batch in the activity log detail view (via PR #60) and confirm name / IP / MAC appear alongside each device row
- [ ] Rename a miner (`/miners/:id` → rename) and expand the same batch again — **original name still shown** (audit invariant)
- [ ] Seed a pre-migration row manually (`INSERT INTO command_on_device_log ...` without captured columns) — verify the RPC returns unset optionals and the frontend falls back to UUID
- [ ] Kill a worker mid-command so the reaper fires; confirm the reaped row still has captured values (the reap path writes through the same `UpsertCommandOnDeviceLog`)


proto/minercommand/v1/command.proto
```diff
@@ -218,6 +218,15 @@ message CommandBatchDeviceResult {
   optional string error_message = 3;
   // When this row was last updated (start of this attempt's terminal write).
   google.protobuf.Timestamp updated_at = 4;
+  // Human-readable device name captured at first write (custom_name, else
+  // manufacturer + " " + model). Unset for historical rows written before
```
mcharles-square: comment leaks implementation details with `(custom_name, else manufacturer + " " + model).`

server/sqlc/queries/command.sql
```diff
@@ -45,23 +45,47 @@ WHERE uuid = $1;
...
+-- fleetmanagement.service.go (custom_name, else manufacturer + " " + model).
 WITH batch AS (
     SELECT id FROM command_batch_log WHERE uuid = $4
+),
+dev AS (
+    SELECT
+        COALESCE(
+            NULLIF(d.custom_name, ''),
+            NULLIF(CONCAT_WS(' ', dd.manufacturer, dd.model), '')
+        ) AS device_name,
```
mcharles-square: this feels likely to drift

server/sqlc/queries/command.sql
```diff
@@ -45,23 +45,47 @@ WHERE uuid = $1;
...
   $1,
   $2,
   $3,
-  $5
-FROM batch
+  $5,
+  dev.device_name,
+  dev.ip_address,
+  dev.mac_address
+FROM batch, dev
```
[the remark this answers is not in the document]
rongxin-liu: Good catch on the behavior change. In practice `dev` is guaranteed non-empty because:

- `fk_command_on_device_log_device` requires `device_id` to exist, with no cascading delete — you can't hard-delete a device that has codl rows referencing it.
- `device.discovered_device_id` is `NOT NULL` with an FK to `discovered_device`, so every device has a joinable discovered_device row.
- `queue_message.device_id` has the same FK, so the worker and reaper paths always pass valid IDs.

Commit dd0630c adds an explicit comment above `FROM batch, dev` documenting this invariant. That said, the `LEFT JOIN` approach you suggested would preserve the FK-violation error surface and isn't much more complex — happy to switch if the safety margin feels worth it here.

server/sqlc/queries/command.sql
```diff
@@ -45,23 +45,47 @@ WHERE uuid = $1;
 -- PostgreSQL version using CTE for the subquery.
 -- error_info is NULL for SUCCESS rows; for FAILED rows it is either the worker
 -- error string (truncated by the caller) or the reaper reason.
+--
+-- device_name/ip_address/mac_address are snapshotted from device + discovered_device
+-- at first write for audit-log display, and deliberately left untouched by the
+-- ON CONFLICT branch so that retries and the reaper never overwrite the
+-- original values. device_name mirrors the runtime composition in
+-- fleetmanagement.service.go (custom_name, else manufacturer + " " + model).
```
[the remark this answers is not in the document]
rongxin-liu: Good catch on the incomplete path. Commit dd0630c takes a different approach to the same underlying concern: the composition rule is now out of SQL entirely and lives only in the Go helper `fleetmanagement.ComposeDeviceName`, which both the live-fleet read path and the command audit read path call. The updated SQL comment points at that helper rather than an inline recipe, so the drift surface is gone.

server/sqlc/queries/command.sql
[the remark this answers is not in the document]
rongxin-liu: Same concern as the Copilot review comment above — see the reply there for the full response. Short version: the FK constraints on both `command_on_device_log.device_id` and `queue_message.device_id` make an unresolvable device_id unreachable in the worker and reaper paths, and commit dd0630c documents that invariant in the SQL. Open to switching to a `LEFT JOIN` pattern if preserving the loud-fail-on-invalid-ID surface is worth the extra lines.

proto/minercommand/v1/command.proto
rongxin-liu: Fixed in dd0630c.

server/sqlc/queries/command.sql
rongxin-liu: The SQL composition rule would indeed be out of sync with the Go rule, so I addressed that in dd0630c.

And just to clarify, the device name recorded in the log would still show what the miner was called when the action ran, not its current name.

mcharles-square (approved):

PR #77, opened 2026-04-24
perf(client): split useBatchOperations into actions-only hook (#32)

### LINKED ISSUE #32: Split useBatchOperations into actions-only and versioned hooks to reduce re-renders
## Context

`useBatchOperations()` subscribes to `useBatchStateVersion()`, so every consumer re-renders on every batch mutation — even components that only need action callbacks (like `SingleMinerActionsMenu` which renders per-row).

## Impact

Low — the re-renders are cheap and only happen on batch state changes, not polling. But with many rows visible, it's O(row-count) unnecessary renders per batch operation.

## Proposed fix

Split into two hooks:
- `useBatchActions()` — returns only `startBatchOperation`, `completeBatchOperation`, `removeDevicesFromBatch`, `cleanupStaleBatches` (no store subscription)
- `useBatchOperations()` — adds `batchStateVersion`, `getActiveBatches`, `getAllBatches` (subscribes to version)

Components like `SingleMinerActionsMenu` and `MinerActionsMenu` would use `useBatchActions()`. Only `Fleet.tsx` and `MinerList` need the full `useBatchOperations()`.

Hook lives at `client/src/protoFleet/features/fleetManagement/hooks/useBatchOperations.ts`.

_Migrated from btc-mining/proto-fleet#1534 (originally surfaced by Codex review on #1523)_

flesher:
## Summary

Closes #32. Adds `useBatchActions()` that returns only the batch action callbacks without subscribing to `batchStateVersion`. Per-row consumers (`SingleMinerActionsMenu`, `MinerActionsMenu`, `BulkWorkerNameModal`, `DeviceSetActionsMenu`) now avoid re-rendering on every batch mutation. `Fleet.tsx` keeps the full `useBatchOperations` since it needs `batchStateVersion` to reconcile batch state against miner state.

## Test plan
- [x] `bin/just lint` clean
- [x] `npx tsc --noEmit` clean
- [x] Related vitest suites pass (113 tests)

### DIFF
--- a/client/e2eTests/protoFleet/pages/miners.ts
+++ b/client/e2eTests/protoFleet/pages/miners.ts
@@ -36,7 +36,9 @@ export class MinersPage extends BasePage {
 
   async validateMinersAdded(minerCount: number = 5) {
     const rows = this.page.getByTestId("list-body").locator("tr");
-    expect(await rows.count()).toBeGreaterThanOrEqual(minerCount);
+    await expect
+      .poll(() => rows.count(), { timeout: PROLONGED_TIMEOUT, intervals: [DEFAULT_INTERVAL] })
+      .toBeGreaterThanOrEqual(minerCount);
   }
 
   private async filterMinersByModel(minerType: string) {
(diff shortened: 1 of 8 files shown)


rongxin-liu (approved):
The JSDoc comments on the `UseMinerActionsParams` interface (lines 94, 96, 98 in `useMinerActions.tsx`) still say (`from useBatchOperations`). Since callers now pass these via `useBatchActions`, worth updating to (`from useBatchActions`)?

Or just dropping the parenthetical since `useMinerActions` receives these as props and doesn't care which hook provided them.

flesher:
ahh i guess auto-merge doesn't get blocked by non inline comment... good to know.

@rongxin-liu good callout, will file follow up

PR #80, opened 2026-04-24
ci: speed up contract tests

ankitgoswami:
## Summary

The **Contract Tests** job consistently takes ~13–14 min (e.g. 13m42s on [run 24802916993](https://github.com/block/proto-fleet/actions/runs/24802916993)) because every container in `just test-contract` re-downloads modules and recompiles from scratch — the host `actions/cache` restore is never plumbed into the containers. This PR pipes the cache through, collapses four cold Go environments into one, and caches the asicrs Rust build layers.

speed

### Changes
- **`test-contract`**: mount host `GOCACHE`/`GOMODCACHE` into every `docker run` and run as the host UID so the `actions/cache` restore is actually reused and written files stay runner-owned.
- **`test-contract`**: compile the contract-test binary once with `go test -c`, then execute it for each suite — removes three redundant compile passes. Suites still run sequentially because the asicrs harness rewrites `server/plugins/asicrs-config.yaml` next to the binary and parallel containers bind-mounting `/work` would race on that file.
- **`_asicrs-build`**: add `--cache-from`/`--cache-to type=gha,scope=asicrs-native` when running in CI, so the Rust build layers are cached across runs.
- **workflow**: add `docker/setup-buildx-action` so the GHA cache backend has a container-based builder to talk to.
- **`.gitignore`**: ignore `tests/plugin-contract/bin/` (the compiled contract-test binary).

## Test plan

- [ ] First CI run primes the caches — expect roughly baseline (~13 min) duration and confirm it completes.
- [ ] Second CI run on this branch: compare job duration against the ~13–14 min baseline; expect a meaningful drop (target: several minutes).
- [ ] Inspect the `Build plugins and run contract tests` step log: no `go: downloading ...` lines inside containers on cache hits.
- [ ] All three suites still pass: `TestAntminerStock`, `TestAntminerVNish`, `TestWhatsMinerStock`.
- [ ] Local: `just test-contract` still works on a dev machine (macOS + Docker Desktop).

### DIFF
--- a/justfile
+++ b/justfile
@@ -76,30 +76,59 @@ test-contract: _asicrs-build
   GO_VERSION=$(grep '^go ' tests/plugin-contract/go.mod | awk '{print $2}')
   IMAGE="golang:${GO_VERSION}-alpine"
 
-  # Build Go plugins once (shared via volume mount)
-  docker run --rm \
-    -v "$(pwd)":/work \
-    -w /work \
-    -e GOFLAGS=-buildvcs=false \
-    "$IMAGE" \
-    sh -c '\
-      mkdir -p server/plugins && \
-      (cd plugin/proto && go build -o ../../server/plugins/proto-plugin .) && \
-      (cd plugin/antminer && go build -o ../../server/plugins/antminer-plugin .) \
-    '
+  # Share the host Go build + module cache with every container so the
+  # actions/cache restore on CI (and the local dev cache) isn't wasted.
+  mkdir -p "${HOME}/.cache/go-build" "${HOME}/go/pkg/mod"
+  DOCKER_RUN=(
+    docker run --rm
+    -v "$(pwd):/work"
+    -v "${HOME}/.cache/go-build:/gocache"
+    -v "${HOME}/go/pkg/mod:/gomodcache"
+    --user "$(id -u):$(id -g)"
+    -e GOCACHE=/gocache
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -76,29 +76,45 @@ test-contract: _asicrs-build
...
+    -w /work
+  )
+
+  # Build Go plugins and compile the contract-test binary in a single container
+  # so the three test runs below only need to execute the binary, not recompile
+  # it. Only this builder container mounts the host Go caches, so the
+  # actions/cache restore on CI (and the local dev cache) isn't wasted; the
+  # test-execution containers below don't need it and shouldn't have write
+  # access to the host's global Go caches.
+  mkdir -p "${HOME}/.cache/go-build" "${HOME}/go/pkg/mod" tests/plugin-contract/bin
```
rongxin-liu, line 94: Worth considering adding

`rm -f tests/plugin-contract/bin/miners.test`

before the build step so that if the builder container fails, the test loop errors out immediately instead of silently running a stale binary from a previous run.

rongxin-liu (approved):

justfile
ankitgoswami: would be defense in depth but it's not really necessary: set -euo pipefail already aborts the script as soon as the builder container exits non-zero, so the test loop never sees a stale binary. The build step chains with &&, so any partial failure inside the container propagates out through docker run's exit code

PR #99, opened 2026-04-27
Switch to mobile first breakpoints rather than range based

### LINKED ISSUE #36: Switch to mobile first breakpoints rather than range based
- For historical reasons our client codebase uses range based breakpoints leading to a lot of extra classnames (ie. `laptop:p-0 desktop:p-0` vs. `laptop-p-0`)
- The default tailwind breakpoints are still functional and may have some usages in the codebase (`sm`, `md`, `lg`, `xl`), these breakpoints have different pixel values which could lead to design inconsistencies.  We should see if we can configure tailwind to disable these

cc @negarn

negarn:
## Summary
- Switch client breakpoint variants to mobile-first semantics for ProtoFleet, ProtoOS, and shared UI components
- Remove redundant range-based responsive utility pairs and disable unused default Tailwind screen variants
- Keep exact-range variants only where the layout still needs phone/tablet-specific behavior

## Test plan
- [ ] Verify ProtoFleet layouts at phone, tablet, laptop, and desktop widths
- [ ] Verify ProtoOS diagnostics, KPI, setup, and settings layouts across responsive widths
- [ ] Confirm action bars, toasts, filters, and tables do not overlap or regress on narrow screens

Closes #36

### DIFF
--- a/client/src/protoFleet/features/activity/pages/ActivityPage.tsx
+++ b/client/src/protoFleet/features/activity/pages/ActivityPage.tsx
@@ -59,17 +59,17 @@ const ActivityPage = () => {
   const { exportCsv, isExportingCsv } = useExportActivity();
   const { eventTypes, scopeTypes, users } = useActivityFilterOptions();
 
-  const [hasLoaded, setHasLoaded] = useState(false);
   const hasStartedLoadingRef = useRef(false);
+  const hasLoadedRef = useRef(false);
   useEffect(() => {
     if (isLoading) {
       hasStartedLoadingRef.current = true;
-    } else if (hasStartedLoadingRef.current && !hasLoaded) {
-      setHasLoaded(true);
+    } else if (hasStartedLoadingRef.current) {
+      hasLoadedRef.current = true;
     }
-  }, [isLoading, hasLoaded]);
+  }, [isLoading]);
 
-  const isInitialLoad = isLoading && activities.length === 0 && !hasLoaded;
+  const isInitialLoad = isLoading && activities.length === 0 && !hasLoadedRef.current;
   const isLoadingMore = isLoading && activities.length > 0;
 
   const hasActiveFilters =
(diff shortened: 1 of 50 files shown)


client/src/protoFleet/features/activity/pages/ActivityPage.tsx
```diff
@@ -59,17 +59,17 @@ const ActivityPage = () => {
...
-  const [hasLoaded, setHasLoaded] = useState(false);
   const hasStartedLoadingRef = useRef(false);
+  const hasLoadedRef = useRef(false);
   useEffect(() => {
     if (isLoading) {
       hasStartedLoadingRef.current = true;
-    } else if (hasStartedLoadingRef.current && !hasLoaded) {
-      setHasLoaded(true);
+    } else if (hasStartedLoadingRef.current) {
+      hasLoadedRef.current = true;
```
flesher, line 68: this change doesn't seem related to breakpoint refactor.  On purpose?

client/src/protoOS/hooks/useWakeMiner/useWakeMiner.ts
```diff
@@ -63,6 +63,18 @@ export const useWakeMiner = ({ afterWake, onSuccess, onError }: UseWakeMinerProp
     intervalIdRef.current = newIntervalId;
   }, [fetchMiningStatus, setMiningStatus]);
 
+  const completeWake = useCallback(() => {
```
flesher, line 66: Also doesn't seem related to breakpoint changes.. On purpose to get past some of the new lint rules, or did this sneak in?

flesher (approved):
Looks great!  just noticed a couple things that seem unrelated, wanted to make sure they were included on purpose.

client/src/protoFleet/features/activity/pages/ActivityPage.tsx
negarn: Yes, this was intentional. It is not part of the breakpoint refactor itself, but this page was touched by the responsive class migration and client ESLint then failed on the existing synchronous `setState` inside an effect (`react-hooks/set-state-in-effect`). Switching the first-load flags to refs keeps the same rendered behavior while satisfying the current lint gate.

client/src/protoOS/hooks/useWakeMiner/useWakeMiner.ts
negarn: Intentional, specifically to get past the current client lint gate rather than a breakpoint behavior change. Full client ESLint surfaced the same `react-hooks/set-state-in-effect` rule here; the helper plus scheduled reset preserves the wake-completion behavior while avoiding a synchronous state update directly inside the effect.

PR #117, opened 2026-04-28
feat: add page to view fleet logs

### LINKED ISSUE #112: feat(client): view fleet logs in admin console
### Problem or motivation

Logs should be easily accessible to operators, especially to more easily report issues on GitHub.
In the future all relevant issue information could be packaged up this way and attached to an issue.

### Proposed solution

Tee the logs in a ring buffer and make them available in the fleet UI.
A live tail of the logs can be seen via a polling RPC.

### Duplicate check

- [x] I have searched existing issues and this is not a duplicate

### Alternatives considered

Alternatively we could provide a shell script to package up journald logs, but this would create more friction.

### Component

Client (Web UI)

### Additional context

_No response_

illegalprime:
fixes #112 , the existing log view component is reused.

localhost_5173_settings_server-logs

This also moves protoOS's Logs component into `shared`.


client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+ * (mcdd in the miner case) so the parser's `prefix.split(": ")[1]` path
+ * cleanly extracts the timestamp.
+ */
+function entryToLine(entry: LogEntry): string {
+  const ts = entry.time ? timestampDate(entry.time).toISOString().replace("T", " ") : "";
+  const level = LEVEL_TOKEN[entry.level] ?? "INFO ";
+  const source = entry.source || "fleetd";
+  const attrSuffix = entry.attrs.length ? " " + entry.attrs.map((a) => `${a.key}=${a.value}`).join(" ") : "";
+  return `proto-fleet: ${ts} | ${level} | ${source} ${entry.message}${attrSuffix}`;
+}
```
mcharles-square, line 42: **Claude Code review — string-bridge fragility (worthwhile cleanup).**

`entryToLine` synthesizes a syslog-shaped string just so `formatLogs()` in `shared/components/Logs/utility.ts` can `.split()` it back into pieces. The trailing-space gymnastics is load-bearing: `logTypes.info = " | INFO  | "` (two spaces), so `LEVEL_TOKEN["INFO "]` plus the surrounding `" | "` reconstitutes the exact token. Drop a typo and the row silently renders untyped.

The data is already structured on the wire. A small "structured" branch in `Logs.tsx` keyed off a discriminated-union input would drop the round-trip, the `LEVEL_TOKEN` table, and the parse-fragility.

client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+const ServerLogsPage = () => {
+  const [logsData, setLogsData] = useState<LogsData>();
+
+  const fetchLogs = useCallback(async (limit: number): Promise<LogsData | undefined> => {
+    try {
+      const response = await serverLogClient.listServerLogs({
+        minLevel: LogLevel.UNSPECIFIED,
+        searchText: "",
+        sinceId: 0n,
+        limit,
```
mcharles-square, line 57: **Claude Code review — server filters are dead on the live path (worthwhile cleanup).**

`minLevel`, `searchText`, and `sinceId` are hardcoded, so each 5s poll refetches the whole window and `Logs.tsx` filters client-side. The careful `LatestID` semantics in `buffer.go` (the cursor-doesn't-skip-clipped-matches guarantee) go unused.

Minimum: pass `sinceId = previousLatestId` and append on the client so each poll ships only the delta. Longer term, a server-streaming RPC fits this shape better than polling.

client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+      });
+      const data: LogsData = {
+        content: response.entries.map(entryToLine),
+        lines: response.entries.length,
+      };
+      setLogsData(data);
```
mcharles-square: **Claude Code review — `fetchMaxLogs` mutates live state (real bug, worse than it looks).**

Both `fetchLogs(POLL_LIMIT)` and `fetchLogs(MAX_LIMIT)` go through `setLogsData`. Hitting Export blows the live tail up to 5000 entries — and the dedupe at `Logs.tsx` lines 99–103 only *adds*; `storedLogs` never shrinks. So once Export runs, the live tail stays at 5000+ until page refresh, not "until the next poll".

Fix: have `fetchMaxLogs` return the data without calling `setLogsData`.

client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return data;
+    } catch (err) {
+      // Swallow here — Logs.tsx interprets `undefined` logsData as a
+      // loading state and the next poll will retry. Surfacing the error
+      // to the user would require restructuring the upstream component
+      // and we're explicitly reusing it as-is.
+      console.error("Failed to fetch server logs", err);
+      return undefined;
```
mcharles-square, line 79: **Claude Code review — first-poll failure is invisible (real bug).**

`console.error` + return `undefined`; `Logs.tsx` interprets undefined `logsData` as "still loading" and renders a spinner forever. For a super-admin debug page, an "unreachable / forbidden / buffer not configured" state is exactly what operators need to see.

Surface the error to the UI — even a minimal banner above the table would beat the silent spinner.

client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
```diff
@@ -0,0 +1,129 @@
...
+        the search bar sticks below an empty band where log lines scroll
+        through visibly. This div fills that band with the page title
+        and a solid `bg-surface-base` so nothing bleeds through. z-20
+        keeps it above the search bar (z-10) when both are pinned.
+      */}
+      <div
+        className={
+          "sticky top-0 z-20 flex h-[100px] items-end bg-surface-base px-4 pb-3 " +
+          "laptop:h-[60px] laptop:items-center laptop:px-6 laptop:pb-0"
+        }
```
mcharles-square, line 105: **Claude Code review — sticky-header offset coupling (worthwhile cleanup).**

`h-[100px] laptop:h-[60px]` here is hand-matched to `top-[100px] laptop:top-[60px]` in `shared/components/Logs/Logs.tsx`. Magic numbers duplicated across two files because the shared component doesn't expose a header slot.

Either give `Logs` a `header` prop (rendered above its sticky search bar so offsets are co-located) or compute the offset from a ref.

client/src/shared/components/Logs/Logs.tsx
```diff
@@ -15,9 +14,15 @@ import { useClickOutside } from "@/shared/hooks/useClickOutside";
 import { padLeft } from "@/shared/utils/stringUtils";
 import { getFileName } from "@/shared/utils/utility";
 
+export interface LogsData {
+  content?: string[];
+  lines?: number;
+  source?: string;
+}
+
```
mcharles-square, line 30: **Claude Code review — CSV filename is hardcoded `miner-logs` (real bug).**

The shared component still calls `getFileName("miner-logs", "csv")` for export, so server-log exports will land as `miner-logs-2026-...csv`. Add a `downloadFilename?: string` prop on this `LogsProps` interface and thread it through to the export handler; default to `"miner-logs"` to preserve current behavior, and have `ServerLogsPage` pass `"server-logs"`.

server/internal/infrastructure/logging/buffer.go
```diff
@@ -0,0 +1,265 @@
...
+	records := b.cache.Values()
+	res := SnapshotResult{Size: len(records)}
+	search := strings.ToLower(opts.Search)
+
+	// Values() returns entries in LRU access order. With our usage pattern
+	// (only Add, never Get) that's effectively insertion order — but a
+	// concurrent burst can reorder Adds relative to ID assignment, so we
+	// sort by ID to guarantee oldest-first iteration regardless of the
+	// underlying physical ordering.
+	sort.Slice(records, func(i, j int) bool { return records[i].ID < records[j].ID })
```
mcharles-square: **Claude Code review — `sort.Slice` on every `Snapshot` call (minor).**

This defends against a race the comment itself notes is unlikely (Add-only usage of the LRU). With `DefaultBufferCapacity = 1000` the cost is negligible — but it's still work-per-poll for a contingency that may not exist.

Options: drop the sort with a comment justifying why ordering is sufficient, or switch to an in-place linear pass since `Values()` is already near-sorted. Low priority.

server/internal/infrastructure/logging/buffer.go
```diff
@@ -0,0 +1,265 @@
...
+	if prefix != "" {
+		key = prefix + "." + a.Key
+	}
+	if a.Value.Kind() == slog.KindGroup {
+		for _, sub := range a.Value.Group() {
+			appendAttr(out, key, sub)
+		}
+		return
+	}
+	*out = append(*out, KeyValue{Key: key, Value: fmt.Sprintf("%v", a.Value.Any())})
```
mcharles-square, line 132: **Claude Code review — attribute values lose type fidelity (minor).**

`fmt.Sprintf("%v", a.Value.Any())` is fine for display, but if anyone later wants typed server-side filtering on numeric/bool attrs, the proto's `value` field is a string and the type info is gone before it leaves the buffer.

Worth a one-line comment here calling that out so it isn't a surprise to a future reader.

server/internal/infrastructure/logging/logger.go
```diff
@@ -6,17 +6,27 @@ import (
...
+		logger = slog.NewJSONHandler(os.Stdout, logOptions)
 	} else {
-		logger = slog.New(slog.NewTextHandler(os.Stdout, logOptions))
+		logger = slog.NewTextHandler(os.Stdout, logOptions)
 	}
-	slog.SetDefault(logger)
+
+	buf := NewBuffer(config.BufferSize, config.Level)
+
+	slog.SetDefault(slog.New(newTeeHandler(logger, buf)))
```
mcharles-square: **Claude Code review — `Init` returns the buffer *and* mutates global state (minor).**

Mixed plumbing: explicit DI (returned `*Buffer`) plus an implicit global (`slog.SetDefault`). The handler injected into `serverlog.NewHandler` and the one `slog.Default()` writes through are linked only by this single line in `main.go`.

Pick one: either return the composed handler too and inject everywhere, or expose `logging.DefaultBuffer()` accessor and stop returning it. Today's middle ground works but obscures the wiring.

server/internal/handlers/serverlog/handler.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/block/proto-fleet/server/internal/domain/session"
+	"github.com/block/proto-fleet/server/internal/infrastructure/logging"
+)
+
+// DefaultLimit is applied when the request asks for limit == 0.
+//
+// We cap the default well below the buffer capacity so the wire payload
+// stays small for typical polls; clients tailing live logs only care about
+// recent records and can request more if they need the full window.
+const DefaultLimit = 500
```
mcharles-square, line 18: **Claude Code review — "default limit" defined in three places (cosmetic).**

`DefaultLimit = 500` here, the proto comment says 500, and `ServerLogsPage.tsx` passes `POLL_LIMIT = 1000` explicitly. Not a bug — but the values aren't co-located, so they can drift. Worth a `// keep in sync with ...` cross-reference or a single source-of-truth constant.

mcharles-square (commented):
## Claude Code review

Assessment of an automated code-review pass against the actual diff. Inline comments below — priority summary:

**Real bugs** — sticky export bloat (ServerLogsPage), wrong CSV filename (`miner-logs-*.csv` for server logs), silent first-poll failure stuck on a spinner.

**Worthwhile cleanups** — string-bridge round-trip in `entryToLine`, unused `sinceId` cursor on every poll, hand-matched sticky-header offsets across two files.

**Minor** — defensive `sort.Slice` in `Snapshot`, mixed explicit-buffer + `slog.SetDefault` plumbing, attr values rendered with `%v`, default-limit constant duplicated across handler/proto/page.

Original analysis had two factual slips that I corrected: the sort runs over up to `DefaultBufferCapacity = 1000` records (not 5000 — that's the request cap), and the export-bloat is permanent until refresh (not "trimmed back by the next poll" — `storedLogs` only grows).

client/src/protoFleet/config/navItems.ts
```diff
@@ -87,4 +87,10 @@ export const secondaryNavItems: SecondaryNavItem[] = [
     parent: "/settings",
     allowedRoles: ["SUPER_ADMIN", "ADMIN"],
   },
+  {
+    path: "/settings/server-logs",
+    label: "Server Logs",
+    parent: "/settings",
+    allowedRoles: ["SUPER_ADMIN"],
```
mcharles-square: `ADMIN` can be added to `allowedRoles` for now as well

client/src/protoFleet/config/navItems.ts
```diff
@@ -87,4 +87,10 @@ export const secondaryNavItems: SecondaryNavItem[] = [
     parent: "/settings",
     allowedRoles: ["SUPER_ADMIN", "ADMIN"],
   },
+  {
+    path: "/settings/server-logs",
+    label: "Server Logs",
+    parent: "/settings",
+    allowedRoles: ["SUPER_ADMIN", "ADMIN"],
```
ankitgoswami, line 94: shouldn't this just be SUPER_ADMIN?

server/internal/infrastructure/logging/buffer.go
```diff
@@ -0,0 +1,141 @@
...
+		Attrs:   collectAttrs(r),
+		Source:  formatSource(r.Source()),
+	}
+	rec.ID = b.nextID.Add(1)
+	b.cache.Add(rec.ID, rec)
+	return nil
+}
+
+func (b *Buffer) WithAttrs(_ []slog.Attr) slog.Handler { return b }
+func (b *Buffer) WithGroup(_ string) slog.Handler      { return b }
```
ankitgoswami, line 67: what's the point of these? for future functionality?

server/internal/infrastructure/logging/buffer.go
```diff
@@ -0,0 +1,265 @@
...
+		// Advancing the cursor before the filter check would silently
+		// move clients past matching records that were clipped by Limit.
+		// We only advance the cursor past records the client either saw
+		// (we returned them) or doesn't care about (they didn't match).
+		afterId := rec.ID > opts.SinceID
+		onLevel := rec.Level >= opts.MinLevel
+		matches := search == "" || strings.Contains(strings.ToLower(rec.Message), search)
+
+		if !afterId || !onLevel || !matches {
+			continue
```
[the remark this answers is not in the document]
ankitgoswami: looks legit

client/src/protoFleet/features/serverLogs/pages/ServerLogsPage.tsx
ankitgoswami: agreed

ankitgoswami (approved):

client/src/protoFleet/config/navItems.ts
illegalprime: It used to be, but added from: https://github.com/block/proto-fleet/pull/117#discussion_r3172316468

server/internal/infrastructure/logging/buffer.go
illegalprime: They're to comply with `slog.Handler`'s interface. The code will pipe the logs to both `slog` and this buffer.

client/src/protoFleet/config/navItems.ts
ankitgoswami: oh, then we need to update requireAdmin to also allow ADMIN right?

PR #106, opened 2026-04-27
feat(sv2): accept stratum2+tcp:// URLs for native-V2 miners

ankitgoswami:
## Summary

End-to-end Stratum V2 enablement for native-V2 miners in proto-fleet. After this lands:

- Operators can save `stratum2+tcp://HOST:PORT/<authority_pubkey>` URLs.
- Pool save / Test Connection runs the **real SRI Noise NX handshake** (secp256k1 + ElligatorSwift + ChaCha20-Poly1305 + BIP340 Schnorr), not a TCP dial. A successful test means we can actually open a Noise channel to the pool.
- Pool assignment to a fleet runs a **commit-time SV1↔SV2 capability gate** that consults each device's plugin via `GetCapabilitiesForModel(manufacturer, model)`. Compatible miners get the SV2 URL; incompatible miners are skipped and surfaced in a toast — partial dispatch, not all-or-nothing rejection.
- Native-V2 miners (Braiins-flashed today; future protoOS / Antminer SV2 / WhatsMiner B-series) connect directly with full V2 encryption.

Bridging SV1 miners to SV2 pools via a translator proxy stays out of scope (will be a follow-up).

GIF

## Architecture

```
Operator                              Server                              Plugin
   |                                    |                                    |
   | save stratum2+tcp://h:p/PK ─────►  | CEL: scheme + port + /pubkey       |
   |                                    | sv2.PoolNoiseKeyFromURL (b58/hex)  |
   |                                    | sv2.HandshakeProbe (Noise NX)      |
   |                                    |                                    |
   | assign pool ───────────────────►   | preflightSV2Capabilities           |
   |                                    |   for each (manuf, model):         |
   |                                    |     GetCapabilitiesForModel ──────►|
   |                                    |     read native_stratum_v2 cap   ◄─|
   |                                    |   filter selector to compatible    |
   |                                    |   stamp skipped types in response  |
   |                                    | processCommand ────► dispatch      |
   | toast: "5 of 11 miners couldn't ◄──|                                    |
   |   use this Stratum V2 pool…"       |                                    |
```

### URL validation (`proto/pools/v1/pools.proto`, `proto/minercommand/v1/command.proto`)
- CEL pattern requires `:port` and `/<pubkey>` for `stratum2+tcp://` URLs.
- Inline client-side errors mirror the CEL: `v2MissingPort`, `v2MissingPubkey`, `unknownScheme` — surfaced as the user types, no server roundtrip.

[...]

PR #129, opened 2026-04-29
feat(sdk): extend GetCapabilitiesForModel with manufacturer + add native_stratum_v2 [1/3]

ankitgoswami:
## Summary

- Extend `ModelCapabilitiesProvider.GetCapabilitiesForModel` to take `manufacturer` alongside `model` so plugins like asicrs (which handles Braiins / VNish / LuxOS / Marathon / ePIC on the same hardware family) can return per-firmware caps. The manufacturer here is the firmware's manufacturer.
- Add `CapabilityNativeStratumV2 = "native_stratum_v2"` to the Go and Rust SDKs.
- Plugin implementations:
  - **asicrs**: sets the cap when manufacturer is Braiins (case-insensitive), with the live-probe lookup retained as a fallback for unknown-but-paired devices, now filtered by both manufacturer and model.
  - **antminer**: never sets the cap (stock Bitmain firmware is SV1-only; the plugin rejects non-stock firmware at discovery).
  - **virtual**: reads `StratumV2Supported` from the matched profile config so tests can exercise both branches.
- Reserve `DeviceMetrics.stratum_v2_support` field 16 in `driver.proto` to prevent future reuse. Earlier exploration proposed per-tick SV2 plumbing; the capability is firmware-stable so a per-(manufacturer, model) caps lookup is sufficient.

## Context

This is **PR 1 of 3** splitting #106 into feature-cohesive sub-PRs:

1. **PR1 (this)** — Plugin SDK contract change. No behavior change for users.
2. **PR2** — SV2 endpoint identity verification (Noise NX handshake + URL CEL + \`ValidatePool\` integration + \`credentialsVerified\` callout).
3. **PR3** — SV2 preflight gate + partial dispatch + warning toast (the first consumer of the new \`native_stratum_v2\` cap).

PR1 lands safely standalone — the capability is set but nothing reads it.

## Test plan

- [x] \`go build ./...\` clean across server, antminer, and virtual modules
- [x] \`cargo build && cargo test\` for asicrs — 35/35 pass (Braiins → cap set; VNish/LuxOS/Marathon/ePIC → cap absent; live-probe fallback exercised)
- [x] \`go vet ./...\` clean
- [x] \`go test ./sdk/... ./internal/domain/plugins/...\` — server SDK + plugins package tests pass
- [ ] CI green on this branch


server/sdk/v1/pb/driver.proto
```diff
@@ -388,9 +391,9 @@ message GetDefaultCredentialsResponse {
   repeated UsernamePassword credentials = 1; // List of credentials to try in order
 }
 
-// GetCapabilitiesForModelRequest requests model-specific capabilities
 message GetCapabilitiesForModelRequest {
-  string model = 1; // Device model (e.g., "Antminer S19", "Antminer S21")
+  string model = 1;
```
flesher, line 425: I think both of these warrant a comment documenting the difference between manufacturer and model

plugin/asicrs/src/driver.rs
```diff
@@ -595,42 +631,27 @@ impl Driver for DriverService {
...
-            if device.ensure_connected().await.is_ok() && device.model().await == model {
-                return Ok(Response::new(pb::GetCapabilitiesForModelResponse {
-                    caps: Some(pb::Capabilities {
-                        flags: device.get_caps().await,
-                    }),
-                }));
-            }
+        // Firmware-derived: Braiins is the only asic-rs-handled firmware
+        // that ships native SV2.
+        if manufacturer.eq_ignore_ascii_case(crate::capabilities::VARIANT_BRAIINS) {
```
flesher, line 649: should this check for DISPLAY_BRAIINS since VARIANT_BRAIINS is lowercase?

flesher (approved):

plugin/asicrs/src/driver.rs
ankitgoswami: doing case insensitive comparison here. VARIANT_BRAIINS is the kind of constant used for logical forks in this plugin

server/sdk/v1/pb/driver.proto
ankitgoswami: done

PR #133, opened 2026-04-29
fix(filtering): include AUTHENTICATION_NEEDED in model + firmware dropdowns

### LINKED ISSUE #124: Fix: GetAvailableModels + GetAvailableFirmwareVersions exclude AUTHENTICATION_NEEDED rows
Pre-existing bug discovered during the #114 plan review (Codex finding).

## Problem
`GetAvailableModels` (`server/sqlc/queries/device.sql:362`) only includes miners with `pairing_status = 'PAIRED'`. The Fleet UI shows both `PAIRED` and `AUTHENTICATION_NEEDED` miners, so the model dropdown silently excludes models that exist only on auth-needed devices.

`GetAvailableFirmwareVersions` (added in #119) deliberately mirrors the same pattern for consistency with the existing query, inheriting the same bug.

## Impact
Ops users with miners stuck in `AUTHENTICATION_NEEDED` see incomplete model + firmware filter dropdowns. The miner shows up in the table but its model/firmware doesn't appear as a filter option — confusing.

## Fix (single PR for both queries)
Update both queries to include `PAIRED` and `AUTHENTICATION_NEEDED`:

```sql
WHERE dp.pairing_status IN ('PAIRED', 'AUTHENTICATION_NEEDED')
```

Add regression tests covering each pairing status individually + combined.

## Why one PR for both
Fixing only one creates silent inconsistency between the model and firmware dropdowns. Both queries follow the same pattern, both have the same bug, both deserve the same fix in the same review cycle.

## Files
- `server/sqlc/queries/device.sql` — both queries
- `server/internal/domain/stores/sqlstores/device_test.go` (or new integration test) — regression coverage

Refs #114

flesher:
## Summary
- `GetAvailableModels` and `GetAvailableFirmwareVersions` excluded miners with `pairing_status = 'AUTHENTICATION_NEEDED'`, so the Fleet UI showed those rows in the table but silently dropped their model/firmware values from the filter dropdowns.
- Both queries now match `pairing_status IN ('PAIRED', 'AUTHENTICATION_NEEDED')`, matching the rows the UI surfaces; `discovered_device.model`/`firmware_version` are populated at discovery (pre-auth), so the values are real.
- Adds 8 integration tests covering each pairing status individually, combined, and excluded statuses for both queries.

Closes #124

## Test plan
- [x] `DB_PASSWORD=fleet go test ./internal/domain/stores/sqlstores/...`
- [ ] Verify in Fleet UI that model/firmware dropdowns include values from AUTHENTICATION_NEEDED miners

### DIFF
--- a/server/generated/sqlc/device.sql.go
+++ b/server/generated/sqlc/device.sql.go
@@ -335,7 +335,7 @@ SELECT DISTINCT dd.firmware_version
 FROM device d
 JOIN discovered_device dd ON d.discovered_device_id = dd.id
 JOIN device_pairing dp ON d.id = dp.device_id
-WHERE dp.pairing_status = 'PAIRED'
+WHERE dp.pairing_status IN ('PAIRED', 'AUTHENTICATION_NEEDED')
   AND d.deleted_at IS NULL
   AND d.org_id = $1
   AND dd.is_active = TRUE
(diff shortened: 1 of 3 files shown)


mcharles-square (approved):
Thanks! Feels like we need a linter for this :)

PR #142, opened 2026-04-30
chore(claude): add AGENTS.md and Claude Code agent tooling

mcharles-square:
## Summary

- Establish a single source of agent guidance the team and AI tools can rely on. `AGENTS.md` is read natively by Codex and as a fallback by Claude Code, so both follow the same repo conventions (codegen rules, git/PR workflow, planning-doc structure, etc.). `CLAUDE.md` redirects there and adds the Claude-specific PR review workflow.
- Codify common workflows as Claude Code slash commands (`/regen`, `/pr-ready`, `/release-notes`, `/triage-pr`, `/plan`) and 14 auto-firing skills targeting the failure modes most often seen in this repo: forgetting to regen after proto/sqlc edits, forgetting to rebuild the Python proto-gen tarball, editing deployed migrations, missing `go work sync`, etc. Full list: `proto-regen`, `db-generation-hygiene`, `migration-immutability`, `python-gen-tarball`, `go-work-sync`, `client-boundaries`, `plugin-contract-tests`, `asicrs-build`, `hermit-tooling`, `docker-compose-dev`, `fake-rig-fixtures`, `buf-version-pinning`, `lefthook-bypass-guard`, `plan-conventions`.
- Block commits on `main`/`master`/detached HEAD via a new lefthook pre-commit (`scripts/lefthook-block-protected-branches.sh`). Pairs with the new "## Git workflow" rule in AGENTS.md. **This is tool-agnostic** — catches Codex, manual `git commit`, and IDE git UIs, not just Claude Code. (An earlier iteration of this PR used a Claude Code `PreToolUse` hook for this; that approach was dropped after review feedback because it depended on `jq` being on PATH, used a cwd-relative script path that broke from subdirectories, and missed `git commit` invocations with global git options like `git -c X commit`. Lefthook avoids all three pitfalls.)

## Test plan

- [x] Lefthook script verified standalone across three states: passes on feature branches, blocks on `main`, blocks on detached HEAD with `<detached HEAD>` placeholder in the message.
- [x] `lefthook validate` against the updated `lefthook.yml` returns "All good".
- [x] All slash commands and skills follow the same minimal skeleton (tight description with concrete path triggers, "What to do" steps, optional "What to avoid" only when it adds counter-intuitive signal). Two simplify passes applied.
- [ ] CI passes — this is a docs/tooling-only change so no source paths in path-filters.yml are touched; expect minimal job activity.
- [ ] (Post-merge) Team members run `just install-hooks` (the standard project setup step) so the new pre-commit guard is active locally.


negarn (approved):

.gitignore
```diff
@@ -4,7 +4,9 @@
 node_modules/
 server/influx_config/.env
 **/.goose/
-.claude/
+# Claude Code repo tooling is committed; keep local per-user settings private.
+.claude/settings.local.json
+.claude/**/*.local.json
```
flesher, line 9: claude code will put worktrees in `.claude/worktrees/` we should ignore as well

.claude/commands/triage-pr.md
```diff
@@ -0,0 +1,38 @@
...
+## Steps
+
+1. Resolve the PR identifier (number or URL) and fetch metadata:
+   `gh pr view $ARGUMENTS --json number,title,state,isDraft,mergeable,mergeStateStatus,headRefName,baseRefName,author,reviewDecision,statusCheckRollup`
+2. Fetch check status:
+   `gh pr checks $ARGUMENTS`
+3. Summarize in this shape:
+   - **PR**: number, title, author, branch
+   - **State**: open/closed/merged, draft, mergeable status
+   - **Reviews**: approval state
```
flesher, line 42: Should we include PR comments in the triage?  Read all feedback, dedupe comments, summarize feedback and rank by priority

flesher (approved):
Looks great. Thanks! couple small notes

PR #145, opened 2026-04-30
RFC 0001: Agent + cloud-server split

ankitgoswami:
## Summary

- Proposes the inaugural RFC process for proto-fleet (`docs/rfcs/`) with a README and template.
- RFC 0001 proposes splitting `fleetd` into a thin on-prem `fleet-agent` and a cloud-deployable `fleetd --mode=server`, with a phased rollout that keeps today's docker-compose deploy (`--mode=combined`) working throughout.
- Includes wire protocol (`AgentGatewayService`), per-org credential model, schema impact, drawbacks, alternatives considered, and 6 phases with mermaid diagrams.

This is a design RFC, not implementation. Land it as `draft` so review can happen on the markdown; subsequent PRs implement phases 1-6.

## Test plan

- [x] Render the markdown on GitHub and confirm mermaid diagrams display
- [x] Verify all relative file links (`../../server/...`) resolve to existing source files at the cited line numbers
- [x] Verify internal anchor links (`#credentials`, `#phase-4-...`) navigate correctly
- [ ] Read end-to-end and check for unresolved questions or contradictions before promoting from `draft` -> `accepted`

### DIFF
--- a/docs/rfcs/0001-agent-server-split.md
+++ b/docs/rfcs/0001-agent-server-split.md
@@ -1,0 +1,162 @@
+# RFC 0001: Agent + cloud-server split
+
+- **Status**: draft
+- **Author(s)**: @ankitgoswami
+- **Created**: 2026-04-30
+- **Last updated**: 2026-05-04
+
+## Summary
+
+Split proto-fleet into:
+
+- An **agent** binary running on-prem next to miners. Owns plugin orchestration and miner I/O.
+- A **server** running either in the cloud (managed) or on-prem in Docker. Owns API, persistence, UI, and fleet-wide state.
+
+Multiple agents per server. Today's `fleetd` keeps a `combined` mode for backward compatibility. Behaves like Foreman/pickaxe.
+
+## Motivation
+
+`fleetd` ships today as one Go binary owning everything from plugin orchestration to the React UI backend. That works for "customer runs Docker on-prem" but blocks two things we want next:
+
+1. **Hosted cloud version**. Customers want a managed proto-fleet they don't deploy themselves. Today's monolith requires Postgres + TimescaleDB on the customer's host.
+2. **Air-gapped miners**. Customers don't want individual miners exposed to the internet. They need an on-prem proxy that fronts their LAN-only miners and forwards telemetry/commands to a remote control plane.
+
+```mermaid
+graph TD
+    UI[React UI] -->|Connect-RPC| FD[fleetd<br/>monolith]
+    FD -->|localhost gRPC| P[plugin children]
+    P -->|HTTP / CGMiner| MINERS[Miners on LAN]
+    FD --> PG[(PostgreSQL)]
+    FD --> TS[(TimescaleDB)]
+```
+
+## Architecture
+
+```mermaid
+graph TD
+    UI[React UI] -->|Connect-RPC| FD[fleetd --mode=server<br/>cloud control plane]
+    FD <-->|gateway streams<br/>Connect-RPC HTTP/2| A1[fleet-agent #1<br/>org A, site 1]
+    FD <-->|gateway streams| A2[fleet-agent #2<br/>org A, site 2]
+    FD <-->|gateway streams| A3[fleet-agent #3<br/>org B]
+    A1 --> M1[Miners site 1]
+    A2 --> M2[Miners site 2]
+    A3 --> M3[Miners org B]
+    FD --> PG[(PostgreSQL)]
+    FD --> TS[(TimescaleDB)]
+```
+
+- The server keeps today's databases and gains a new gateway service for agents.
+- The agent is a single Go binary that boots plugin children, scans for miners, polls them, and proxies telemetry and command traffic to the server.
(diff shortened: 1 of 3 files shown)


docs/rfcs/0001-agent-server-split.md
```diff
@@ -0,0 +1,162 @@
...
+    FD --> TS[(TimescaleDB)]
+```
+
+- The server keeps today's databases and gains a new gateway service for agents.
+- The agent is a single Go binary that boots plugin children, scans for miners, polls them, and proxies telemetry and command traffic to the server.
+- Combined mode keeps everything in one process for existing on-prem deployments.
+
+Key properties:
+
+- **Wire**: Connect-RPC over HTTP/2, with separate streams for telemetry, events, heartbeats, and server-to-agent control traffic. All streams are agent-initiated, so the cloud never needs inbound connectivity into customer networks.
```
mcharles-square, line 54: 👍

docs/rfcs/0001-agent-server-split.md
```diff
@@ -0,0 +1,162 @@
...
+- **Telemetry source of truth**: cloud TimescaleDB. The agent has a small on-disk spool for offline windows.
```
mcharles-square, line 55: spooling makes sense here. another thing not covered here that this got me thinking about is how we manage agent initiated action if fleet isnt able connect to the agent. I'm thinking of the case of auto curtailment under some circumstances for instance agent cant connect to fleet but know that based on the power price its better to curtail

docs/rfcs/0001-agent-server-split.md
```diff
@@ -0,0 +1,162 @@
...
+- **WAN dependency for commands**. UI command clicks traverse the WAN twice; combined mode keeps localhost latency.
+- **Enrollment and migration are interactive**. Each agent needs an operator-confirmed pairing, and migrating an org to cloud-mode requires upfront ownership-map declaration plus a Proto re-pair pass. No "walk away" flow.
+- **Compromising one agent host yields the org's stored stratum credentials**. The disclosure is bounded in attack value (no payout-address change, no withdrawal). The actual hashrate-redirection blast radius is gated by the device-scoped authz invariant in any credential model. Per-agent envelope encryption tightens the disclosure boundary further and is a future option.
+- **Revocation requires rotation, not just a delete**. Both the org data key and the affected credentials must be rotated; some per-device cases need manual miner recovery.
+
+## Alternatives considered
+
+- **Ports/adapters carve-out first**. Establish package roots and dependency direction before any wire-protocol work. Cleaner from day one, but defers customer-visible capability by weeks. We do the carve-out last instead, informed by what earlier phases revealed.
```
mcharles-square, line 82: would this work actually add weeks?

docs/rfcs/0001-agent-server-split.md
```diff
@@ -0,0 +1,162 @@
...
+- **Phase 2**: single-agent.
+- **Phase 3**: multi-agent.
+- **Phase 4**: packaged for production deploy.
+- **Phase 5**: architecturally tidy (internal package split).
+- **Phase 6**: self-sufficient (offline buffer, credential edits).
+
+| Phase | Ships | Behavior change |
+| ----- | ----- | --------------- |
+| 1 | Gateway proto, agent-related schema, stub handler, mode flag | None |
+| 2 | Agent binary; per-org credential encryption; per-agent Proto JWT signing with re-pair migration; device-scoped authz; revocation primitives (org-data-key rotation, per-credential rotation, revocation-pending state, gateway-access cutoff); `--mode=server` skips local I/O | New deployment shape, opt-in. Single agent end-to-end with working incident-response controls |
```
flesher, line 156: Do you think we can leave our UI as is up until this step when were assuming a single agent? 
- Does onboarding need to get updated to install and connect agent?
- With a single agent we *could* keep our discovery and pairing flow as is, but we'll need to rethink this with multiple agents

mcharles-square (commented):
Generally looks good

docs/rfcs/0001-agent-server-split.md
ankitgoswami: Single agent should be mostly the same, definitely need at least some UI/UX tweaks for multiple agents. We probably also have to add filters and a dedicated agents page.

docs/rfcs/0001-agent-server-split.md
ankitgoswami: I wasn't thinking that agents would have the capability to take action on auto curtailment. We could add a CLI to these agents so operators would have the option to sending commands directly. But I'm unsure of how important that use-case is.

docs/rfcs/0001-agent-server-split.md
ankitgoswami: Probably not, but being a bit conservative here. What I can see happening is that we do a large refactor but make wrong assumptions about responsibility boundaries etc leading to some churn. So overall, I think it'll probably be faster to build first then clean up.

flesher (approved):

Issue #116, opened 2026-04-28
Add curtailment foundation


rongxin-liu:
### Problem or motivation

We're introducing curtailment to proto-fleet. Curtailment is the ability to reduce a fleet's mining power on demand - manually at first, and later in response to grid signals, site constraints, or price events - and restore it cleanly afterwards.

For v1, operators select a miner scope and a target kilowatt reduction. The system will rank eligible candidates, dispatch FULL device curtailment until the requested kW target is met, verify the reduction through telemetry, and restore miners in controlled batches. Restore needs to be observable per miner, avoid turning on miners that were already inactive before curtailment, and respect maintenance safety checks and emergency override semantics.

The first release is intentionally kW-targeted only. Direct device-level Proto efficiency mode can be advertised by plugins that support it, but Fleet event dispatch in v1 is `FIXED_KW` plus FULL device curtailment. Future webhook-driven triggers, closed-loop modes, and Fleet-level efficiency events should layer onto the same primitives.

### Proposed solution

This issue covers the first backend foundation slice for v1 curtailment: define the public API surface, SDK/plugin contract, generated code, capability flags, route registration, validation behavior, and initial plugin implementations that later persistence, selection, dispatch, reconciliation, and restore work will build on.

The prework this depends on - moving command preflight filtering into `commandSvc` so curtailment can register a `CurtailmentActiveFilter` alongside the existing `ScheduleConflictFilter` - is already merged in #110 (commit `13a3188`).

**Scope**

New proto package `proto/curtailment/v1/curtailment.proto`:

- `CurtailmentService` with `PreviewCurtailmentPlan`, `StartCurtailment`, `UpdateCurtailmentEvent`, `StopCurtailment`, `GetActiveCurtailment`, and `ListCurtailmentEvents`.
- v1 exposes one operator-selectable event mode: `CURTAILMENT_MODE_FIXED_KW` with `FixedKwParams`.
- Request shapes use explicit scope and mode parameter fields for the kW-only contract.
- Validation pins v1 to the implemented combinations where possible: `FIXED_KW`, `fixed_kw`, Fleet level `FULL`/default, strategy `LEAST_EFFICIENT_FIRST`/default, and priority `NORMAL`/default/`EMERGENCY`; `HIGH` remains reserved.
- Maintenance override flags must be paired so callers cannot force maintenance inclusion unless maintenance targets are also included.
- Active-event and cursor-paginated list shapes are present for later domain wiring.

[...]

PR #173, opened 2026-05-05
feat(curtailment): admin RPC, override fields, and session-only auth

### LINKED ISSUE #171: Add curtailment admin RPC, override fields, and session-only auth
### Background

The curtailment foundation (#118) shipped the v1 proto contract, SDK methods, and capability flags. A subsequent review surfaced follow-up additions to the proto/handler/interceptor wiring layer — operationally and security-shaped concerns that the foundation alone did not cover, scoped as a follow-up rather than retconned into the original foundation.

The directional intent is operational and security correctness at the contract layer:

- Dead-reconciler recovery needs a manual operator escape hatch — `AdminTerminateEvent` — so a wedged process is recoverable without direct DB intervention. The recovery surface should not be reachable via a long-lived bearer token.
- The recovery RPC needs idempotency so operator double-clicks or retried runbook calls do not re-apply the terminal transition.
- Large-fleet operators need per-request overrides on candidate-eligibility floor, restore batch size, and an unbounded-duration acknowledgement so per-org defaults are not the only knobs. Override fields should be admin-only at the handler layer with explicit `buf.validate` sanity bounds.
- External integrations need to drive curtailment via the public API. `StartCurtailment` / `StopCurtailment` / `UpdateCurtailmentEvent` and the read RPCs should remain API-key-accessible; the per-user, role-inheriting API-key model is the existing contract for write access elsewhere in the codebase.
- `AdminTerminateEvent` itself needs an Admin role gate from day one so the recovery RPC is not exercisable by lower-privileged sessions even while the body returns `Unimplemented` in the early phase.

### Design direction

Three surfaces contribute to the recovery + override + auth-scope shape: the proto contract, handler wiring, and interceptor configuration. No persistence, no reconciler/dispatcher logic — those live in the persistence/dispatch/restore/read-API tickets that follow.

**`AdminTerminateEvent` RPC**

- A new `AdminTerminateEvent(AdminTerminateEventRequest) returns (AdminTerminateEventResponse)` on `CurtailmentService`. Takes `(event_uuid, target_state, reason, idempotency_key)` and forces a non-terminal event to a terminal state.
- `target_state` should be restricted via `buf.validate.field.enum.in: [6, 7]` to `CANCELLED` (=6) and `FAILED` (=7). `COMPLETED` and `COMPLETED_WITH_FAILURES` belong rejected — this RPC fires when restore did not actually run, and marking such an event as cleanly completed would misreport the real outcome in the audit trail.
- `idempotency_key` belongs on the request so a re-issued call with the same key can return the original event without re-applying, once the read-API business logic lands.
- The handler enforces an Admin / SuperAdmin role check (`requireAdminFromContext`) before the body so non-admin sessions receive `PermissionDenied` from day one. The body itself can return `Unimplemented` until the read-API ticket lands.
- The handler-side admin helper should remap missing `session.Info` to `Unauthenticated` rather than leaking `Internal`, so a future interceptor-wiring regression reports "no identity" instead of "server bug".
- `CategoryCurtailment` belongs in the activity event-category taxonomy so the read-API ticket has a category to emit `curtailment_admin_terminate` activity entries under. No event-type Go constant is needed — the codebase emits activity types as bare strings at the call site.
- The surface backs the dead-reconciler operational runbook — auto-recovery on `fleetd` restart is the first response, and the RPC is the manual fallback when the process truly cannot be restored.

**Request field additions (admin-only at the handler)**

- `optional uint32 candidate_min_power_w_override` on `StartCurtailmentRequest` and `PreviewCurtailmentPlanRequest` — per-request override of the per-org default. Bounded to `[1, 10_000_000]` (10 MW per miner sanity ceiling), preserving the "zero means use default" convention by excluding zero.
- `optional uint32 restore_batch_size_override` on `StopCurtailmentRequest` — per-request override of the adaptive batch size. Bounded to `[1, 10_000]`, again excluding zero so unset means "use default". Documented to take precedence over any prior `UpdateCurtailmentEvent` value for the duration of restore.
- `bool allow_unbounded` on `StartCurtailmentRequest` — explicit acknowledgement that skips `max_duration_default_sec` normalization. Without it, `max_duration_seconds=0` is normalized to the per-org default at handler entry in the start/dispatch ticket.

The handler-level admin gate fires when any override field is set on Preview / Start / Stop. The gate is keyed on role, not auth method: non-admin session and API-key callers should be rejected; admin-role session and API-key callers reach the current `Unimplemented` stubs. Plain Start/Stop/Update calls with no override remain API-key-accessible, matching the existing public-API contract.

**Session-only auth scope**

- Only `AdminTerminateEvent` belongs in `SessionOnlyProcedures` (`server/internal/handlers/interceptors/config.go`). A leaked API key cannot drive the recovery escape hatch.
- All other curtailment RPCs (`PreviewCurtailmentPlan`, `StartCurtailment`, `UpdateCurtailmentEvent`, `StopCurtailment`, `GetActiveCurtailment`, `ListCurtailmentEvents`) remain API-key-accessible so external integrations can drive curtailment via the public API.
- Code comments should cross-reference the `SessionOnlyProcedures` registration and handler-side admin gate as a paired invariant — neither layer alone is sufficient, and the cross-reference helps a future maintainer notice when one drifts away from the other.

**Generated outputs**

Regenerate Go and TypeScript protobuf outputs alongside the proto change.

### Properties the contract should provide

- The proto exposes `AdminTerminateEvent`, the `idempotency_key` field, and the three override / acknowledgement request fields, with the documented `buf.validate` bounds.
- Generated Go and TypeScript protobuf outputs regenerate cleanly.
- The `AdminTerminateEvent` handler stub enforces `AdminRoleName` / `SuperAdminRoleName` and returns `Unimplemented` at this layer; non-Admin callers receive `PermissionDenied` from the role check, not from the body.
- Missing `session.Info` on an admin-gated handler path returns `Unauthenticated`, not `Internal`.
- Override fields on Preview / Start / Stop trigger the same admin gate when set; the gate is keyed on role, not auth method.
- `AdminTerminateEvent` is registered in `SessionOnlyProcedures`; an interceptor test asserts that an API-key-authenticated request to it receives `PermissionDenied`.
- `StartCurtailment`, `StopCurtailment`, `UpdateCurtailmentEvent`, and the read RPCs remain reachable via API-key auth so external integrations can drive curtailment via the public API.
- Existing curtailment-foundation behavior is unchanged for non-curtailment commands and for read RPCs.

### Out of scope (handed off to follow-on tickets)

- The `curtailment_org_config` table and the per-org `candidate_min_power_w` default the override resolves against — owned by the persistence/preview ticket.
- Role-gate enforcement on EMERGENCY / maintenance and `max_duration_default_sec` normalization at handler entry — owned by the start/dispatch ticket.
- The adaptive batch sizing that `restore_batch_size_override` overrides — owned by the restore ticket.
- `AdminTerminateEvent` business logic and the `curtailment_admin_terminate` activity entry — owned by the read-API ticket.
- The end-to-end auth-chain integration test that wires `NewAuthInterceptor` and the curtailment handler together with fake user/role stores. The interceptor and handler layers are tested separately at this layer; the integration coupling lands when the read-API ticket has business logic to test against.
- A scoped-API-key primitive that would let `AdminTerminateEvent` accept API-key callers with an explicit `curtailment.admin` scope — separate larger feature.
- Consolidating role checks across the curtailment and apikey handlers into a shared `domain/auth` helper — separate small refactor.

### Component

Server (Go backend)

rongxin-liu:
## Background

Curtailment is the proto-fleet feature that reduces a fleet's mining power on demand — operators can preview a plan, start a curtailment event for a selected scope of miners, verify the reduction via telemetry, and restore the miners safely afterwards. Use cases include grid-program participation, demand response, and operator-initiated power reduction.

The curtailment foundation (#116, merged in #118) shipped the v1 proto contract — six RPCs (`PreviewCurtailmentPlan`, `StartCurtailment`, `UpdateCurtailmentEvent`, `StopCurtailment`, `GetActiveCurtailment`, `ListCurtailmentEvents`), the full enum surface (modes, strategies, levels, priorities, event/target states), capability flags, and Connect handler stubs that all return `Unimplemented`. Persistence, the selector, the reconciler, the restorer, and the read-API logic land in subsequent tickets that fill in those stubs.

This PR is a contract-layer follow-up to that foundation: a small set of additions to the same proto / handler / interceptor surface that need to be in place before the downstream tickets wire up bodies. Every new RPC stub still returns `Unimplemented`; the value here is contract shape and auth scoping.

## Summary

Four scoped changes:

- **`AdminTerminateEvent` RPC** for the dead-reconciler operational runbook. Forces a non-terminal event to a terminal state. `target_state` CEL is restricted to `CANCELLED` (=6) and `FAILED` (=7); `COMPLETED` states are intentionally rejected so the recovery RPC cannot misreport an event whose restore did not actually run. Carries an `idempotency_key` field so operator double-clicks during the runbook collapse to a single recovery action.
- **Three admin-gated optional override request fields**, each with `buf.validate` sanity ceilings:
  - `optional uint32 candidate_min_power_w_override` on `PreviewCurtailmentPlanRequest` and `StartCurtailmentRequest` — per-org default override for the candidate-eligibility floor; bounded to `[1, 10_000_000]` (10 MW per miner).
  - `bool allow_unbounded` on `StartCurtailmentRequest` — explicit acknowledgement to skip `max_duration_default_sec` normalization.
  - `optional uint32 restore_batch_size_override` on `StopCurtailmentRequest` — takes precedence over any prior `UpdateCurtailmentEvent` value for the duration of restore; bounded to `[1, 10_000]`. The "zero means use default" convention used elsewhere in the proto is preserved by the lower bound on each override.
- **Session-only registration** of `AdminTerminateEvent` only. The recovery escape hatch must not be reachable via a long-lived bearer token. `StartCurtailment` / `StopCurtailment` / `UpdateCurtailmentEvent` and the read RPCs remain API-key-accessible so external integrations can drive curtailment via the public API. The session-only listing and the handler-side admin gate are now cross-referenced in code comments as a paired invariant — neither alone is sufficient.
- **`CategoryCurtailment`** added to the activity event-category taxonomy so subsequent tickets can emit curtailment-domain events under a dedicated category.

Also: when the handler-side admin gate sees no `session.Info` in context, it now returns `Unauthenticated` instead of propagating `Internal` from `session.GetInfo`. The auth interceptor should prevent this in production, but if interceptor wiring ever regresses, the response code reflects "no identity" rather than "server bug" — quieter alerts, no retry-storm encouragement.

## Security posture

| RPC | API-key auth? | Admin role required? |
|---|---|---|
| `PreviewCurtailmentPlan` | yes | only when an override field is set |
| `StartCurtailment` | yes | only when an override field is set |
| `StopCurtailment` | yes | only when an override field is set |
| `UpdateCurtailmentEvent` | yes | no (current contract) |
| `GetActiveCurtailment` | yes | no |
| `ListCurtailmentEvents` | yes | no |
| `AdminTerminateEvent` | **no — session only** | yes |

The handler-level admin gate (`requireAdminFromContext`) fires on `AdminTerminateEvent` always, and on Preview / Start / Stop when the corresponding override field is set on the request. The current API-key model is per-user with role inheritance, so an admin-role API key can drive override-bearing requests — matching how the apikey handler itself gates admin operations on role alone. A leaked admin API key already has full curtailment-write blast radius via plain `StartCurtailment`; restricting the override paths further would not close that gap.

`AdminTerminateEvent`'s session-only registration narrows the blast radius for the operator-of-last-resort recovery escape hatch specifically. A future scoped-API-key primitive would be the right place to relax it.

## What changed

Eight logical commits:

1. `feat(curtailment): add admin RPC and admin-gated override fields` — proto contract additions, regenerated Go + TypeScript outputs.
2. `feat(activity): add curtailment event category` — `CategoryCurtailment` enum entry plus a table-driven test covering all three `Valid()` switches in `activity/models.go` (`EventCategory`, `ActorType`, `ResultType`).
3. `feat(curtailment): gate admin and override paths at the handler` — single `requireAdminFromContext` helper called from `AdminTerminateEvent` (always) and from Preview / Start / Stop when the corresponding override field is set. Action-verb error messages are routed through two package-level constants (`actionSupplyOverrideFields`, `actionTerminateEvents`).
4. `feat(interceptors): register curtailment write RPCs as session-only` — initial registration of all four write/admin procedures.
5. `feat(curtailment): allow API key access to write RPCs except admin recovery` — narrows the previous commit so only `AdminTerminateEvent` remains session-only. Test surface updated to match: invert the role-gate matrix to assert admin-via-API-key reaches `Unimplemented` on override paths, expand the "API-key-accessible" assertion list, scope the runtime SessionOnly interceptor test to the single remaining session-only procedure.
6. `refactor(curtailment): trim handler doc and round out viewer × auth-method matrix` — small post-review polish: trim a 4-line doc comment to one line and add three missing viewer-via-API-key cases on Start/Stop so the override-gate test matrix is symmetric across auth methods.
7. `refactor(curtailment): rename AdminTransitionEvent to AdminTerminateEvent` — addresses the inline review nit. The RPC's `target_state` CEL is restricted to terminal states only, so `Terminate` describes operational behavior more accurately than `Transition` (which describes the underlying state-machine mechanism). Mechanical rename across the proto, generated outputs, handler, tests, and SessionOnly registration.
8. `refactor(curtailment): apply review feedback` — review-driven hardening pass. Adds `idempotency_key` (= 4) to `AdminTerminateEventRequest`; adds `buf.validate` `{gte: 1, lte: ...}` sanity ceilings on the two override `uint32` fields; remaps the handler-side admin gate's missing-session case from `Internal` to `Unauthenticated` (tests updated accordingly); cross-references the paired AdminTerminateEvent role gates in code comments at both sites so neither layer is removed independently.

- Bodies for `PreviewCurtailmentPlan`, `StartCurtailment`, `UpdateCurtailmentEvent`, `StopCurtailment`, `GetActiveCurtailment`, `ListCurtailmentEvents` — those land in subsequent persistence, dispatch, and read-API tickets.
- `AdminTerminateEvent` business logic — handler stub returns `Unimplemented` after the role gate. The transition logic and `curtailment_admin_terminate` activity emission land with the read-API ticket.
- Selector consumption of `candidate_min_power_w_override` — the proto field exists; the persistence/preview ticket consumes it.
- Adaptive batch sizing that `restore_batch_size_override` overrides — lands with the restore ticket.
- Handler-side `max_duration_default_sec` normalization driven by `allow_unbounded` — lands with the start/dispatch ticket.
- A scoped-API-key primitive that would let `AdminTerminateEvent` accept API-key callers with an explicit `curtailment.admin` scope — separate larger feature.
- Consolidating role checks across the curtailment and apikey handlers into a shared `domain/auth` helper — separate small refactor PR.
- End-to-end interceptor → handler chain integration test (each layer is unit-tested in isolation today; bundled E2E coverage is deferred).

## Test plan

- [x] `buf lint` clean
- [x] `npm run lint` (eslint, max-warnings 0) clean
- [x] `golangci-lint run` clean across all three lint targets (`server`, `plugin/proto`, `plugin/antminer`)
- [x] Targeted Go tests pass for the changed packages:
  - `go test ./internal/handlers/curtailment/...`
  - `go test ./internal/handlers/interceptors/...`
  - `go test ./internal/domain/activity/models/...`
- [x] `just gen` regenerates Go + TypeScript outputs cleanly with no orphan diff (proto + generated outputs are in the same commit per `AGENTS.md` rule 2).
- [x] `TestHandler_OverrideFieldsRoleGate` covers 16 cases — the full `(Preview / Start-candidate / Start-allow_unbounded / Stop) × (session / API-key) × (viewer / admin)` matrix: viewer is rejected on every override path regardless of auth method; admin reaches `Unimplemented` on every override path regardless of auth method. Verifies the override gate is keyed on role, not on auth method.
- [x] `TestHandler_NoOverrideSkipsRoleGate` confirms Preview / Stop without overrides reach `Unimplemented` regardless of session info.
- [x] `TestHandler_AdminTerminateEventRoleGate` covers admin / super-admin / viewer / empty-role paths against `AdminTerminateEvent` directly.
- [x] `TestHandler_AdminTerminateEventValidation` covers the buf.validate constraints: `target_state` rejects all six non-CANCELLED/FAILED values (`UNSPECIFIED`, `PENDING`, `ACTIVE`, `RESTORING`, `COMPLETED`, `COMPLETED_WITH_FAILURES`); `event_uuid` and `reason` `min_len=1` enforced. Validator-passed cases now surface `CodeUnauthenticated` from the handler-side admin gate (no session in test context), matching the remapped error code.
- [x] `TestHandler_AdminTerminateEventRejectsMissingSession` asserts the handler-side admin gate returns `CodeUnauthenticated` when no `session.Info` is in context.
- [x] `TestAuthInterceptor_AdminTerminateEventRejectsApiKeyAuth` exercises the runtime `authenticate()` path with a Bearer header against `AdminTerminateEvent` and asserts `PermissionDenied`.
- [x] `TestCurtailmentAdminProcedureIsSessionOnly` and `TestCurtailmentNonAdminProceduresStayApiKeyAccessible` pin the registration list both directions.

Closes #171
Refs #116
Refs #118


proto/curtailment/v1/curtailment.proto
```diff
@@ -32,6 +32,11 @@ service CurtailmentService {
     rpc ListCurtailmentEvents(ListCurtailmentEventsRequest)
         returns (ListCurtailmentEventsResponse);
 
+    // Admin recovery RPC: force a non-terminal event to a terminal state.
+    // Session-only, Admin role. Manual fallback for the dead-reconciler runbook.
+    rpc AdminTransitionEvent(AdminTransitionEventRequest)
```
ankitgoswami: nit: would AdminTerminateEvent be a better name?

ankitgoswami (approved):

proto/curtailment/v1/curtailment.proto
rongxin-liu: Good call — applied in 696c609. The RPC, its messages (`AdminTerminateEventRequest`/`Response`), the SessionOnly procedure constant, the action-verb error string, and the corresponding test names all renamed. `Terminate` better matches the CEL constraint (`target_state in [CANCELLED, FAILED]`) which makes this RPC end-of-life only.

PR #182, opened 2026-05-06
Add numeric telemetry and subnet filters to fleet miners

flesher:
https://github.com/user-attachments/assets/41992f5e-5d08-467e-adcc-fa06248fad66

🤖 This PR adds numeric telemetry filters and subnet/CIDR filtering to the ProtoFleet miners experience end to end. It wires the new filter types through the fleet-management API and store layer, exposes them in the client filter UI and saved views, and includes follow-up fixes for selection scope, filter clearing, IPv6 validation, and mobile E2E stability.

**BE**
- Add numeric range and subnet/CIDR fields to the fleet management filter proto and generated server/client contracts.
- Parse and validate telemetry range filters and subnet filters in fleet-management service logic, including IPv4 and routable IPv6 CIDRs/bare IPs.
- Extend SQL filter/query building so miner list queries, device identifier lookups, and state counts all honor numeric and subnet filters consistently.
- Add store and service test coverage for numeric filtering, CIDR filtering, state counts, model-group plumbing, and query fragment generation.

**FE**
- Add miner list UI support for numeric range filters and subnet entry, including modal-based editing for numeric/text-area filters and updated nested filter dropdown behavior.
- Persist the new filters through URL params, saved views, and filter summary/chip formatting so they round-trip correctly.
- Align client-side subnet validation with existing discovery behavior by accepting routable IPv4/IPv6 input while rejecting unsupported scoped or link-local IPv6.
- Fix filter UX regressions so `Clear all filters` removes numeric/subnet filters and bulk selection resets when those filters change.
- Fold the view-actions popover behavior into `ViewsBar` and remove the now-redundant standalone component/plumbing.
- Update client unit tests, stories, and the mobile miners E2E page helper to cover the new filter flows and stabilize nested-filter dismissal on mobile.


client/src/protoFleet/features/fleetManagement/utils/filterUrlParams.ts
```diff
@@ -123,22 +155,36 @@ export function encodeFilterToURL(filter: MinerListFilter): URLSearchParams {
     setMulti(params, URL_PARAMS.ZONE, [...filter.zones].sort());
   }
 
+  filter.numericRanges.forEach((range) => {
+    const key = numericKeyFromProtoField(range.field);
+    if (!key) return;
+    if (range.min !== undefined) params.append(numericMinParam(key), String(range.min));
+    if (range.max !== undefined) params.append(numericMaxParam(key), String(range.max));
+  });
+
```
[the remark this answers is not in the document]
flesher: 🤖 Triaged as non-blocking for this PR. The current UI only permits one numeric range per field, so normal app usage cannot emit duplicate `${key}_min` / `${key}_max` params. I left parsing as-is for now; if we later allow multiple ranges per field in the client/proto model, we should collapse or reduce `getAll()` explicitly at that point.

client/src/shared/utils/filterValidation.ts
```diff
@@ -0,0 +1,100 @@
...
+export const validateCidrLine = (line: string): string | null => {
+  const trimmed = line.trim();
+  if (trimmed === "") return "Empty value";
+
+  if (trimmed.includes("/")) {
+    return isValidCidr(trimmed) ? null : "Not a valid CIDR (e.g. 255.255.255.0/24)";
+  }
+
+  return isValidIpv4(trimmed) ? null : "Not a valid IPv4 address or CIDR";
+};
```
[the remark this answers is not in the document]
flesher: 🤖 Fixed earlier in this PR. The client subnet validator/normalizer now accepts bare IPv6 and IPv6 CIDRs, normalizes bare IPv6 to `/128`, and rejects the unsupported link-local/scoped forms so the UI matches the existing server/discovery contract.

server/internal/domain/stores/sqlstores/device_filters.go
```diff
@@ -221,6 +279,46 @@ func appendFilterSQL(sb *strings.Builder, args []any, argNum int, orgID int64, f
 		argNum++
 	}
 
+	if len(fp.numericRanges) > 0 {
+		for _, r := range fp.numericRanges {
+			col := numericFieldColumn(r.Field)
+			if col == "" {
+				continue
```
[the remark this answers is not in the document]
flesher: 🤖 Fixed in `b43a79f`. `GetDeviceIdentifiersByOrgWithFilter` now adds `latestMetricsCTE` + `minerTelemetryInnerJoin` whenever numeric filters are active, and I added query-builder coverage plus an integration test for the numeric-filtered identifier path.

flesher:
🤖 Addressed the actionable Codex security review findings in `b43a79f`:

- `GetDeviceIdentifiersByOrgWithFilter` now joins `latest_metrics` when numeric filters are active.
- `GetMinerStateCounts` now uses the dynamic path for numeric/CIDR filters so `total_state_counts` stays aligned with the filtered list.
- CIDR filtering now guards blank stored IPs with `NULLIF(discovered_device.ip_address, )::inet`.
- The earlier frontend findings around IPv6 validation, clear-all filters, and selection reset were also fixed in this PR.

I verified the server changes with focused `go test` coverage against the local DB config and the client changes with targeted Vitest runs.

server/internal/domain/stores/sqlstores/device.go
```diff
@@ -928,28 +1016,166 @@ func (s *SQLDeviceStore) executeListQuery(ctx context.Context, orgID int64, curs
...
+	args := []any{orgID}
+	argNum := 2
+
+	filterNeedsTelemetry := len(fp.numericRanges) > 0
+	if filterNeedsTelemetry {
+		fmt.Fprintf(&sb, latestMetricsCTE+" ", "NULL")
+	}
+
+	sb.WriteString(`SELECT COUNT(*)` + minerFromJoins)
+	if filterNeedsTelemetry {
```
ankitgoswami, line 1059: looks like this pattern is repeated a few times with slight variations. Maybe we need something like a appendTelemetryCTEPrefix helper

server/internal/domain/stores/sqlstores/device_filters.go
```diff
@@ -34,6 +34,25 @@ type minerFilterParams struct {
...
+	ipCIDRsFilter sql.NullString
+	// ipCIDRValues are pre-stringified prefixes (already normalized by
+	// parseFilter) suitable for pq.Array on a $N::cidr[] parameter.
+	ipCIDRValues []string
+}
+
+// filterRequiresTelemetry reports whether any predicate in filter targets a
+// device_metrics column. When true, callers must build the latest_metrics CTE
+// and INNER-join it so missing telemetry naturally excludes the row.
+func filterRequiresTelemetry(filter *stores.MinerFilter) bool {
```
ankitgoswami: i don't see this used anywhere, was it supposed to be used for filterNeedsTelemetry?

server/internal/domain/stores/sqlstores/device_filters.go
```diff
@@ -221,6 +279,46 @@ func appendFilterSQL(sb *strings.Builder, args []any, argNum int, orgID int64, f
...
+		// Match the UI's em-dash semantics: OFFLINE miners never expose a
+		// telemetry value, so a numeric predicate should not surface them
+		// even if a fresh metric exists.
+		sb.WriteString(" AND (device_status.status IS NULL OR device_status.status != 'OFFLINE')")
+	}
+
+	if fp.ipCIDRsFilter.Valid {
+		fmt.Fprintf(sb,
+			" AND EXISTS (SELECT 1 FROM unnest($%d::cidr[]) c"+
+				" WHERE NULLIF(discovered_device.ip_address, '')::inet <<= c)",
```
ankitgoswami: this was pretty hard to understand (without feeding to an LLM) and I have concerns around index usage. I wonder if we should just migrate this column to inet at this point?

server/internal/domain/stores/sqlstores/device_filters.go
flesher: 🤖 Correct, this was dead code. I had originally split telemetry gating into a helper, but the real callsites ended up keying off `len(fp.numericRanges) > 0` after `buildMinerFilterParams()`. I removed the unused helper and its obsolete unit test coverage in `d8d3148`.

ankitgoswami (approved):
this looks super useful!

server/internal/domain/stores/sqlstores/device_filters.go
flesher: 🤖 I agreed with the concern and implemented the follow-up in `304e260`.

Instead of casting the `text` column inline in the filter, we now add a generated `inet` column (`discovered_device.ip_address_inet`) plus a GiST network index, and switch the CIDR predicate/sort expression to use that native column directly. That makes the SQL much easier to read and gives us a path for index support on subnet containment.

server/internal/domain/stores/sqlstores/device.go
flesher: 🤖 Agreed. I pulled that repeated `latest_metrics` CTE-prefix assembly into a small helper in `611e5ff`.

That keeps the repeated `filterNeedsTelemetry` branching in one place while still handling the one variation where `buildStateCountsQuerySQL` needs to continue an existing `WITH` chain.

PR #185, opened 2026-05-06
feat(curtailment): persistence + preview selector

### LINKED ISSUE #140: Add curtailment persistence and preview selector
#116 / #118 establishes the curtailment foundation: the public API shape, SDK/plugin `Curtail` / `Uncurtail` contract, level-specific capabilities, and reachable but unimplemented RPC handlers.

#171 / #173 layered the admin surface on top: the `AdminTerminateEvent` recovery RPC (session-only), three admin-gated override fields on Preview/Start/Stop (`candidate_min_power_w_override`, `allow_unbounded`, `restore_batch_size_override`), the handler-side `requireAdminFromContext` helper, and the `CategoryCurtailment` activity event category.

The next step is to make the first operator-facing RPC useful. `PreviewCurtailmentPlan` should let an operator see which miners would be curtailed, why they were selected or skipped, and the estimated power reduction before any event is created or command is dispatched. v1 preview is kW-targeted through `FIXED_KW`.

### Proposed solution

Add the curtailment persistence layer and implement `PreviewCurtailmentPlan` behind the existing curtailment service contract. This remains read-only: preview computes and returns a plan but does not create curtailment events, write target rows, dispatch commands, or start reconciliation.

Builds on the API/SDK foundation from #116 / #118, the admin layer from #171 / #173, and the command preflight groundwork from #110.

**Scope**

Persistence:

- Add migrations for `curtailment_event`, `curtailment_target`, and the singleton `curtailment_reconciler_heartbeat` table.
- Add indexes for active-event lookup, idempotency, external-reference dedupe, pending target work, active target lookup by device identifier, and event creation time.
- Seed the heartbeat singleton row.
- Enforce the maintenance override invariant at the database layer: `force_include_maintenance=true` requires `include_maintenance=true`.
- Persist the admin-gated request fields landed in #173 onto `curtailment_event`: `allow_unbounded BOOLEAN NOT NULL DEFAULT FALSE` (the explicit acknowledgement to skip `max_duration_default_sec` normalization) and `effective_batch_size INT` (the adaptive restore batch size; populated at restore start in a follow-up ticket but the column lands here).
- Add a `curtailment_org_config` table for per-org tunables: `max_duration_default_sec INT NOT NULL DEFAULT 14400` (4 hours), `candidate_min_power_w INT NOT NULL DEFAULT 1500`, `post_event_cooldown_sec INT NOT NULL DEFAULT 600`. Seed one row per existing org in the same migration transaction so subsequent reads always hit a value. The down migration drops the table.
- Add sqlc queries and store wrappers needed for curtailment events, targets, and preview candidate reads.

Preview selector:

- Implement `PreviewCurtailmentPlan`.
- Resolve scope from whole org, device sets, or explicit device identifiers; validate org ownership before persistence or dispatch (do not rely on generic command `include_devices` resolution as the authorization boundary).
- Filter candidates by org, pairing/device status, telemetry freshness, `curtail_full` capability, cooldown, and maintenance override settings.
- Require the dual telemetry signal for candidates: sufficient power draw and active hashrate/share activity.
- Rank candidates least-efficient first.
- Resolve `candidate_min_power_w` in two tiers — the per-org default from `curtailment_org_config`, plus the admin-gated `candidate_min_power_w_override` field on the request when non-nil. The handler-side admin role gate from #173 has already enforced role at this point; the selector reads the effective threshold once and uses it consistently across the filter and rejection-detail attribution.
- Support the v1 preview mode `FIXED_KW`: select ranked candidates until the requested kW reduction is met, or fail for insufficient available load.
- Return ordered candidates, skipped candidates with canonical reasons, estimated reduction, remaining candidate power, and echoed `fixed_kw` params. Insufficient-load rejection includes a structured detail (`available_kw`, `requested_kw`, `tolerance_kw`, `candidate_min_power_w`, per-reason exclusion counts) so the UI can render the diagnostic context directly.

Validation and behavior:

- Keep Fleet event-level curtailment at `FULL` only.
- Use strict `0 kW` default tolerance when `FIXED_KW.tolerance_kw` is omitted or explicitly zero; operators can explicitly set positive tolerance when a near-miss is acceptable.
- Keep non-preview RPCs returning `Unimplemented` until follow-up work implements start, stop, reconciliation, restore, and read APIs.

Tests / safety:

- Add a `CurtailmentEventState` enum-stability test pinning `CANCELLED == 6` and `FAILED == 7`. `AdminTerminateEventRequest.target_state` from #173 uses `(buf.validate.field).enum.in: [6, 7]` — a numeric pin that goes silently wrong if the enum is reordered. The test makes the validator coupling visible at build time.
- Tighten the org-scoping acceptance with a concrete cross-tenant query test: an `event_uuid` belonging to org B returns empty/not-found when queried under org A's context.

**Out of scope**

- `StartCurtailment` business logic.
- Dispatching `Curtail` commands through command execution.
- Active-event reconciliation and drift handling.
- `StopCurtailment` and staggered restore (the `effective_batch_size` column lands here; the restorer that populates it is a follow-up ticket).
- `GetActiveCurtailment`, `ListCurtailmentEvents`, and `UpdateCurtailmentEvent` implementations.
- `CurtailmentActiveFilter` registration for blocking schedule/manual commands.
- An admin RPC for mutating `curtailment_org_config` — the table and seed exist; the write surface lands later.
- Webhook-triggered curtailment.
- Fleet-level efficiency events.
- Smart PDU / PDU-outlet / rack target controls.
- Frontend settings page and preview/start flow.

**Acceptance criteria**

- [ ] Migration up/down applies cleanly.
- [ ] The heartbeat singleton row exists after migration.
- [ ] `curtailment_event` carries `allow_unbounded BOOLEAN NOT NULL DEFAULT FALSE` and `effective_batch_size INT`.
- [ ] `curtailment_org_config` exists with one row per existing org after the migration; the down migration drops the table.
- [ ] The maintenance override database constraint rejects `force_include_maintenance=true` with `include_maintenance=false`.
- [ ] sqlc generation succeeds and curtailment store tests cover event/target persistence primitives needed by preview.
- [ ] `PreviewCurtailmentPlan` returns ordered candidates and skipped candidates for whole-org, device-set, and explicit-device scopes.
- [ ] Candidate filtering covers stale telemetry, no hashrate/share activity, unreliable power telemetry, offline devices, unsupported pairing/device states, maintenance exclusion, cooldown, and missing `curtail_full` capability.
- [ ] Selector consumes `candidate_min_power_w_override` from the request when non-nil, otherwise reads the per-org default from `curtailment_org_config`.
- [ ] `FIXED_KW` preview covers target reached, last-miner overshoot, explicit tolerance near-miss, strict omitted/zero-tolerance behavior, and insufficient-load rejection with structured detail.
- [ ] Cross-tenant query test: an `event_uuid` belonging to org B returns empty/not-found when queried under org A's context.
- [ ] `CurtailmentEventState` enum-stability test asserts `CANCELLED == 6` and `FAILED == 7` (pins the validator coupling shipped in #173).
- [ ] Non-preview curtailment RPCs remain reachable and continue returning `Unimplemented`.

### Component

Server (Go backend)

rongxin-liu:
## Background

Curtailment is the proto-fleet feature that reduces a fleet's mining power on demand — operators preview a plan, start an event, verify reduction via telemetry, then restore safely.

The contract foundation (#118) shipped the proto surface with handler stubs returning `Unimplemented`. The admin layer (#173) added `AdminTerminateEvent`, three admin-gated override fields, session-only registration of the recovery RPC, and `requireAdminFromContext`. This PR is the next ticket: persistence + the first operator-facing RPC. Builds on #118, #173, and the command-preflight groundwork from #110. Supersedes the closed #141.

## Summary

Lands the curtailment persistence layer plus a working **read-only `PreviewCurtailmentPlan`**. Preview computes a plan but does not write events, write target rows, dispatch commands, or start reconciliation. The remaining curtailment RPCs continue to return `Unimplemented` and ship in subsequent tickets.

Six logical commits, layered top-down:

1. `6f7d59c` **Foundation** — migration `000040`, sqlc queries, store interface + sqlstore, domain models, `FixedKw` mode + tests, `TestCurtailmentEventStateNumericPins` guard for the `AdminTerminateEvent` validator from #173.
2. `c6d9b9a` **Selector** — pure-logic `BuildPlan` with the dual-signal filter, ranking, and mode hand-off. 10 unit tests.
3. `b5e20b1` **Service + handler** — orchestration, the cross-table candidate-aggregation query, proto translators, fleetd startup wiring.
4. `8b2f64e` **Service tests** — 14 cases against an in-memory fake store covering scope resolution, the filter matrix, cooldown vs. EMERGENCY bypass, the maintenance-pair gate, the admin-gated override, cross-tenant isolation, and insufficient-load detail propagation.
5. `0f3cc14` **Lint cleanup** — golangci `exhaustive` / `gosec` / `unconvert`.
6. `58c90ec` … `d329ff3` **Review-pass fixes** — bug fixes from a multi-reviewer pass plus a comment scrub and small simplifications. Detailed below.

### Migration `000040`

- `curtailment_event` — full lifecycle columns plus the admin fields shipped in #173 (`allow_unbounded`, `effective_batch_size`). CHECK constraints for the maintenance-pair invariant and non-empty `external_source` / `external_reference` / `idempotency_key` / `reason`. Partial UNIQUE indexes for idempotency, webhook dedupe, and active-event lookup.
- `curtailment_target` — composite PK `(curtailment_event_id, device_identifier)`. Partial indexes for pending work and active-by-device lookup.
- `curtailment_reconciler_heartbeat` — singleton row with `CHECK (id = 1)`. Seeded so the staleness alert always reads a row.
- `curtailment_org_config` — per-org tunables. One row per existing non-deleted org seeded in the same transaction; down migration drops the table.

### Preview pipeline

- `Service.Preview` validates the request → resolves scope (whole-org / device-list; device-set returns `Unimplemented`) → loads org config → resolves the effective `candidate_min_power_w` (admin-gated request override > org default) → fetches active-event and cooldown exclusion sets (`EMERGENCY` priority bypasses cooldown) → classifies candidates with skip-reason attribution → hands the partition to `BuildPlan` + `FixedKw`.
- `BuildPlan` is pure (no I/O, no time, no shared state): dual-signal filter (`power_w >= candidate_min_power_w` AND `hash_rate_hs > 0`, with distinct skip reasons for phantom-load / dead-monitor / below-threshold) → rank by `avg_efficiency` descending with unknown-efficiency last → mode hand-off.
- `FixedKw` walks the ranking and produces target-reached / undershoot-tolerated / insufficient-load. `tolerance_kw == 0` collapses the undershoot branch.
- `handlers/curtailment.Handler` accepts an optional `*Service`: `nil` keeps the existing stub-test contract; populated wires the real implementation. The admin gate from #173 still runs first when the request carries an override field. `OutcomeInsufficientLoad` maps to `InvalidArgument` carrying `available_kw`, `requested_kw`, `tolerance_kw`, and per-reason exclusion counts (Connect-RPC error-detail propagation is a future enhancement).
- `cmd/fleetd/main.go` instantiates `NewSQLCurtailmentStore` + `curtailmentDomain.NewService` and threads them into the curtailment route registration.
- A new `domain/curtailment/models` package holds boundary types so the rest of the curtailment domain doesn't import sqlc-generated code.

### Review-pass fixes

Layered on top of commits 1–4, from a multi-reviewer pass. All P0 / P1 *behavior* findings addressed; coverage-only items routed to follow-ups.

- **P0** — New orgs created post-migration no longer break Preview with `NotFound`. `GetOrgConfig` is now an `INSERT ... ON CONFLICT DO UPDATE` upsert that always returns a row with table-level DEFAULTs.
- **P1** — `ListCurtailmentCandidatesByOrg` no longer crashes the row scan when a device has no `device_status` row joined (`COALESCE(ds.status::text, ''::text)::text`).
- **P1** — `InsufficientLoadDetail` per-reason counters (below-threshold / phantom-load / dead-monitor) now populate; `BuildPlan` tracks each category locally and merges into the rejection detail post-`mode.Select`.
- **P1** — Empty `DeviceStatus` is skipped with `SkipStaleTelemetry` instead of admitted as eligible.
- **P1** — Device-list scope with cross-org `device_identifiers` rejects with `NotFound` listing the unrecognized IDs instead of returning `InsufficientLoad`.
- **P2** — `latest_hourly` CTE is org-scoped and 24h-bucket-bounded.
- **P2** — `candidate_min_power_w_override = 0` is rejected at the service layer.
- **P2** — Capability gate (partial). Devices missing driver metadata are skipped with `SkipCurtailFullUnsupported`. The full plugin-registry-driven check is follow-up work.
- **P3** — `reason_selected` is derived from the request strategy via `strategyReasonLabel(s)` instead of hard-coded.
- **P3** — Migration's per-org seed excludes soft-deleted organizations.

## Operational notes for reviewers

- **Write-on-read for `curtailment_org_config`** — `GetOrgConfig` upserts on every Preview call. The conflict arm is a PK-keyed no-op; on first hit for a new org it inserts a row with table DEFAULTs (`max_duration_default_sec=14400`, `candidate_min_power_w=1500`, `post_event_cooldown_sec=600`).
- **`latest_hourly` CTE plan** — the new shape pulls strictly fewer rows (org-scoped + 24h window).
- **Migration immutability** — `000040` is not yet on `main`, so this PR edits it in place. The next ticket inherits the post-merge immutability rule.

- `StartCurtailment`, `Curtail` dispatch, reconciler, and drift handling.
- `StopCurtailment` and staggered restore. The `effective_batch_size` column lands here; the restorer that populates it lives in a follow-up.
- `GetActiveCurtailment`, `ListCurtailmentEvents`, `UpdateCurtailmentEvent` implementations.
- `CurtailmentActiveFilter` registration on `commandSvc`.
- An admin RPC for mutating `curtailment_org_config` — the table and seed exist; the write surface lands later.
- Device-set scope resolution; full plugin-registry-driven `curtail_full` capability gating (partial driver-metadata gate landed inline); webhook-triggered curtailment; Fleet-level efficiency events; smart PDU / outlet / rack target controls.
- Frontend settings page and preview/start flow.
- Coverage-only follow-ups from the review pass: handler / `translate.go` / SQL-integration tests, migration CHECK DB-level test, `OutcomeUndershootTolerated` end-to-end test, `InsufficientLoad` printf format pinning, `Strategy` / `Mode` / `Level` drop-or-wire decision, P3 polish.

## Test plan

- [x] `buf lint` clean; `golangci-lint run` clean for the changed packages; `sqlc generate` clean against the migration; `go build ./server/...` clean.
- [x] `go test ./internal/domain/curtailment/... ./internal/handlers/curtailment/... ./internal/domain/activity/models/...` — 39+ cases green.
- [x] `TestCurtailmentEventStateNumericPins` asserts the validator pins from #173 (`CANCELLED == 6`, `FAILED == 7`).
- [x] `modes/FixedKw` covers all three outcomes, boundary ordering, empty-input guard, and constructor input validation.
- [x] `selector_test.go` covers the dual-signal matrix, pre-filtered forwarding, ranking (worst-efficiency-first, unknown-last, stable equal-efficiency), realized + remaining kW arithmetic, and an end-to-end `FixedKw` integration.
- [x] `service_test.go` covers (against an in-memory fake store): happy path; request validation; scope resolution; the pre-selector filter matrix; maintenance-pair gate; cooldown vs. EMERGENCY bypass; active-event exclusion; `candidate_min_power_w_override` precedence; **cross-tenant isolation**; insufficient-load detail propagation; the partial capability gate; and four review-pass regressions (`EmptyDeviceStatusSkipsAsStale`, `DeviceListScopeRejectsCrossOrgIdentifiers`, `DualSignalCountersPropagateIntoInsufficientLoadDetail`, `MissingDriverSkipsAsCurtailFullUnsupported`).

Closes #140
Refs #110
Refs #118
Refs #173


rongxin-liu:
Closing in favor of a fresh PR off the same branch. Earlier iterations accumulated stale review feedback (now resolved in code) that adds noise to ongoing review. Re-opening with the same diff and PR description on the same branch; will link the replacement here once it's up.

rongxin-liu:
Replacement: #188. Same branch, same diff, same description — fresh review thread.

PR #221, opened 2026-05-12
feat(fleetnode): operator CLI for enroll + refresh + status

ankitgoswami:
## What's in this PR

CLI under \`server/cmd/fleetnode/\`:

```text
fleetnode enroll --server-url=<url> [--name] [--force] [--allow-insecure-transport] [--state-dir]
fleetnode refresh [--state-dir]
fleetnode status  [--state-dir]
```

### Security posture

| Hardening | Where |
| --- | --- |
| Secrets via stdin only, masked on TTY, shared scanner when piped | \`secrets.go\` (\`secretReader\`) |
| No \`--code=\` / \`--api-key=\` flags | absent from \`EnrollCmd\` |
| Partial \`SaveState\` immediately after \`Register\` (Ctrl-C is recoverable) | \`enroll.go\` step 7 |
| \`ErrRegisterRejected\` → operator-friendly recovery hints (revoke / --name / fresh code) | \`enroll.go\` |
| \`ErrBeginAuthRejected\` does NOT claim "bad api_key" — lists all server-side causes | \`enroll.go\` + \`refresh.go\` |
| \`refresh\` prompts for missing api_key, saves *before* handshake | \`refresh.go\` |
| \`refresh\` preserves local credentials on Unauthenticated | library + CLI message |
| \`--force\` warns about orphaned server-side fleet_node row | \`enroll.go\` |
| \`status\` never prints api_key or session_token plaintext (booleans only) | \`status.go\` |
| 30s timeout + no-redirect HTTP client | \`fleetnodebootstrap/client.go\` |
| Atomic 0o600 inside 0o700, symlink rejection, advisory flock, nil-state guards | \`fleetnodebootstrap/\` |

### Tests

11 test files covering the matrix above (\`secrets_test\`, \`enroll_test\`, \`refresh_test\`, \`status_test\`, \`validate_test\`, plus a shared \`fake_gateway_test\` fixture).

## Dependencies

Re-adds \`golang.org/x/term\` as a direct dep (used by \`secretReader\`). \`go mod tidy\` clean.

## Test plan

- [x] \`PATH=./bin:\$PATH go build ./server/...\`
- [x] \`PATH=./bin:\$PATH go vet ./server/...\`
- [x] \`cd server && PATH=../bin:\$PATH golangci-lint run -c .golangci.yaml ./...\` — 0 issues
- [x] \`PATH=./bin:\$PATH go test ./server/cmd/fleetnode/... -count=1\` — all pass
- [x] \`DB_PASSWORD=fleet PATH=./bin:\$PATH go test ./server/... -count=1\` — 61 packages OK, no FAILs
- [x] CI runs the full suite
- [x] Manual smoke 

### End-to-end smoke (manual)

1. \`just dev\` → gateway on \`:4000\`.
2. Sign into the admin UI as SuperAdmin or Admin (the cookie is required — \`CreateEnrollmentCode\` is in \`SessionOnlyProcedures\`).
3. Mint a code via the UI or:
   \`curl -b cookies.txt -X POST http://localhost:4000/fleetnodeadmin.v1.FleetNodeAdminService/CreateEnrollmentCode -H 'Content-Type: application/json' -d '{}'\`
4. \`cd server && PATH=../bin:\$PATH go build -o ../bin/fleetnode ./cmd/fleetnode\`
5. \`./bin/fleetnode enroll --server-url=http://localhost:4000 --allow-insecure-transport\` — CLI prompts for code (masked), prints fingerprint, prompts for api_key (masked).
6. \`./bin/fleetnode status\` — fleet_node_id, fingerprint, server_url, \`api_key_present: true\`, \`session_token_present: true\`, session expiry.
7. \`./bin/fleetnode refresh\` — session expiry shifts forward.
8. **Partial-save recovery**: re-enroll with \`--force\`, Ctrl-C between code prompt and api_key prompt → \`status\` shows the new fleet_node_id with \`api_key_present: false\` → \`fleetnode refresh\` prompts for the api_key and finishes the enrollment.

### DIFF
--- a/server/cmd/fleetnode/enroll.go
+++ b/server/cmd/fleetnode/enroll.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/block/proto-fleet/server/internal/fleetnodebootstrap"
+)
+
+type EnrollCmd struct {
+	ServerURL              string `required:"" help:"base URL of the fleet server, e.g. https://fleet.example.com"`
+	Name                   string `help:"fleet node name; defaults to os.Hostname() when empty"`
+	Force                  bool   `help:"overwrite an existing populated state file"`
+	AllowInsecureTransport bool   `name:"allow-insecure-transport" help:"permit non-https server URLs for non-loopback hosts; testing only"`
+}
+
+func (e *EnrollCmd) Run(c *Context) error {
+	return e.run(c, os.Stdin, os.Stdout, os.Stderr)
+}
+
+func (e *EnrollCmd) run(c *Context, stdin io.Reader, stdout, stderr io.Writer) error {
+	if err := fleetnodebootstrap.ValidateServerURL(e.ServerURL, e.AllowInsecureTransport); err != nil {
+		return err
+	}
+	return fleetnodebootstrap.WithStateLock(c.StateDir, func() error {
+		return e.runLocked(c, stdin, stdout, stderr)
+	})
+}
+
+func (e *EnrollCmd) runLocked(c *Context, stdin io.Reader, stdout, stderr io.Writer) error {
+	path := fleetnodebootstrap.StatePath(c.StateDir)
+	st, exists, err := fleetnodebootstrap.LoadState(path)
+	if err != nil {
+		return err
+	}
+	if exists && st.FleetNodeID != 0 && !e.Force {
+		return fmt.Errorf("state already populated at %s; pass --force to overwrite", path)
+	}
+
+	name := e.Name
+	if name == "" {
+		host, err := os.Hostname()
+		if err != nil {
+			return fmt.Errorf("resolve hostname: %w", err)
+		}
+		name = host
+	}
+
+	if exists && st.FleetNodeID != 0 && e.Force {
+		_, _ = fmt.Fprintf(stderr, "warning: --force discarded local state for fleet_node_id=%d. If %q is still registered server-side, Register will fail; revoke the prior fleet node in the operator UI first or pass --name=<unique-value>.\n", st.FleetNodeID, name)
+	}
+
+	secrets := newSecretReader(stdin, stderr)
+	code, err := secrets.read("Paste the one-time enrollment code from the operator UI:\n> ")
+	if err != nil {
+		return fmt.Errorf("read enrollment code: %w", err)
+	}
+	if code == "" {
+		return errors.New("empty enrollment code")
+	}
+
+	result, err := fleetnodebootstrap.Register(context.Background(), fleetnodebootstrap.RegisterParams{
+		ServerURL:              e.ServerURL,
+		Name:                   name,
+		Code:                   code,
+		AllowInsecureTransport: e.AllowInsecureTransport,
+	})
+	if err != nil {
+		if errors.Is(err, fleetnodebootstrap.ErrRegisterRejected) {
+			return fmt.Errorf("register rejected by server: %w\n  recovery: revoke the prior fleet node in the operator UI (then re-run with --force), or pass --name=<unique-value> to register as a new fleet node. If the enrollment code was already used, typo'd, or expired, request a fresh one from the operator UI", err)
+		}
+		return fmt.Errorf("register: %w", err)
+	}
+	state := result.State
+
+	// Persist before the api_key prompt so a Ctrl-C cannot orphan the
+	// server-side fleet_node row; the operator can complete the enrollment
+	// by running `fleetnode refresh` and entering the api_key when prompted.
+	if err := fleetnodebootstrap.SaveState(path, state); err != nil {
+		return err
+	}
+
+	_, _ = fmt.Fprintf(stderr, "Fleet node registered (fleet_node_id=%d, name=%q).\n", state.FleetNodeID, name)
+	_, _ = fmt.Fprintf(stderr, "Identity fingerprint: %s\n", state.IdentityFingerprint)
+	_, _ = fmt.Fprintf(stderr, "Compare this fingerprint against the value shown in the operator UI before pasting the api_key.\n")
+
+	apiKey, err := secrets.read("Once you confirm enrollment, the UI will display an api_key. Paste it here:\n> ")
+	if err != nil {
+		return fmt.Errorf("read api_key: %w", err)
+	}
+	if apiKey == "" {
+		return errors.New("empty api_key")
+	}
+
+	if err := fleetnodebootstrap.CompleteEnrollment(context.Background(), state, apiKey); err != nil {
+		if errors.Is(err, fleetnodebootstrap.ErrBeginAuthRejected) {
+			return fmt.Errorf("server rejected BeginAuthHandshake: %w\n  the server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Verify the api_key matches the one minted in the UI, then retry with `fleetnode refresh`; local credentials are preserved", err)
+		}
+		return fmt.Errorf("complete enrollment: %w", err)
+	}
+	if err := fleetnodebootstrap.SaveState(path, state); err != nil {
+		return err
+	}
+	_, _ = fmt.Fprintf(stdout, "enrolled fleet_node_id=%d fingerprint=%s state=%s\n", state.FleetNodeID, state.IdentityFingerprint, path)
+	return nil
+}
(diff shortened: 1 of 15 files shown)


server/cmd/fleetnode/enroll.go
```diff
@@ -0,0 +1,109 @@
...
+		return fmt.Errorf("read api_key: %w", err)
+	}
+	if apiKey == "" {
+		return errors.New("empty api_key")
+	}
+
+	if err := fleetnodebootstrap.CompleteEnrollment(context.Background(), state, apiKey); err != nil {
+		if errors.Is(err, fleetnodebootstrap.ErrBeginAuthRejected) {
+			return fmt.Errorf("server rejected BeginAuthHandshake: %w\n  the server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Verify the api_key matches the one minted in the UI, then retry with `fleetnode refresh`; local credentials are preserved", err)
+		}
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in 6190c78. The CLI now wraps only the sentinel `ErrBeginAuthRejected` via `%w`, dropping the underlying server-provided text. The operator-facing message is deterministic; `errors.Is(err, ErrBeginAuthRejected)` still works. Test updated to use `ErrorIs` and to assert the leaked "invalid api_key" substring is absent.

server/cmd/fleetnode/refresh.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := fleetnodebootstrap.SaveState(path, st); err != nil {
+			return err
+		}
+	}
+
+	if err := fleetnodebootstrap.Refresh(context.Background(), st); err != nil {
+		if errors.Is(err, fleetnodebootstrap.ErrBeginAuthRejected) {
+			return fmt.Errorf("server rejected BeginAuthHandshake: %w\n  the server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Local credentials are preserved; retry once the operator-side cause is resolved", err)
+		}
+		return fmt.Errorf("refresh: %w", err)
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in 6190c78, same sanitization as enroll. Refresh wraps only the sentinel; test asserts the server text doesn't leak.

server/cmd/fleetnode/refresh.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (r *RefreshCmd) Run(c *Context) error {
+	return r.run(c, os.Stdin, os.Stdout, os.Stderr)
+}
+
+func (r *RefreshCmd) run(c *Context, stdin io.Reader, stdout, stderr io.Writer) error {
+	return fleetnodebootstrap.WithStateLock(c.StateDir, func() error {
+		return r.runLocked(c, stdin, stdout, stderr)
+	})
+}
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in 6190c78. `refresh` now calls `LoadState` (pure, no side effects) before `WithStateLock`, so the "no state" bailout happens before any `MkdirAll`/lock-file creation on never-enrolled hosts. Inside the lock we re-load the state to handle the (rare) race where the file disappears between the check and the lock. Added `TestRefreshCmd_DoesNotTouchStateDirWhenNeverEnrolled` that asserts the state dir is not created when refresh bails out.

server/cmd/fleetnode/secrets.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return "", fmt.Errorf("read from terminal: %w", err)
+		}
+		return strings.TrimSpace(string(b)), nil
+	}
+	if sr.scanner == nil {
+		sr.scanner = bufio.NewScanner(sr.stdin)
+		sr.scanner.Buffer(make([]byte, 0, 1024), scannerMaxLine)
+	}
+	if !sr.scanner.Scan() {
```
flesher, line 50: Scan will return false if input does not have trailing new line.  scanner will have no error so it will result in "no input on stdin" which is misleading.  Maybe check for scanner.Text() as well.

server/cmd/fleetnode/enroll.go
```diff
@@ -0,0 +1,109 @@
...
+			return fmt.Errorf("register rejected by server: %w\n  recovery: revoke the prior fleet node in the operator UI (then re-run with --force), or pass --name=<unique-value> to register as a new fleet node. If the enrollment code was already used, typo'd, or expired, request a fresh one from the operator UI", err)
+		}
+		return fmt.Errorf("register: %w", err)
+	}
+	state := result.State
+
+	// Persist before the api_key prompt so a Ctrl-C cannot orphan the
+	// server-side fleet_node row; the operator can complete the enrollment
+	// by running `fleetnode refresh` and entering the api_key when prompted.
+	if err := fleetnodebootstrap.SaveState(path, state); err != nil {
```
flesher, line 82: if this fails after the row on the server was created the node would not have the fleet node id to revoke it.  

```go
if err := fleetnodebootstrap.SaveState(path, state); err != nil {
    return fmt.Errorf(
        "save state after Register: %w\n"+
            "  recovery: a fleet_node row was created server-side "+
            "(fleet_node_id=%d, name=%q) but no local credentials were persisted. "+
            "Revoke that fleet node in the operator UI before retrying "+
            "`fleetnode enroll`, or pass `--name=<unique-value>` to register as a new fleet node",
        err, state.FleetNodeID, name,
    )
}
```

server/cmd/fleetnode/enroll.go
```diff
@@ -0,0 +1,109 @@
...
+		return errors.New("empty api_key")
+	}
+
+	if err := fleetnodebootstrap.CompleteEnrollment(context.Background(), state, apiKey); err != nil {
+		if errors.Is(err, fleetnodebootstrap.ErrBeginAuthRejected) {
+			return fmt.Errorf("%w. The server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Verify the api_key matches the one minted in the UI, then retry with `fleetnode refresh`; local credentials are preserved", fleetnodebootstrap.ErrBeginAuthRejected)
+		}
+		return fmt.Errorf("complete enrollment: %w", err)
+	}
+	if err := fleetnodebootstrap.SaveState(path, state); err != nil {
```
flesher, line 111: same here, but even worse since the server issues api key that the fleet node would not know about. 

```go
if err := fleetnodebootstrap.SaveState(path, state); err != nil {
    return fmt.Errorf(
        "save state after CompleteEnrollment: %w\n"+
            "  recovery: enrollment succeeded server-side "+
            "(fleet_node_id=%d, name=%q) and an api_key was issued, "+
            "but no local credentials were persisted. Revoke that fleet node "+
            "in the operator UI and re-enroll; the issued api_key cannot be recovered",
        err, state.FleetNodeID, name,
    )
}
```

flesher (approved):

server/cmd/fleetnode/enroll.go
ankitgoswami: good catch!

server/cmd/fleetnode/enroll.go
ankitgoswami: 👍

server/cmd/fleetnode/secrets.go
ankitgoswami: looks like it is handling the case when there is no trailing new line correctly - I'll add a test to confirm

server/cmd/fleetnode/enroll.go
ankitgoswami: Fixed in a9bb8ae. Used your suggested wording verbatim — when the post-Register SaveState fails, the error now surfaces `fleet_node_id` and `name` so the operator can locate the orphan row in the admin UI before retrying. Clean point, thanks.

server/cmd/fleetnode/enroll.go
ankitgoswami: Fixed in a9bb8ae. Same shape — wraps with `fleet_node_id` and `name`, and the message explicitly calls out that the issued api_key cannot be recovered so the operator knows revoke+re-enroll is the only path. Used your wording verbatim.

server/cmd/fleetnode/secrets.go
ankitgoswami: Verified in a9bb8ae with a new regression test, `TestSecretReader_FinalLineWithoutTrailingNewline`. Turns out the default `bufio.ScanLines` splitter explicitly handles the no-trailing-newline case via its `atEOF` branch — see the stdlib snippet:

```go
// bufio.ScanLines:
if atEOF { return len(data), dropCR(data), nil }
```

So `printf "code" | fleetnode enroll` works: the first `read()` returns `"code"`, and only a second `read()` hits the "no input on stdin" branch (which is correct — the operator forgot to pipe the api_key line). The test pins this behavior so future-us doesn't doubt it. `scanner.Text()` after `Scan() == false` returns the empty string per the stdlib docs, so the suggested extra check would be a no-op.

PR #232, opened 2026-05-14
feat(curtailment): Stop, staggered restore, max-duration enforcement

### LINKED ISSUE #231: Curtailment: Stop, staggered restore, and max-duration enforcement
## Background

Curtailment ships in stages. The contract foundation, persistence + preview, and the operator-facing Start + dispatch + reconciler all land in prior work. With those in place, an operator can preview a curtailment plan, persist it, and watch the reconciler drive non-terminal events forward through dispatch, telemetry confirmation, and drift detection. What's missing is the path back: an operator-facing way to end an event, and the system's ability to bring miners back online safely.

This work closes that gap and removes the production gate on `StartCurtailment`.

## StopCurtailment

The operator-facing end-of-event RPC. Validates that `min_curtailed_duration_sec` has elapsed (with `EMERGENCY` priority bypassing), transitions the event to `restoring`, and in the same transaction flips every non-terminal target to `desired_state='active'`, resets `state` to `pending`, and clears the per-phase metadata (`retry_count`, `last_dispatched_at`, `last_batch_uuid`, `confirmed_at`, `last_error`). Terminal targets are untouched. The reset gives the restorer a DB-backed, unambiguous work queue that survives a `fleetd` restart.

The Admin-gated `restore_batch_size_override` already lives on the request; the handler keeps the existing role check and lets the value flow into the event's effective batch size.

## Restorer

A DB-backed step inside the reconciler tick rather than a separate goroutine, so a restart resumes restore from persisted state without rehydrating in-memory workers. Each tick, the restorer scans `restoring` events and claims at most one batch per event when both gates pass:

- No prior batch in flight — no targets with `desired_state='active'` and `state ∈ {dispatched, drifted}`.
- Inter-batch interval elapsed — the newest restore-phase `last_dispatched_at` is null or older than `restore_batch_interval_sec`.

The claim selects up to `effective_batch_size` pending targets ordered by a stable key, transitions them to `dispatched`, stamps `last_dispatched_at`, dispatches `Uncurtail` through the standard command pipeline (carrying the reconciler's self-bypass actor so the active-curtailment filter doesn't block it), and records `last_batch_uuid`.

Confirmation runs on later ticks against `baseline_power_w` — captured once at selection and never mutated. The threshold is `baseline_power_w × 0.5` for full-restore; null baseline falls back to the dual-signal predicate the selector uses (`power_w ≥ candidate_min_power_w` AND `hash_rate_hs > 0` sustained for one minute). Reaching all-terminal in a batch plus an elapsed interval allows the next claim. When every target reaches a terminal state, the event transitions to `completed` or `completed_with_failures`.

## Adaptive batch sizing

At restore start, `effective_batch_size = max(restore_batch_size_default, ceil(0.01 × selected_target_count))` clamped to `[10, 100]`, persisted on the event. A 5,000-miner event therefore restores in batches of around 50. The precedence between event-creation value, future `UpdateCurtailmentEvent` writes, the per-request override, and the adaptive formula lands alongside the implementation. The override's proto bound also gets tightened to a real fleet-class ceiling rather than the current ungrounded ten-thousand cap.

## `max_duration_seconds` enforcement

Reconciler-side enforcement transitions an `active` event to `restoring` when the cap elapses. This closes the prior-work gap where a target sitting in `dispatched` with no positive telemetry stays there indefinitely — the per-target confirm path requires positive evidence and does not consume retry budget on missing telemetry. Time-based termination at the event boundary covers that case without per-target redispatch heuristics, and it gives operators a safety net even when the `Stop` RPC is unreachable.

## Restore semantics worth flagging

- **Restore does not re-check `device_status`.** The operator who set `force_include_maintenance=true` owns restore timing. Operators expecting a maintenance window to outlast a curtailment event must size `max_duration_seconds`, not rely on schedule-coupled timing.
- **Per-device FIFO awareness.** Stop does not cancel any already-enqueued `Curtail`; the `Uncurtail` queues behind earlier pending or processing work. The restorer must account for this rather than assuming `Uncurtail` supersedes.

## Flipping the Start gate

`StartCurtailment` has been live but gated off in production because an `active` event with no operator-facing exit path is an operational footgun. With Stop, the restorer, and `max_duration_seconds` enforcement all in place — and a staging soak behind them — the gate flips, or the field is removed entirely.

## Out of scope

- `UpdateCurtailmentEvent` for operator-safe mutation of `reason` and restore controls — lands with the read/audit work.
- `GetActiveCurtailment` and `ListCurtailmentEvents` — same.
- Metrics pipeline wiring and the heartbeat staleness alert — blocked on the still-unresolved metrics-pipeline shape decision.
- Webhook ingestion and signal adapters.

rongxin-liu:
## Background

Curtailment ships in stages. The contract foundation, persistence + preview, and the operator-facing Start + dispatch + reconciler all landed via #192. What's still missing is the path back: an operator-facing way to end an event, and the system's ability to bring miners back online safely. This PR closes that gap and removes the production gate on `StartCurtailment`.

### StopCurtailment

The operator-facing end-of-event RPC. Validates `min_curtailed_duration_sec` has elapsed (admin caller can supply `force=true` to bypass the hysteresis gate), transitions the event to `restoring`, and in the same transaction flips every non-terminal target to `desired_state='active'`, resets `state` to `pending`, and clears the per-phase metadata (`retry_count`, `last_dispatched_at`, `last_batch_uuid`, `confirmed_at`, `last_error`). Terminal targets are untouched. The reset gives the restorer a DB-backed, unambiguous work queue that survives a `fleetd` restart.

`min_curtailed_duration_sec` only gates active events — pending events haven't curtailed anything yet, so the elapsed-since-`started_at` check doesn't apply. Idempotent re-Stop: a second call against an already-restoring event echoes the persisted row without touching targets. Terminal events return `FailedPrecondition`; cross-org event UUIDs return `NotFound`. These semantics are documented on the `StopCurtailment` proto RPC declaration so SDK consumers see the idempotent-restoring vs. non-retryable-terminal split without reading the handler.

`StopCurtailmentRequest` carries `event_uuid` (tag 1) and an admin-gated `force` boolean (tag 3). Tag 2 is reserved — both the number and the name `restore_batch_size_override` — so an old client's varint payload on the deleted slot can never silently alias into `force=true`. The handler enforces the admin role on `force=true` before reaching the service, so non-admin callers see `PermissionDenied` and never trip the bypass.

### Restorer

A DB-backed step inside the reconciler tick rather than a separate goroutine, so a restart resumes restore from persisted state without rehydrating in-memory workers. Each tick scans `restoring` events and:

1. Confirms any dispatched restore targets via telemetry (`baseline_power_w × 0.5` threshold, falling back to a positive-hash signal when baseline is null). A target whose candidate row has vanished (device unpaired or deleted) burns retry budget via `recordDispatchFailure` instead of pinning the event indefinitely.
2. Transitions the event terminal — `COMPLETED` or `COMPLETED_WITH_FAILURES` based on RestoreFailed count — once every target is terminal.
3. Claims the next batch when both gates pass:
   - **No prior batch in flight**: zero restore-phase targets in `Dispatched` or `Drifted`.
   - **Inter-batch interval elapsed**: newest restore-phase `last_dispatched_at` is null or older than `restore_batch_interval_sec`.

The claim selects up to `effective_batch_size` pending targets ordered by `device_identifier`, dispatches a single batched `Uncurtail` call covering all of them (sharing one `batch_uuid`), and per-device commits state transitions based on the kept/skipped split returned by the command service. Bulk dispatch errors and per-device skips route through the same `recordDispatchFailure` retry path the Curtail dispatch uses, and the per-device audit string comes from a single `skippedDeviceReason` helper shared with single-device dispatch (Reason > FilterName > generic preflight fallback). The reconciler self-bypass on `CurtailmentActiveFilter` covers `Uncurtail` traffic too, so restore batches aren't blocked by the very event they're restoring.

Each tick body is bounded by a `2 × TickInterval` deadline. Per-event contexts derive from the same tick budget, so the loop-entry gate and per-event deadline together cap total tick wall-time. A stuck DB during a large restore (`ResetCurtailmentTargetsForRestore` over thousands of targets, slow `ListCandidates` on a large fleet) returns an error to the tick instead of hanging the goroutine; the heartbeat upsert uses its own background-derived 5s deadline so liveness reporting stays honest even when the tick body errors out. Telemetry-timeout aging in the restore phase logs a dedicated slog line before `recordDispatchFailure` so triage can distinguish timed-out-and-wrote from timed-out-DB-error.

### Adaptive batch sizing — computed at Start

`effective_batch_size = max(restore_batch_size, ceil(0.01 × selected_target_count))` clamped to `[10, 100]`, stamped on the event row at Start time from the selected-target count and read back by Stop, the reconciler's max-duration arm, and the restorer's batch claim. The column is non-null from event creation. `StartCurtailmentResponse.event.effective_batch_size` echoes the computed value, so SDK consumers see the resolved batch size in the same response as the new event UUID rather than needing a follow-up `GetActiveCurtailment` poll.

A 5,000-miner event therefore restores in batches of 50; a 10K+ event caps at 100. Operators tune the formula's floor via the `restore_batch_size` field on `StartCurtailmentRequest`.

### One non-terminal event per org

Migration 000051 adds a partial unique index `uq_curtailment_event_one_non_terminal_per_org ON curtailment_event (org_id) WHERE state IN ('pending','active','restoring')`. `InsertEventWithTargets` maps the unique-violation to a typed `interfaces.ErrCurtailmentNonTerminalEventExists` sentinel, and `Service.Start` catches it: a concurrent Start that races the selector's per-device exclusion check now fails at insert time with `AlreadyExists`, surfaced with the existing event's UUID + state. The `AlreadyExists` DebugMessage embeds that identity as the stable substring `(event_uuid=<uuid>, state="<state>")` — documented on the `StartCurtailment` RPC declaration so SDK consumers can recover the existing event without parsing an undocumented format.

This closes the cross-event scope-overlap gap at the DB level. The rule is org-wide rather than device-set-aware (two non-overlapping device sets in the same org are still mutually exclusive), which is sufficient for v1's manual-trigger use case.

Migration 000051 also bounds the operational controls at the DB layer with CHECK constraints on `max_duration_seconds` (`NULL OR 0 < x <= 604800`) and `restore_batch_interval_sec` (`0 <= x <= 3600`). The migration is safe without `CONCURRENTLY` / `NOT VALID` because `curtailment_event` has zero rows at deploy (the production `StartCurtailment` gate is removed by the same PR); the file documents this assumption inline for any future re-application against a populated table.

### max_duration_seconds enforcement

Reconciler-side enforcement transitions an `active` event to `restoring` when the cap elapses since `started_at`. This closes the prior gap where a target sitting in `Dispatched` with no positive telemetry stays there indefinitely under drift detection alone — the cap now drags the event into restore regardless. `allow_unbounded` events and events with nil `started_at` short-circuit out.

When `max_duration_seconds` is unset and `!allow_unbounded`, `Service.Start` fills the value from `orgConfig.MaxDurationDefaultSec` and re-checks the absolute upper bound (`maxFiniteDurationSeconds = 604800`) on the normalized value. A misconfigured org default surfaces as `InvalidArgument` (`org's max_duration_default_sec must be <= 604800`) instead of tunnelling past `validateStartRequest` and tripping the DB CHECK constraint as `Internal`.

### StartCurtailment gate flipped

With Stop, the restorer, and reconciler-side max-duration enforcement all in place, an active event now has both an operator-facing exit (Stop) and a wall-clock safety net. The `Handler.startEnabled` field, in-handler gate check, and the coupling test that pinned the gate are removed. `main.go` constructs the handler with the lean `NewHandler(curtailmentSvc)`.

## Restore semantics worth flagging

- **Restore does not re-check `device_status`.** The operator who set `force_include_maintenance=true` owns restore timing. Operators expecting a maintenance window to outlast a curtailment event must size `max_duration_seconds`, not rely on schedule-coupled timing.
- **Per-device FIFO awareness.** Stop does not cancel any already-enqueued `Curtail`; the `Uncurtail` queues behind earlier pending or processing work. The restorer accounts for this via the in-flight gate rather than assuming `Uncurtail` supersedes.
- **`BeginCurtailmentRestoration` concurrency model.** The load-bearing guarantee is the `WHERE state IN ('pending','active')` row-lock guard combined with the per-org partial unique index added in migration 000051. The store's `ErrNoRows` re-read path distinguishes "already restoring" (idempotent) from "already terminal" (`FailedPrecondition`).

## Wiring

`cmd/fleetd/main.go` constructs the handler with the single-argument `NewHandler` and the wiring comment documents the current surface: Preview, Start, and Stop are wired; the reconciler drives non-terminal events terminal (including max-duration-based forced restore); `UpdateCurtailmentEvent`, `GetActiveCurtailment`, and `ListCurtailmentEvents` return `Unimplemented` pending the read APIs ticket.

- `UpdateCurtailmentEvent` for operator-safe mutation of `reason` and restore controls — lands with the read/audit work.
- `GetActiveCurtailment` and `ListCurtailmentEvents` — same.
- Metrics pipeline wiring and the heartbeat staleness alert — blocked on the unresolved metrics-pipeline shape decision.
- Webhook ingestion and signal adapters.

## Test plan

`go build ./...` and `go vet ./...` clean. `golangci-lint run ./internal/domain/curtailment/... ./internal/handlers/curtailment/... ./cmd/fleetd/...` reports 0 issues. `go test` on the curtailment unit surface is green (DB-dependent integration tests fail on local connection-refused as usual).

Coverage by area:

- **`Service.Stop`** — happy path, idempotent already-restoring, terminal rejection, `min_curtailed_duration_sec` gate with `force=true` bypass, EMERGENCY priority does not bypass the gate at Stop time (regression), pending-event short-circuit, store-error propagation.
- **`Service.Start`** — `TestService_Start_StampsEffectiveBatchSize` pins the adaptive formula at Start time (floor 10, `restore_batch_size` floor, 5K → 50, 10K ceiling at 100). `TestService_Start_NonTerminalOverlapReturnsAlreadyExists` pins the per-org partial unique index path; `TestService_Start_NonTerminalOverlapWithGetActiveErrorReturnsBareAlreadyExists` pins the identity-lookup-fallback sub-branch. `TestService_Start_RejectsOrgDefaultAboveAbsoluteCeiling` pins the post-normalization upper-bound guard on the org-default path. `TestService_Start_RejectsAdminRestoreBatchIntervalAboveAbsoluteCeiling` pins the absolute ceiling for admin callers (above the non-admin cap but still bounded).
- **`ComputeEffectiveBatchSize`** — small-fleet floor, 5K-fleet → 50, 10K+ ceiling at 100, `restore_batch_size` floors the formula, negative `restore_batch_size` handling.
- **`StopCurtailment` handler** — happy path through the full handler → service → store flow, missing session `Unauthenticated`, malformed UUID `InvalidArgument`, `force` admin role gate (`PermissionDenied` for non-Admin), admin `force=true` flowing through and bypassing the min-duration gate, `ListTargetsByEvent` error propagation after a committed transition.
- **`StartCurtailment` handler** — happy path asserts `EffectiveBatchSize` on the wire response (proves `toStartResponse` populates the computed value rather than always-zero).
- **Reconciler `enforceMaxDuration`** — elapsed transitions to restoring (no drift dispatch), not-elapsed no-ops, `allow_unbounded` skips cap, nil `started_at` no-ops, `BeginRestoreTransition` error path skips drift this tick.
- **Reconciler `observeRestoring`** — claim dispatches one `Uncurtail` per batch with shared `batch_uuid`, in-flight gate blocks claim, interval gate blocks claim, telemetry-confirmed dispatched promotes to `Resolved` (and triggers terminal transition same tick), all-resolved → `COMPLETED`, mixed-with-failures → `COMPLETED_WITH_FAILURES`. `TestReconciler_Restoring_MissingCandidateDuringConfirmConsumesRetryBudget` pins the nil-candidate guard in the restore phase, mirroring the curtail-phase analog.
- **Reconciler tick budget** — `TestReconciler_RunTickStopsWhenTickBudgetExpires` pins the per-event context bound; later events cancel when the shared tick budget expires.
- **`isRestored`** — power above/at/below `baseline × 0.5` thresholds, baseline-nil positive-hash fallback, baseline-nil zero-hash not-restored, no-telemetry not-restored, baseline-zero hash fallback.
- **`skippedDeviceReason`** — priority-order contract (`Reason` wins, then `FilterName`, then generic preflight fallback) pinned across all three branches.

Closes #231
Refs #192


server/internal/domain/curtailment/models/models.go
```diff
@@ -229,6 +238,23 @@ type InsertTargetParams struct {
...
+// terminal state. Terminal targets (resolved / restore_failed / released) are
+// not restored, so they shouldn't influence batch sizing.
+func CountNonTerminalTargets(targets []*Target) int {
+	n := 0
+	for _, t := range targets {
+		switch t.State {
+		case TargetStateResolved, TargetStateRestoreFailed, TargetStateReleased:
+			continue
+		case TargetStatePending, TargetStateDispatched,
+			TargetStateConfirmed, TargetStateDrifted:
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in c9379e36: `CountNonTerminalTargets` now treats unknown/future `TargetState` values as non-terminal via the default branch, and `models_test.go` covers a synthetic future state so batch sizing will not silently undercount it.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -660,3 +680,380 @@ func isCurtailed(latestPowerW *float64, baselinePowerW *float64, latestHashRateH
...
+			slog.Warn("curtailment reconciler: restore device filter-skipped",
+				"event_id", ev.ID, "device", t.DeviceIdentifier, "reason", reason)
+			r.recordDispatchFailure(ctx, ev, t, reason, models.TargetStatePending)
+			continue
+		}
+		emptyErr := ""
+		params := interfaces.UpdateCurtailmentTargetStateParams{
+			State:            models.TargetStateDispatched,
+			LastDispatchedAt: &now,
+			LastBatchUUID:    &batchID,
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in c9379e36: the sqlc update now treats an explicitly supplied empty `last_error` as a clear signal and writes SQL `NULL`, while omitted `last_error` still preserves the previous value. The generated sqlc bindings were regenerated from the query source.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -201,7 +217,7 @@ func (r *Reconciler) runTick(ctx context.Context) {
 	}
 
 	for _, ev := range events {
-		r.processEvent(ctx, ev)
+		r.processEvent(tickCtx, ev)
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in c9379e36: `runTick` now uses a bounded context for listing events and creates a fresh bounded context for each event before calling `processEvent`, so an earlier slow event cannot spend the deadline budget for later events. Added `TestReconciler_RunTickUsesFreshContextPerEvent` to cover the starvation regression.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -660,3 +680,380 @@ func isCurtailed(latestPowerW *float64, baselinePowerW *float64, latestHashRateH
...
+			reason = "filtered by command preflight"
+		}
+		skippedSet[s.DeviceIdentifier] = reason
+	}
+
+	now := r.now()
+	batchID := result.BatchIdentifier
+	for _, t := range claim {
+		if reason, skipped := skippedSet[t.DeviceIdentifier]; skipped {
+			slog.Warn("curtailment reconciler: restore device filter-skipped",
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in c9379e36: restore dispatch now gates target state updates on `CommandResult.DispatchedDeviceIdentifiers`, and the command service now reports only identifiers that resolved to device IDs and were enqueued. Targets missing from the dispatched set stay `Pending` with a retryable error instead of incorrectly blocking the in-flight restore gate. Added coverage for the partial-enqueue path.

server/sqlc/queries/curtailment.sql
```diff
@@ -146,6 +146,15 @@ FROM curtailment_event
 WHERE event_uuid = sqlc.arg('event_uuid')
     AND org_id = sqlc.arg('org_id');
 
+-- name: GetActiveCurtailmentEvent :one
+-- Org-scoped recovery path for pending/active/restoring events.
+SELECT *
+FROM curtailment_event
+WHERE org_id = sqlc.arg('org_id')
+    AND state IN ('pending', 'active', 'restoring')
+ORDER BY id DESC
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in c61eae35: `GetActiveCurtailmentEvent` now orders `active`/`restoring` events ahead of `pending` events, then breaks ties by newest id. This keeps `GetActiveCurtailment` focused on the event currently affecting miners when a newer pending event also exists. Regenerated sqlc output is included.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -201,7 +217,9 @@ func (r *Reconciler) runTick(ctx context.Context) {
 	}
 
 	for _, ev := range events {
-		r.processEvent(ctx, ev)
+		eventCtx, eventCancel := context.WithTimeout(ctx, 2*r.cfg.TickInterval)
+		r.processEvent(eventCtx, ev)
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in 7c6359fa: `runTick` now creates a tick-scoped `tickCtx` with the existing `2*TickInterval` budget and derives each event context from that tick context, breaking out when the tick budget expires. This keeps per-event isolation while preserving bounded tick runtime; `TestReconciler_RunTickStopsWhenTickBudgetExpires` covers the regression.

server/internal/handlers/curtailment/translate.go
```diff
@@ -477,3 +480,147 @@ func formatExclusionCounters(d *modes.InsufficientLoadDetail) string {
...
+	return &pb.StopCurtailmentResponse{Event: toEventProto(event)}
+}
+
+// toEventProto maps a persisted event row to the wire CurtailmentEvent.
+// Scope, mode_params, target rollup, and decision_snapshot echo land with
+// the read APIs where the trimmed-snapshot shape is settled.
+func toEventProto(event *models.Event) *pb.CurtailmentEvent {
+	out := &pb.CurtailmentEvent{
+		EventUuid: event.EventUUID.String(),
+		State:     eventStateProto(event.State),
```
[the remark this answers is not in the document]
rongxin-liu: Addressed in 7c6359fa: `GetActiveCurtailment` and `StopCurtailment` now hydrate event payloads with current targets, per-state rollups, scope, mode params, and decision snapshot. Handler tests assert target progress and rollup fields in both active and stop responses.

server/internal/domain/curtailment/models/models.go
```diff
@@ -229,6 +238,25 @@ type InsertTargetParams struct {
...
+			n++
+		default:
+			n++
```
ankitgoswami: could just remove the second case, no?

server/internal/domain/curtailment/service.go
```diff
@@ -131,6 +133,37 @@ func (s *Service) Start(ctx context.Context, req StartRequest) (*Plan, error) {
 		v := orgConfig.MaxDurationDefaultSec
 		req.MaxDurationSeconds = &v
 	}
+	if req.MaxDurationSeconds != nil {
```
ankitgoswami: similar checks also got added to validateStartRequest, so probably not needed here? lets try and have these validations in one place.

proto/curtailment/v1/curtailment.proto
```diff
@@ -472,9 +482,12 @@ message StopCurtailmentRequest {
...
     // Admin-only override of the adaptive batch size computed at restore
-    // start. Takes precedence over any prior UpdateCurtailmentEvent value.
-    // Upper bound is a sanity ceiling; zero means "use default" so it is excluded.
-    optional uint32 restore_batch_size_override = 2 [(buf.validate.field).uint32 = {gte: 1, lte: 10000}];
+    // start. Takes precedence over the adaptive [10, 100] formula. Upper
+    // bound 200 is a real fleet-class ceiling: 2x the inrush-safe upper
+    // band, enough to short-circuit staggering for a single-rack restore
+    // but narrow enough that a typo cannot dump 10K miners at once.
+    // Zero means "use default" so it is excluded by gte=1.
+    optional uint32 restore_batch_size_override = 2 [(buf.validate.field).uint32 = {gte: 1, lte: 200}];
```
ankitgoswami: I'm not entirely sure why we need this? If I understand the code correctly, this is when the operator goes "I want to restore faster than I planned". Is that a requirement worth the added complexity (and race condition)?

server/internal/domain/curtailment/service.go
```diff
@@ -652,6 +739,136 @@ func marshalScopeJSON(s Scope) ([]byte, error) {
...
+// Adaptive batch-sizing constants. [10, 100] is the inrush envelope; the
+// admin override accepts [1, 200] (proto enforces the ceiling).
+const (
+	minBatchSizeFloor                  int32 = 10
+	maxBatchSizeCeiling                int32 = 100
+	restoreBatchSizeOverrideUpperBound int32 = 200
+)
+
+// Stop transitions a non-terminal event to `restoring`, stamps
+// effective_batch_size from the adaptive formula (or the operator override),
```
ankitgoswami: if we remove the operator override: https://github.com/block/proto-fleet/pull/232/changes#r3277657028
we could calculate effective_batch_size in start instead of stop. I think that would make things simpler. And we won't have it starting out as null so that could improve predictability. The calculations won't be that different, and small differences in batch sizes don't matter imo.

server/sqlc/queries/curtailment.sql
```diff
@@ -146,6 +146,20 @@ FROM curtailment_event
...
+SELECT *
+FROM curtailment_event
+WHERE org_id = sqlc.arg('org_id')
+    AND state IN ('pending', 'active', 'restoring')
+ORDER BY CASE
+        WHEN state IN ('active', 'restoring') THEN 0
+        ELSE 1
+    END,
+    id DESC
+LIMIT 1;
```
ankitgoswami, line 160: seems smelly - do we need a partial unique index?:

```
CREATE UNIQUE INDEX uq_curtailment_event_one_non_terminal_per_org
ON curtailment_event (org_id)
WHERE state IN ('pending', 'active', 'restoring');

```

server/internal/domain/curtailment/service.go
```diff
@@ -652,6 +739,136 @@ func marshalScopeJSON(s Scope) ([]byte, error) {
...
+		}
+	}
+	return nil
+}
+
+// checkMinCurtailedDurationGate enforces `min_curtailed_duration_sec` on
+// active events. Pending events haven't curtailed anything yet, so the
+// hysteresis gate doesn't apply. EMERGENCY priority on the event bypasses.
+func checkMinCurtailedDurationGate(event *models.Event, now time.Time) error {
+	if event.Priority == models.PriorityEmergency {
```
ankitgoswami: Priority is set at start time so it's less useful to use it to bypass this check at stop time I think? maybe we need an admin gated force field in StopCurtailmentRequest for this purpose

server/internal/handlers/curtailment/translate.go
```diff
@@ -477,3 +490,314 @@ func formatExclusionCounters(d *modes.InsufficientLoadDetail) string {
...
+	switch s {
+	case models.EventStatePending:
+		return pb.CurtailmentEventState_CURTAILMENT_EVENT_STATE_PENDING
+	case models.EventStateActive:
+		return pb.CurtailmentEventState_CURTAILMENT_EVENT_STATE_ACTIVE
+	case models.EventStateRestoring:
+		return pb.CurtailmentEventState_CURTAILMENT_EVENT_STATE_RESTORING
+	case models.EventStateCompleted:
+		return pb.CurtailmentEventState_CURTAILMENT_EVENT_STATE_COMPLETED
+	case models.EventStateCompletedWithFailures:
```
ankitgoswami, line 715: nit: could have a single EventStateCompleted with an accompanying restore_failed_count - more informative and avoids using the state to encode the fact that some things failed.

ankitgoswami (approved):
left some comments but not 100% sure about the requirements so don't know if they are blocking.

server/internal/domain/curtailment/models/models.go
rongxin-liu: Fixed in f4cd9934.

server/internal/domain/curtailment/service.go
rongxin-liu: Fixed in f4cd9934.

proto/curtailment/v1/curtailment.proto
rongxin-liu: Fixed in 1e5cf13d — `restore_batch_size_override` removed; `effective_batch_size` now computed at Start.

server/internal/domain/curtailment/service.go
rongxin-liu: Fixed in 1e5cf13d.

server/sqlc/queries/curtailment.sql
rongxin-liu: Fixed in 0a4e7a2f — partial unique index added to migration 000051; Service.Start maps the unique-violation to AlreadyExists.

server/internal/domain/curtailment/service.go
rongxin-liu: Fixed in daa3ee48 — added admin-gated `force` field on StopCurtailmentRequest; EMERGENCY-priority no longer bypasses min_curtailed_duration_sec on Stop.

server/internal/handlers/curtailment/translate.go
rongxin-liu: Keeping dual COMPLETED / COMPLETED_WITH_FAILURES state per product call — operators see failure status directly in the event-list column without needing to read a count.

server/internal/domain/curtailment/service.go
```diff
@@ -131,6 +134,39 @@ func (s *Service) Start(ctx context.Context, req StartRequest) (*Plan, error) {
 		v := orgConfig.MaxDurationDefaultSec
 		req.MaxDurationSeconds = &v
 	}
+	// Upper-bound (`<= maxFiniteDurationSeconds`) is enforced by
+	// validateStartRequest. The admin-gate below is intrinsically
+	// post-normalization: it compares the resolved value to the org default.
+	if req.MaxDurationSeconds != nil &&
```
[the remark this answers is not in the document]
rongxin-liu: Fixed in 35cc67c4.

proto/curtailment/v1/curtailment.proto
```diff
@@ -469,12 +483,17 @@ message UpdateCurtailmentEventResponse {
...
     string event_uuid = 1 [(buf.validate.field).string.min_len = 1];
 
-    // Admin-only override of the adaptive batch size computed at restore
-    // start. Takes precedence over any prior UpdateCurtailmentEvent value.
-    // Upper bound is a sanity ceiling; zero means "use default" so it is excluded.
-    optional uint32 restore_batch_size_override = 2 [(buf.validate.field).uint32 = {gte: 1, lte: 10000}];
+    // Admin-only bypass of `min_curtailed_duration_sec` on active events.
+    // Defaults to false; admin-gated handler-side. Zero/unset means the
+    // hysteresis gate applies normally.
+    bool force = 2;
```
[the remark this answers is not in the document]
rongxin-liu: Fixed in 78ce3017.

proto/curtailment/v1/curtailment.proto
```diff
@@ -469,12 +483,17 @@ message UpdateCurtailmentEventResponse {
 
 // StopCurtailment starts restore; min duration may block normal priority.
 message StopCurtailmentRequest {
+    // Field 2 previously carried `optional uint32 restore_batch_size_override`
+    // (pre-release; per-Stop override removed in favor of Start-time stamping).
+    // Reserve the old name so a future re-add never collides.
+    reserved "restore_batch_size_override";
+
```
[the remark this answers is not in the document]
rongxin-liu: Already fixed in 78ce3017 (your review was queued against 4549263f, before that commit landed).

PR #234, opened 2026-05-14
fix(fleetnode): drop http.Client.Timeout on authenticated gateway client

ankitgoswami:
## Summary

The authenticated gateway client introduced in #227 reuses \`newGatewayHTTPClient()\` whose \`http.Client.Timeout: 30s\` covers the *entire* request lifecycle (including reading the response body). For long-lived streaming RPCs that's the streamed message channel itself — so any \`ControlStream\` / \`UploadTelemetry\` / \`UploadEvents\` open via this client would be forcibly torn down at the 30-second mark.

Latent today (\`fleetnode run\` only does the unary \`UploadHeartbeat\`, which finishes well under 30s), but the next PR that wires up any streaming RPC would have discovered it immediately. Fixing the foundation here keeps that PR focused on its actual feature.

## Changes

- **New \`newStreamingGatewayHTTPClient()\`** in \`authclient.go\`: no top-level \`Timeout\`; keeps the no-redirect \`CheckRedirect\`. Used by \`NewAuthenticatedGatewayClient\`.
- **Unchanged**: \`newGatewayHTTPClient()\` (used by bootstrap's unary \`Register\` / \`BeginAuthHandshake\` / \`CompleteAuthHandshake\`). Those calls are bounded and the 30s circuit-breaker is correct for the interactive enroll/refresh flows.
- **Daemon**: \`sendHeartbeat\` now wraps each call in \`context.WithTimeout(heartbeatRequestTimeout)\`. Preserves the \"single hung heartbeat doesn't pin the daemon\" guarantee that the client-level \`Timeout\` used to provide, but applies it only to unary calls; future streaming methods will pass the loop ctx without an artificial deadline.
- **Regression test** pins that \`newStreamingGatewayHTTPClient().Timeout == 0\` and that \`CheckRedirect\` still refuses redirects.

## Test plan

- [x] \`PATH=./bin:\$PATH go build ./server/...\`
- [x] \`PATH=./bin:\$PATH go vet ./server/...\`
- [x] \`cd server && PATH=../bin:\$PATH golangci-lint run -c .golangci.yaml ./...\` — 0 issues
- [x] \`DB_PASSWORD=fleet PATH=./bin:\$PATH go test ./server/... -count=1\` — 68 packages OK
- [ ] CI runs the full suite

### DIFF
--- a/server/cmd/fleetnode/run.go
+++ b/server/cmd/fleetnode/run.go
@@ -22,6 +22,7 @@ import (
 const (
 	defaultHeartbeatInterval = 30 * time.Second
 	sessionRefreshLeeway     = 1 * time.Hour
+	heartbeatRequestTimeout  = 30 * time.Second
 )
 
 type RunCmd struct {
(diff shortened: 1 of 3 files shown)


ankitgoswami:
Closing in favor of bundling the streaming-friendly HTTP client into the first PR that actually uses streaming. Commit b0d5e1a5 stays on the `ankitg/fleetnode-streaming-timeout` branch for cherry-pick when needed.

PR #278, opened 2026-05-20
server: migrate notifications stack to grafana

illegalprime:
Migrate notifications stack to Grafana instead of VM, VMAlert, & Alertmanager.


deployment-files/run-fleet.sh
```diff
@@ -542,7 +559,50 @@ if [ "$ENABLE_BETA_NOTIFICATIONS" = "true" ]; then
         echo "Error: --enable-beta-notifications was passed but $COMPOSE_NOTIFICATIONS_FILE is missing."
         exit 1
     fi
-    echo "Notifications stack: enabled (otel-collector, victoria-metrics, vmalert, alertmanager)"
+
+    # The Grafana sidecar runs the alerting engine + UI; give it a
+    # rotated admin password the first time we boot the stack so the
+    # default "admin / admin" never ships.
```
ankitgoswami, line 565: 👍

server/migrations/000052_notification_metric_samples.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS notification_metric_sample;
```
ankitgoswami: does this also need remove_retention_policy + remove_compression_policy?

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+
+// alertmanagerPayload mirrors the Prometheus Alertmanager v4 webhook
+// envelope. Only the fields we use downstream are decoded.
+type alertmanagerPayload struct {
+	Status string              `json:"status"`
+	Alerts []alertmanagerAlert `json:"alerts"`
+}
+
+// alertmanagerAlert is one alert inside the batch. StartsAt/EndsAt are
+// kept as strings because some Alertmanager-compatible senders emit
```
ankitgoswami: units are actually time.Time here

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+
+	var payload alertmanagerPayload
+	if err := json.Unmarshal(body, &payload); err != nil {
+		writeError(w, http.StatusBadRequest, "invalid JSON payload")
+		return
+	}
+
+	if len(payload.Alerts) == 0 {
+		// An empty batch is well-formed but uninteresting; ack and
+		// return so Grafana doesn't retry.
```
ankitgoswami: AI has def gone a bit overboard with comments in this PR lol

ankitgoswami: Can you have it remove any comments that explain "what" the code is doing as opposed to "why"?

server/internal/domain/stores/sqlstores/organization.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func NewSQLOrganizationStore(conn *sql.DB) *SQLOrganizationStore {
+	return &SQLOrganizationStore{
+		SQLConnectionManager: NewSQLConnectionManager(conn),
+	}
+}
+
+func (s *SQLOrganizationStore) ListActiveOrganizationIDs(ctx context.Context) ([]int64, error) {
+	orgs, err := s.GetQueries(ctx).ListOrganizations(ctx)
```
ankitgoswami: we should probably update ListOrganizations to check DeletedAt instead of this

deployment-files/README.md
```diff
@@ -93,42 +93,57 @@ The script will auto-detect existing certificates and use HTTPS mode automatical
```
ankitgoswami, line 93: the grafana provisioning bundle living under deployment-files/server/monitoring/grafana/ is not a good idea. it's product code (rules are tightly coupled to the migration schema + webhook contract), it should be at server/monitoring/grafana/ or something.

deployment-files/run-fleet.sh
```diff
@@ -671,6 +733,182 @@ if ! docker compose "${COMPOSE_FILES[@]}" up --remove-orphans -d --wait --wait-t
...
 
+# ----------------------------------------------------------------------------
+# Grafana Read-Only DB Role Provisioning
+# ----------------------------------------------------------------------------
+
+# Create or rotate the dedicated `grafana_ro` DB role Grafana uses to
+# query notification_metric_sample. We do this here (not in a SQL
+# migration) so the password never has to be committed to source and
+# can be rotated just by editing $ENV_FILE and re-running this script.
+provision_grafana_db_role() {
```
ankitgoswami, line 744: isn't it a problem that this won't run in dev?

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+	// Unscoped alert — fan out only when this is a known global
+	// self-monitoring rule. Any other unscoped alert keeps the historic
+	// behaviour of landing as a single org-less row so we don't silently
+	// duplicate device-scoped or auth-style events that just happen to
+	// arrive without a label.
+	if !isGlobalSelfMonitoringAlert(alert.Labels) {
+		return 1, h.insertEvent(ctx, event, alert)
+	}
+
+	orgIDs := h.activeOrgIDsForFanOut(ctx, alert)
```
ankitgoswami: can optimize this by calling before the alert loop once and passing ids into persistAlert

server/cmd/fleetd/main.go
```diff
@@ -469,6 +471,10 @@ func start(config *Config) error {
 	mux := http.NewServeMux()
 
 	mux.HandleFunc("/health", health.NewHandler())
+	if config.Metrics.WebhookToken == "" {
+		slog.Warn("FLEET_METRICS_WEBHOOK_TOKEN is not set; alertmanager webhook will reject every delivery")
+	}
+	mux.Handle("POST "+alertmanagerwebhook.Path, alertmanagerwebhook.NewHandler(activitySvc, config.Metrics.WebhookToken, organizationStore))
```
ankitgoswami: we should gate this on FLEET_METRICS_ENABLED right?

server/migrations/000052_notification_metric_samples.down.sql
ankitgoswami: also will this work without CASCADE?

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+		// Copy the event per iteration and take the address of the
+		// slice element directly: models.Event holds OrganizationID
+		// by pointer, and pointing at a loop variable risks every
+		// row sharing the same address. The slice backs the pointers
+		// with distinct memory for every element, so the per-row
+		// OrganizationID is unambiguous if anyone later inspects the
+		// captured events (the test rig asserts this distinctness).
+		scoped := event
+		scoped.OrganizationID = &orgIDs[i]
+		persisted += h.insertEvent(ctx, scoped, alert)
```
ankitgoswami: should set a timeout on these

deployment-files/server/monitoring/grafana/provisioning/alerting/proto-fleet-rules.yaml
```diff
@@ -0,0 +1,209 @@
...
+                GROUP BY organization_id, device_id
+                HAVING last(value, time) = 0
+          - refId: B
+            datasourceUid: __expr__
+            model:
+              refId: B
+              type: math
+              expression: $A
+        noDataState: OK
+        execErrState: Alerting
```
ankitgoswami: consider setting this to Error, else the noise from these alerts would drone out something like the DB being down

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+	slog.Debug("alertmanager webhook delivered",
+		"alerts", len(payload.Alerts),
+		"persisted", persisted,
+		"truncated", truncated,
+		"status", payload.Status,
+	)
+
+	// If nothing landed — typically a DB outage or activity-log
+	// rejection affecting every alert in the batch — return 5xx so
+	// Grafana retries the delivery. Acking 204 here would let a
```
ankitgoswami: this would lead to duplicates right? maybe we should add a unique partial index for alert.* event_types

server/docker-compose.base.yaml
```diff
@@ -16,10 +16,8 @@ services:
       FLEET_COMMAND_BATCH_STATUS_UPDATE_POLLING_INTERVAL: "1s"
       FLEET_QUEUE_DEQUE_LIMIT: "500"
       FLEET_QUEUE_MAX_FAILURE_RETRIES: "5"
-      OTEL_EXPORTER_OTLP_ENDPOINT: "http://otel-collector:4317"
-      FLEET_VMALERT_URL: "http://vmalert:8880"
-      FLEET_ALERTMANAGER_URL: "http://alertmanager:9093"
       FLEET_TELEMETRY_ENDPOINT: "http://otel-collector:4318"
+      FLEET_GRAFANA_URL: "http://grafana:3000"
```
ankitgoswami: looks like this is unused?

server/docker-compose.notifications.yaml
```diff
@@ -1,20 +1,21 @@
+# Dev-mode notifications overlay used by `just dev-notifs`.
+#
+# Mirrors deployment-files/docker-compose.notifications.yaml but mounts
+# the in-tree provisioning bundle directly so edits to the YAML files
+# are picked up on container restart without a release rebuild.
 services:
   fleet-api:
     environment:
       FLEET_TELEMETRY_ENABLED: "${FLEET_TELEMETRY_ENABLED:-true}"
```
ankitgoswami, line 9: this was dropped in another file, should it be removed from here too?

server/docker-compose.base.yaml
ankitgoswami, line 143: can probably remove this now?

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+	if status == "" {
+		status = statusFiring
+	}
+
+	description := alert.Annotations["summary"]
+	if description == "" {
+		description = fmt.Sprintf("alert %s %s", alertName, status)
+	}
+
+	result := models.ResultFailure
```
ankitgoswami: The Result field shouldn't be encoding alert state imo. Result means "did the operation this row audits succeed?". it describes the outcome of the recorded action and here the recorded action is the alert firing which was successful. 

cc: @rongxin-liu

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+	if status == statusResolved {
+		result = models.ResultSuccess
+	}
+
+	scopeType := alert.Labels[labelTemplate]
```
ankitgoswami: this also seems like we're repurposing it to store something different

server/internal/handlers/alertmanagerwebhook/handler.go
ankitgoswami: I think we're overloading activity events for something they weren't designed for - maybe we just need a alert_history table instead of this

server/internal/infrastructure/metrics/provider.go
```diff
@@ -2,162 +2,441 @@ package metrics
...
-	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
 )
 
 const ServiceName = "proto-fleet-api"
 
+const maxPostgresBindParameters = 65535
+
+var maxSamplesPerInsert = maxPostgresBindParameters / columnsPerSample
+
 type Config struct {
```
ankitgoswami, line 20: The flushLoop + pendingRetry + nextBackoff + appendBoundedRetry retry layer seems over-engineered now that we're simply doing a local db insert. Or am I missing something?

server/internal/infrastructure/metrics/store.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return params
+}
+
+// Close is a no-op — the *sql.DB is owned by the caller.
+func (s *sqlStore) Close() error { return nil }
+
+// inMemoryStore is the test double exposed to provider_test.go. It
+// records every sample handed to InsertSamples so tests can assert on
+// metric names and label sets without a real TimescaleDB.
```
ankitgoswami: we could drop this and use a real DB in tests. Would lead to better tests and less abstraction. not a blocker though.

server/internal/handlers/alertmanagerwebhook/handler.go
```diff
@@ -0,0 +1,446 @@
...
+// default rules is a few kilobytes. The limit is generous (so a bursty
+// grouped delivery still fits) but bounded so a malformed sender can't
+// exhaust memory.
+const maxBodyBytes = 1 << 20 // 1 MiB
+
+// maxAlertsPerRequest caps the number of alerts a single webhook delivery may carry.
+const maxAlertsPerRequest = 100
+
+// maxRowsPerRequest caps the total number of activity_log inserts a single webhook delivery may issue.
+const maxRowsPerRequest = 1000
```
ankitgoswami, line 28: Maybe I'm missing something but all these + maxSamplesPerInsert are kinda just defending the same surface no? Lets just pick one or two?

ankitgoswami (dismissed):
Great job figuring this out! I think there are some architectural things to sort out but the stack and integrations look like they would work well for our needs ❤️

server/internal/infrastructure/metrics/emit.go
flesher, line 19: ditto

flesher, line 11: we could add SiteID

server/internal/infrastructure/metrics/emit.go
```diff
@@ -29,110 +26,110 @@ type TelemetryPollLabels struct {
 	Result         string
```
flesher, line 29: ditto

flesher (commented):
Phase 1 work for multi site included site-stamped hypertables for: `errors`, `miner_state_snapshots`, `device_metrics`(migration 000047).  We should probably include SiteID with `notification_metric_sample` too.

server/internal/handlers/alertmanagerwebhook/handler.go
illegalprime: If `persisted == 0` then nothing hit the DB, so no duplicates?

deployment-files/run-fleet.sh
illegalprime: In dev we just use the main creds.

server/docker-compose.base.yaml
illegalprime: yup, leftover from a rebase off the UI branch

server/docker-compose.notifications.yaml
illegalprime: A bit confusing but this is our internal metrics that's only enabled in dev atm. It makes sense to move it back to  the dev-only compose too.

server/docker-compose.base.yaml
illegalprime: It's still in use by grafana, otel, and jaeger right?

mcharles-square (approved):

server/internal/infrastructure/metrics/provider.go
illegalprime: TBD next time

server/internal/handlers/alertmanagerwebhook/handler.go
illegalprime: TBD next time

PR #283, opened 2026-05-21
feat(rbac): foundation — catalog, schema, sqlc, seed, backfill (PR 1 of 4)

mcharles-square:
First PR of the four-part [granular RBAC plan](https://github.com/block/proto-fleet/pull/282) (#282 — the plan PR, updated alongside this branch as the design evolved).

This is the foundation cut: schema, permission catalog, sqlc bindings, seed migration with per-org built-in roles, startup reconciliation, the user_organization backfill, and the onboarding dual-write. **The existing `RequireAdmin` middleware stays in place and continues to be the live authorization path.** PR 2 will swap callers to a permission-based gate; until that lands, this PR is purely additive — the new tables are populated and the resolver inputs are ready, but no code path reads them yet.

### Schema — migration 000052 (`server/migrations/000052_create_permission_tables.{up,down}.sql`)

- `permission`, `role_permission`, and `user_organization_role` tables.
- `ALTER TABLE role` adds `is_builtin`, `builtin_key`, and `organization_id` — **every role is now owned by an organization** (built-in or custom). The legacy global `uq_role_name UNIQUE(name)` from migration 000002 is dropped.
- **Two partial unique indexes on `role`** (live rows only):
  - `uq_role_org_builtin_key` — per-org built-in identity, keyed on `(organization_id, builtin_key)`.
  - `uq_role_org_custom_name` — per-org custom uniqueness keyed on `(organization_id, LOWER(BTRIM(name)))`. Case-insensitive and whitespace-trim-tolerant: `"Admin"`, `"admin"`, and `" ADMIN "` collapse to one row.
- **Two partial unique indexes on `user_organization_role`** (NULL-safe, live-row scoped):
  - `uq_user_org_role_org_scope` for `scope_type = 'org'` (omits `scope_id` to dodge Postgres's NULL-distinct trap).
  - `uq_user_org_role_site_scope` for `scope_type = 'site'`.
  - Both predicated on `deleted_at IS NULL` so re-assignment after `UnassignRole` is allowed.
- **Composite FK from `user_organization_role(role_id, organization_id)` → `role(id, organization_id)`** (`UNIQUE (id, organization_id)` on `role` is the target). An assignment row can no longer reference a role from a different org — DB-enforced tenant isolation.
- **Composite FK to `site(id, org_id)` for site-scoped assignments**, `DEFERRABLE INITIALLY DEFERRED` for transactional reassignments.
- **CHECK constraints**:
  - `chk_role_builtin_key_matches_flag` — `is_builtin = TRUE ↔ builtin_key IS NOT NULL`.
  - `chk_role_custom_name_not_reserved` — custom rows cannot be named `SUPER_ADMIN`, `ADMIN`, or `FIELD_TECH` (case-insensitive). Legacy global rows (`organization_id IS NULL`) are exempt so existing deployments don't fail the validation on upgrade.
  - `chk_user_org_role_scope_id_matches_type` — org-scope ⇒ `scope_id IS NULL`; site-scope ⇒ `scope_id IS NOT NULL`.

### Permission catalog — `server/internal/domain/authz/catalog.go` + `proto/authz/v1/authz.proto`

29 permission keys across 10 resource groups (`fleet`, `miner`, `rack`, `site`, `serverlog`, `curtailment`, `fleetnode`, `apikey`, `user`, `role`). Source of truth for everything downstream. Tests assert catalog completeness, key shape (`^[a-z]+:[a-z_]+$`), and resource grouping.

### sqlc queries (`server/sqlc/queries/`)

Org-aware throughout — every role-touching query takes `organization_id`. New / changed:

- `permission.sql` — list/get/upsert catalog + role-permission insert/delete + `PrunePermissionsOutsideKeys` (used by SUPER_ADMIN full-reconcile only).
- `role.sql` — `ListBuiltinRolesForOrg`, `GetBuiltinRoleForOrg`, `UpsertBuiltinRoleForOrg`, `ListActiveOrganizationIDs`, `ListCustomRolesForOrg`, `CreateCustomRole`, `UpsertCustomRoleForOrg`, `UpdateCustomRoleName`, `SoftDeleteCustomRole`. The legacy `UpsertRole(name, description)` was retargeted to `UpsertCustomRoleForOrg(name, description, organization_id)` and its `ON CONFLICT` matches the new partial index expression.
- `user_organization_role.sql` — `AssignRole`, `UnassignRole`, `ListAssignmentsForUser`, `ListAssignmentsForRole`, `ListEffectivePermissionsForUser` (single-query resolver source), and the last-org-scope-SUPER_ADMIN guards used by the future delete/deactivate paths.
- All cross-table reads also assert `r.organization_id = uor.organization_id` as belt-and-suspenders defense in depth.

### Seed + per-org reconciliation — migration 000053 + `server/internal/domain/authz/{builtin,reconcile}.go`

- 000053 seeds the catalog and **one set of three built-in role rows per active organization** by cross-joining `organization × built-in spec` (including soft-deleted orgs, since membership doesn't cascade). Existing `user_organization` rows that pointed at the legacy global ADMIN/SUPER_ADMIN rows are repointed to each user's per-org replacement; the legacy globals are then soft-deleted.
- `Reconcile(ctx, db)` runs at boot from `cmd/fleetd/main.go` inside a `pg_advisory_xact_lock`'d transaction. It iterates `ListActiveOrganizationIDs` and calls `SeedOrgBuiltins(ctx, q, orgID)` for each.
- **Per-builtin policy:**
  - `SUPER_ADMIN` — fully reconciled to `AllPermissions()` per org. Tampering is repaired on every boot.
  - `ADMIN` / `FIELD_TECH` — **seeded once on creation; left alone after that.** When a per-org row is first inserted, every permission in the seed formula is attached. Subsequent boots leave its `role_permission` set untouched. The operator owns the role after creation. Re-asserting the seed set every boot would have silently restored permissions the operator deliberately revoked (e.g., `miner:update_pools`, `miner:firmware_update`) — exactly the bug Codex caught in round 1.
- `SeedOrgBuiltins` is also called from the onboarding flow (see below) so a brand-new org gets its built-ins in the same transaction as org creation.

### Backfill — migration 000054 + onboarding dual-write

- 000054 mirrors every active `user_organization` row into `user_organization_role` as an org-scope assignment, after 000053 has repointed legacy `role_id` references to per-org rows. Soft-deleted user_organization rows are skipped.
- **Onboarding dual-write**: `CreateAdminUserWithOrganization` (founding user, in `sqlstores/user.go`) and `CreateUserOrganizationRole` (subsequent users) now write **both** `user_organization` AND `user_organization_role` inside the same transaction. Without this, fresh installs would have a founding SUPER_ADMIN with no row in the assignment table, and the resolver would deny everything once PR 2 starts consuming it.
- `auth.Service.CreateUser` swapped from `GetRoleByName("ADMIN")` (which would return an arbitrary per-org ADMIN row across orgs) to `GetBuiltinRoleForOrg(orgID, AdminRoleName)`.

## Plan deviation (still in force)

The plan's U5 also renamed `user_organization.role_id` to `role_id_deprecated_do_not_use` and added a raising trigger on non-NULL writes. **That part is moved to PR 2's opening migration** so it lands together with the caller swap. Doing it in PR 1 would break the existing onboarding flow the moment this deploys. The plan's intent — loud failure during the soak window — is preserved; the soak boundary just shifts from "between PR 1 and U12" to "between PR 2 and U12". Noted inline in `000054_backfill_user_organization_to_assignments.up.sql`.

## What this PR does NOT do

- Does not add the `RequirePermission` middleware or resolver — PR 2.
- Does not swap any existing `RequireAdmin` callers — PR 2.
- Does not rename `user_organization.role_id` or add the raising trigger — PR 2.
- Does not add role/assignment management RPCs — PR 3.
- Does not change `UserInfo.permissions` — PR 3.
- Does not touch the React client — PR 3.
- Does not drop the legacy `user_organization.role_id` column — PR 4.

## Review history baked in

Four rounds of automated review (Codex Security + Codex P-tier + Copilot) drove substantial design changes captured above. Headline outcomes:

- **Cross-tenant authorization bleed closed.** The original design had global built-in rows that a SUPER_ADMIN editing ADMIN would affect across every org. Switched to per-org built-ins.
- **NULL-distinct assignment-uniqueness loophole closed.** Replaced a naive `UNIQUE (… , scope_id)` with two partial unique indexes; soft-delete no longer blocks re-assignment.
- **Cross-tenant role-assignment loophole closed.** Composite FK from `user_organization_role` to `role(id, organization_id)` makes a cross-org pointer structurally impossible.
- **Editable-built-in footgun closed.** Reconciliation moved from "re-assert seed set every boot" (which silently restored revoked permissions) to "seed on first creation, leave alone after."
- **Reserved-name collisions closed.** `chk_role_custom_name_not_reserved` is case-insensitive and whitespace-trim-tolerant; the partial unique index for customs is case-folded so `"Admin"`/`"admin"` collapse to one row per org.
- **Rollback safety.** 000054 down hard-deletes mirrored assignments (was soft-delete, which held the FK and blocked role deletes); 000052 down's custom-cleanup runs before any table or column drops; legacy global rows are preserved through the rollback so `user_organization.role_id` doesn't dangle.
- **Soft-deleted orgs get built-ins too** so the 000053 repoint and 000054 backfill produce FK-valid rows even when a soft-deleted org has active memberships.

All review threads from the four rounds are resolved on the PR.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `lefthook run pre-commit` and `lefthook run pre-push` (golangci-lint + client-typecheck) both green
- [x] `just gen` clean — sqlc statically parses the migrations and produces compiling bindings; proto generation produces `authz/v1/authz_pb.ts` and `authz/v1/authz.pb.go`
- [x] Non-DB tests pass: `go test ./internal/domain/authz/... -run 'TestAllPermissions|TestCatalog|TestLookup|TestResourceOrder'`
- [ ] **DB-dependent tests need Docker** — `just dev` to bring up Postgres, then `go test ./internal/domain/authz/...`. Covers: fresh-install seeding (`TestReconcile_FreshInstall_*`), idempotency, per-org isolation (`TestReconcile_PerOrgIsolation_*`), seed-on-create semantics (`TestReconcile_OperatorRevokedSensitivePermStaysRevokedOnRestart`, `TestReconcile_NewCatalogPermissionDoesNotPropagateToExistingBuiltins`), SUPER_ADMIN tamper repair, NULL-safe assignment uniqueness (`TestAssignment_DuplicateLiveOrgScopeRejected`, `TestAssignment_SoftDeletedRowDoesNotBlockReassign`), backfill behavior, soft-deleted-org seeding (`TestSeed_SoftDeletedOrgStillGetsBuiltins`), onboarding dual-write (`TestOnboarding_FoundingUserGetsOrgScopeSuperAdminAssignment`), reserved built-in names case-insensitive (`TestCustomRole_ReservedBuiltinNamesRejectedCaseInsensitively`), case-insensitive per-org custom uniqueness (`TestCustomRole_CaseInsensitiveUniquenessPerOrg`), and same-name-across-orgs-allowed (`TestCustomRole_SameNameAcrossOrgsAllowed`).
- [ ] **Migration smoke** — `just dev` applies 000052–000054 cleanly; `psql` confirms per-org built-in role rows exist with the expected permission counts and all CHECK / FK constraints are in place.

## Rollback

Each migration has a paired `*.down.sql`. The reconciler is invoked once at boot; reverting this PR removes that call. Down migrations were rewritten across the review rounds to be order-correct (custom cleanup before any DROPs, table-existence dependencies respected) and to preserve the legacy global ADMIN/SUPER_ADMIN rows that downstream rollback steps reference. 000054 hard-deletes the mirrored assignment rows it created instead of soft-deleting, so the role-table down-DELETE that follows succeeds against the `ON DELETE RESTRICT` composite FK.


server/migrations/000051_create_permission_tables.up.sql
```diff
@@ -0,0 +1,92 @@
...
+    -- write "org-scope but pointing at site 42" or vice versa.
+    CONSTRAINT chk_user_org_role_scope_id_matches_type CHECK (
+        (scope_type = 'org'  AND scope_id IS NULL) OR
+        (scope_type = 'site' AND scope_id IS NOT NULL)
+    ),
+
+    -- One row per (user, role, scope) — the unique key is the structural
+    -- guarantee that re-saving the same assignment is idempotent.
+    CONSTRAINT uq_user_org_role_scope UNIQUE
+        (user_id, organization_id, role_id, scope_type, scope_id),
```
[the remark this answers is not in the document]
mcharles-square: > The `uq_user_org_role_scope` constraint does not actually enforce uniqueness for org-scope rows because PostgreSQL treats `NULL` values as distinct in unique constraints; with `scope_type='org'`, `scope_id` is always `NULL`, so duplicate `(user_id, organization_id, role_id, 'org', NULL)` rows are allowed.

Fixed: dropped `uq_user_org_role_scope` and replaced with two partial unique indexes (`uq_user_org_role_org_scope` for `scope_type='org'` and `uq_user_org_role_site_scope` for site scope), both scoped to `deleted_at IS NULL`. The org variant omits `scope_id` so the NULL-distinct trap cannot bite. New test `TestAssignment_DuplicateLiveOrgScopeRejected` locks in the behavior.

server/migrations/000051_create_permission_tables.up.sql
```diff
@@ -0,0 +1,92 @@
...
+    -- combinations are rejected at the DB layer so application bugs cannot
+    -- write "org-scope but pointing at site 42" or vice versa.
+    CONSTRAINT chk_user_org_role_scope_id_matches_type CHECK (
+        (scope_type = 'org'  AND scope_id IS NULL) OR
+        (scope_type = 'site' AND scope_id IS NOT NULL)
+    ),
+
+    -- One row per (user, role, scope) — the unique key is the structural
+    -- guarantee that re-saving the same assignment is idempotent.
+    CONSTRAINT uq_user_org_role_scope UNIQUE
```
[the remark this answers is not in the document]
mcharles-square: > `uq_user_org_role_scope` won't actually prevent duplicates for org-scope rows because `scope_id` is NULL for `scope_type='org'` and PostgreSQL unique constraints treat NULLs as distinct.

Fixed differently: rather than `NULLS NOT DISTINCT` (which would also need PG 15+), I went with two partial unique indexes scoped to live rows — `uq_user_org_role_org_scope` for `scope_type='org'` and `uq_user_org_role_site_scope` for `scope_type='site'`. Backfill `ON CONFLICT` updated to target the org partial index. Bonus: the live-only predicate also unblocks re-assignment after `UnassignRole` (the soft-deleted row no longer collides), which the Codex Security MEDIUM finding also flagged.

server/migrations/000053_seed_builtin_roles.down.sql
```diff
@@ -0,0 +1,17 @@
...
+DELETE FROM role_permission
+WHERE role_id IN (SELECT id FROM role WHERE is_builtin = TRUE);
+
+DELETE FROM role WHERE builtin_key = 'FIELD_TECH';
+
+UPDATE role
+SET is_builtin = FALSE,
+    builtin_key = NULL
+WHERE builtin_key IN ('SUPER_ADMIN', 'ADMIN');
+
```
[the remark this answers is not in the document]
mcharles-square: > This down migration deletes from `permission` without first ensuring there are no remaining `role_permission` rows referencing it. Since `role_permission.permission_id` is `ON DELETE RESTRICT` (000051), this will fail if any non-built-in roles (or any other role_permission rows) exist.

Fixed: 000052 down now does `DELETE FROM role_permission` unconditionally (not just rows attached to built-ins) before deleting from `permission`. Safe because 000051 down drops the table outright next.

server/sqlc/queries/user_organization_role.sql
```diff
@@ -0,0 +1,105 @@
+-- Multi-assignment join queries. A user can hold multiple (role,
+-- scope_type, scope_id) rows in the same organization; the resolver
+-- (U6) loads every active row for a (user, org) pair on each request.
+
+-- name: AssignRole :one
+-- Insert a single assignment. Caller is responsible for the
+-- privilege-parity check (U8) before this fires. The unique constraint
+-- uq_user_org_role_scope catches re-assignment of the same
+-- (user, role, scope_type, scope_id) and surfaces as AlreadyExists.
```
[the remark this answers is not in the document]
mcharles-square: > The comment here assumes `uq_user_org_role_scope` will catch re-assignment of the same (user, role, scope_type, scope_id). With the current schema (`scope_id` NULL for org-scope), PostgreSQL's UNIQUE semantics won't treat `(… , NULL)` as conflicting, so org-scope duplicates can slip through and `ON CONFLICT` won't dedupe.

Addressed: the constraint was replaced with two partial unique indexes (`uq_user_org_role_org_scope` and `uq_user_org_role_site_scope`), and the `AssignRole` comment now reflects the new indexes and explicitly notes that re-assigning after soft-delete is allowed because the indexes are predicated on `deleted_at IS NULL`.

server/internal/domain/authz/catalog.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// CatalogByResource groups the catalog by resource for UI display. Resource
+// order matches the catalog declaration order.
+func CatalogByResource() map[string][]CatalogEntry {
+	groups := make(map[string][]CatalogEntry)
+	for _, entry := range catalog {
+		groups[entry.Resource] = append(groups[entry.Resource], entry)
+	}
+	return groups
```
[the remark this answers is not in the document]
mcharles-square: > `CatalogByResource` returns a `map[string][]CatalogEntry`, but the comment says "Resource order matches the catalog declaration order." Map iteration order is not stable in Go, so callers can't rely on it.

Fixed: corrected the docstring to scope the order guarantee to entries within a value slice (which IS declaration-order). Added a new exported helper `ResourceOrder()` that returns the resource identifiers in declaration order, so callers that want a deterministic outer iteration can drive it from that slice and index into the map.

mcharles-square:
## Codex Security Review — response

Thanks for the careful read. Both findings are addressed in commit `0e47b16`.

### HIGH — Custom roles global across organizations

> `CreateCustomRole` inserts custom roles into the global `role` table with no `organization_id` … a user with `role:manage` in one organization can create or modify role definitions that may be assigned in other organizations.

This was a real cross-tenant bleed risk and went deeper than just custom roles — the *built-in* `ADMIN` and `FIELD_TECH` rows were also global and (per the plan) editable, so a SUPER_ADMIN in org A editing ADMIN would also affect every user holding ADMIN in org B.

**Switched the design to per-org built-ins (every role is owned by an organization):**

- `role` now carries `organization_id BIGINT NOT NULL` (legacy global rows are soft-deleted by 000052 after their references are repointed; future migrations will tighten to NOT NULL).
- Dropped `uq_role_name` (the global name unique from 000002). Replaced with two partial unique indexes on live rows:
  - `uq_role_org_builtin_key` — built-ins unique per `(organization_id, builtin_key)`.
  - `uq_role_org_custom_name` — customs unique per `(organization_id, name)`.
- `CreateCustomRole` and `ListCustomRolesForOrg` now take `organization_id`. Custom roles can never be created without an owning org and can never be listed cross-tenant.
- Built-in seeding moved into `authz.SeedOrgBuiltins`, called by both the boot reconciler (looping over every active organization) and `CreateAdminUserWithOrganization` (inside the existing org-creation transaction). A new org gets its own SUPER_ADMIN/ADMIN/FIELD_TECH rows before any user is assigned.
- 000052 migrates existing prod state by cross-joining `organization × built-in spec`, creating per-org rows, repointing `user_organization.role_id` from the legacy global rows to the per-org replacements, and soft-deleting the legacy rows.

New regression test `TestReconcile_PerOrgIsolation_EditingOneOrgDoesNotAffectAnother` proves the property.

### MEDIUM — Assignment uniqueness does not work for org-scope rows and conflicts with soft delete

> `uq_user_org_role_scope` includes nullable `scope_id`. PostgreSQL treats `NULL` values as distinct in normal unique constraints … because `deleted_at` is not part of a partial uniqueness strategy, site-scope assignments cannot be re-created after soft delete.

Fixed both halves by replacing `uq_user_org_role_scope` with two partial unique indexes scoped to live rows:

- `uq_user_org_role_org_scope` on `(user_id, organization_id, role_id) WHERE scope_type = 'org' AND deleted_at IS NULL` — omits `scope_id` entirely so the NULL-distinct trap can't bite.
- `uq_user_org_role_site_scope` on `(user_id, organization_id, role_id, scope_id) WHERE scope_type = 'site' AND deleted_at IS NULL` — soft-deleted rows leave the index.

New tests:

- `TestAssignment_DuplicateLiveOrgScopeRejected` — second live insert with the same `(user, org, role, 'org', NULL)` is rejected.
- `TestAssignment_SoftDeletedRowDoesNotBlockReassign` — `AssignRole` after `UnassignRole` of the same tuple succeeds.

The backfill `ON CONFLICT` was also retargeted to the new org partial index.

server/migrations/000053_seed_builtin_roles.up.sql
```diff
@@ -0,0 +1,139 @@
...
+-- ---------------------------------------------------------------
+-- 2. Per-org built-in role rows. Cross-join active orgs with the
+--    fixed list of built-in specs.
+-- ---------------------------------------------------------------
+INSERT INTO role (name, description, is_builtin, builtin_key, organization_id)
+SELECT b.name, b.description, TRUE, b.builtin_key, org.id
+FROM organization org
+CROSS JOIN (VALUES
+    ('SUPER_ADMIN', 'Full system access. Immutable.', 'SUPER_ADMIN'),
+    ('ADMIN',       'Org admin. Editable by a SUPER_ADMIN.', 'ADMIN'),
```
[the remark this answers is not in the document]
mcharles-square: > This migration creates one `ADMIN` role per organization while legacy code still calls `GetRoleByName("ADMIN")` in `CreateUser` and then writes that `role_id` into `user_organization`; because that lookup is name-only, it can now return an arbitrary org's ADMIN row (or the soft-deleted legacy global ADMIN), so new users may be bound to the wrong role record for their org.

Fixed in `dcf5224`. Added `UserManagementStore.GetBuiltinRoleForOrg(orgID, builtinKey)` with a SQL query that filters on `is_builtin = TRUE AND organization_id = $1 AND builtin_key = $2 AND deleted_at IS NULL` (matches the partial unique index `uq_role_org_builtin_key`). `auth.Service.CreateUser` now uses it with the user's actual orgID instead of the global `GetRoleByName`. Soft-deleted legacy global ADMIN row is also excluded by the `deleted_at IS NULL` filter.

mcharles-square:
## Codex Security Review (round 2) — response

All three findings addressed in `dcf5224`. Plan PR #282 updated in `bac7657` to reflect the corrected semantics.

### HIGH — Editable built-in roles regain revoked permissions on restart

> `reconcileBuiltinPermissions` unconditionally reassigns every seed permission before only pruning `SUPER_ADMIN`. For `ADMIN` and `FIELD_TECH`, this means any operator-removed seed permission is silently restored on the next `fleetd` restart.

Spot-on. The original wording "additive-only" was misleading — the implementation re-asserted the seed set on every boot, which silently restored anything an operator deliberately revoked. Fixed by switching to **seed-on-create**: `seedOrgBuiltins` now calls `GetBuiltinRoleForOrg` first; if the row does not exist, INSERT and seed all permissions; if it exists, SUPER_ADMIN gets full reconcile (assign + prune) and ADMIN/FIELD_TECH get nothing. Once a per-org built-in row exists the operator owns it.

Two regression tests added:

- `TestReconcile_OperatorRevokedSensitivePermStaysRevokedOnRestart` — operator revokes `miner:update_pools`, `miner:firmware_update`, and `miner:reboot` from ADMIN; multiple reconciles leave all three revoked.
- `TestReconcile_NewCatalogPermissionDoesNotPropagateToExistingBuiltins` — inserts a new permission row directly to simulate catalog growth; reconcile leaves existing ADMIN and FIELD_TECH unchanged.

Trade-off: new catalog keys added in future releases no longer auto-propagate to existing orgs' ADMIN/FIELD_TECH. Operators add them via the role editor. That's the correct ownership boundary — "the operator owns this role" cannot also mean "the deploy can quietly widen it."

### HIGH — Role assignments can reference roles from another organization

> `user_organization_role.role_id` references only `role(id)`, not `(role_id, organization_id)`.

Fixed structurally:

- New constraint `uq_role_id_org_id UNIQUE (id, organization_id)` on `role`.
- `user_organization_role`'s FK on `role_id` replaced with a composite FK `(role_id, organization_id) → role(id, organization_id)`. An assignment can no longer point at a role owned by a different org.
- `ListEffectivePermissionsForUser`, `CountOrgScopeSuperAdminsExcludingAssignment`, and `CountOrgScopeSuperAdminsExcludingUser` now also assert `r.organization_id = uor.organization_id` in their joins as defense in depth — the FK is the structural guarantee, the join filter is a belt-and-suspenders check that ensures the resolver never even considers a cross-tenant row even if the FK were dropped in a future migration.

### MEDIUM — New users are not mirrored into the new assignment table

> Onboarding and user creation still write only to legacy `user_organization`; they do not also insert the corresponding org-scope row into `user_organization_role`. Fresh installs are especially exposed: migrations run before onboarding, so the founding `SUPER_ADMIN` gets no row in `user_organization_role`.

Both `CreateAdminUserWithOrganization` (founding user) and `CreateUserOrganizationRole` (subsequent users) now **dual-write**: legacy `user_organization` row plus an org-scope row in `user_organization_role`. Both writes happen inside the same transaction as the user/org create so a failure cleanly rolls back the whole flow.

New test `TestOnboarding_FoundingUserGetsOrgScopeSuperAdminAssignment`:

1. Calls `CreateAdminUserWithOrganization` against a fresh DB.
2. Asserts exactly one assignment row exists for the user/org.
3. Asserts the assignment is `scope_type='org'`, `scope_id IS NULL`.
4. Asserts the assigned `role_id` matches the org's per-org SUPER_ADMIN row (not a legacy global).

server/migrations/000052_create_permission_tables.up.sql
```diff
@@ -0,0 +1,141 @@
...
+
+ALTER TABLE role
+    ADD CONSTRAINT fk_role_organization FOREIGN KEY (organization_id)
+        REFERENCES organization(id) ON DELETE CASCADE;
+
+-- Global name uniqueness from migration 000002 is no longer correct
+-- under per-org roles — each org should be able to name its own
+-- "FieldTech Plus" custom role independently. Replace with partial
+-- unique indexes scoped to live rows.
+ALTER TABLE role DROP CONSTRAINT uq_role_name;
```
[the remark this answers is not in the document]
mcharles-square: > This migration drops `uq_role_name`, but `UpsertRole` in `server/sqlc/queries/role.sql` still uses `ON CONFLICT (name) DO UPDATE`; after this change, any `UpsertRole` call will fail at runtime with "no unique or exclusion constraint matching the ON CONFLICT specification."

Fixed in `2a84d63`. Replaced `UpsertRole` with `UpsertCustomRoleForOrg(name, description, organization_id)` whose `ON CONFLICT` targets the partial unique index `uq_role_org_custom_name WHERE is_builtin = FALSE AND deleted_at IS NULL`. Built-ins continue to use `UpsertBuiltinRoleForOrg` so the same query never tries to upsert a built-in row through the custom path. Updated callers: `testutil/database_queries.go` switched to `authz.SeedOrgBuiltins` for SUPER_ADMIN seeding (matches the real onboarding flow), and `apikey/handler_test.go` uses `UpsertCustomRoleForOrg` for its VIEWER custom role. Both call sites also dual-write into `user_organization_role` so test users have the assignment row the resolver will need.

server/migrations/000051_create_permission_tables.up.sql
```diff
@@ -0,0 +1,141 @@
...
+-- Multi-assignment join: a user can hold multiple (role, scope) pairs
+-- in the same organization. scope_type is 'org' (scope_id IS NULL) or
+-- 'site' (scope_id references site.id within the same organization).
+-- Building scope is deferred to a follow-up plan; when it ships, the
+-- CHECK is relaxed and a second composite FK is added.
+CREATE TABLE user_organization_role (
+    id              BIGSERIAL   PRIMARY KEY,
+    user_id         BIGINT      NOT NULL REFERENCES "user"(id) ON DELETE CASCADE,
+    organization_id BIGINT      NOT NULL REFERENCES organization(id) ON DELETE CASCADE,
+    role_id         BIGINT      NOT NULL REFERENCES role(id) ON DELETE RESTRICT,
```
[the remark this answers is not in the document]
mcharles-square: > The new `user_organization_role` schema only constrains `role_id` to `role(id)` and does not enforce that `user_organization_role.organization_id` matches `role.organization_id`. That allows cross-tenant assignments (org A row pointing at org B role)

Fixed in `dcf5224`. Added `UNIQUE (id, organization_id)` on `role` (so it can be the target of a composite FK), and replaced `user_organization_role`'s `role_id` FK with a composite FK `(role_id, organization_id) → role(id, organization_id)`. An assignment row pointing at a role from a different org is now structurally impossible. As defense in depth, `ListEffectivePermissionsForUser`, `CountOrgScopeSuperAdminsExcludingAssignment`, and `CountOrgScopeSuperAdminsExcludingUser` also assert `r.organization_id = uor.organization_id` in their joins.

server/migrations/000052_seed_builtin_roles.up.sql
```diff
@@ -0,0 +1,139 @@
...
+-- ---------------------------------------------------------------
+INSERT INTO role (name, description, is_builtin, builtin_key, organization_id)
+SELECT b.name, b.description, TRUE, b.builtin_key, org.id
+FROM organization org
+CROSS JOIN (VALUES
+    ('SUPER_ADMIN', 'Full system access. Immutable.', 'SUPER_ADMIN'),
+    ('ADMIN',       'Org admin. Editable by a SUPER_ADMIN.', 'ADMIN'),
+    ('FIELD_TECH',  'Field tech. Read fleet data, blink the locator LED, download logs, manage racks. Editable by a SUPER_ADMIN.', 'FIELD_TECH')
+) AS b(name, description, builtin_key)
+WHERE org.deleted_at IS NULL
```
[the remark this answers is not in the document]
mcharles-square: > Remove the `org.deleted_at IS NULL` filter here or handle those orgs in the backfill path: `SoftDeleteOrganization` only marks `organization.deleted_at`, so soft-deleted orgs can still have active `user_organization` rows.

Fixed in `54336d8` by removing the filter. Soft-deleted orgs now get their three per-org built-in rows during 000052 (three bounded rows per org), the repoint in step 3 succeeds, and 000053's backfill produces FK-valid rows even when a soft-deleted org has active memberships. `ListActiveOrganizationIDs` still excludes deleted orgs from the boot reconciler so we don't waste cycles reconciling their permissions on every restart — but the rows exist and would be picked up on undelete. New test `TestSeed_SoftDeletedOrgStillGetsBuiltins` locks the behavior in.

server/migrations/000052_create_permission_tables.down.sql
```diff
@@ -0,0 +1,24 @@
...
+
+ALTER TABLE role
+    DROP CONSTRAINT IF EXISTS fk_role_organization,
+    DROP COLUMN IF EXISTS organization_id,
+    DROP COLUMN IF EXISTS builtin_key,
+    DROP COLUMN IF EXISTS is_builtin;
+
+-- Restore the global name uniqueness that 000002 originally shipped.
+ALTER TABLE role
+    ADD CONSTRAINT uq_role_name UNIQUE (name);
```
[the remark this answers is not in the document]
mcharles-square: > Before this migration, `uq_role_org_custom_name` allows two organizations to have active custom roles with the same `name`; during rollback, adding `uq_role_name` on `role(name)` will then fail with a duplicate-key error and abort the down migration.

Fixed in `6105c60`. The 000051 down now hard-deletes every custom row (and its role_permission rows) before restoring `uq_role_name`. PR 1 ships no RPC to create custom roles so production rollback today is a no-op; the cleanup is forward-safe for once UpdateCustomRole / CreateCustomRole RPCs land. Operators who need the data have to snapshot before rollback — added a clear comment in the down file noting that.

mcharles-square:
## Codex Security Review (round 3) — response

All three findings addressed in `6105c60`.

### HIGH — RBAC rollback is blocked by retained assignment FK references

> `000053` down only soft-deletes `user_organization_role` rows, but those rows still retain FK references to per-org built-in `role` rows. The next rollback step, `000052` down, then tries to `DELETE` those built-in roles, which will fail under `fk_user_org_role_role`.

Fixed: 000053 down now **hard-deletes** the mirrored assignment rows (using `DELETE ... USING user_organization` instead of `UPDATE ... SET deleted_at`). Also dropped the redundant `uo.deleted_at IS NULL` filter so the down also clears rows produced by the dual-write paths during the soak window. After 000053 down runs, no live or soft-deleted rows hold the FK, and 000052 down's DELETE on per-org built-in roles succeeds.

### MEDIUM — Custom roles can use built-in role names while authorization still trusts role names

> A custom role named `ADMIN` or `SUPER_ADMIN` could be mistaken for the built-in role by current name-based gates, creating a privilege-escalation path.

Fixed at the DB layer in 000051 up:

```sql
ALTER TABLE role
    ADD CONSTRAINT chk_role_custom_name_not_reserved CHECK (
        is_builtin = TRUE
        OR name NOT IN ('SUPER_ADMIN', 'ADMIN', 'FIELD_TECH')
    );
```

Application code that creates custom roles will surface this as a clear constraint violation. The structural block holds even if a future migration or a direct DB insert tries to create a colliding name.

### MEDIUM — New backfill tests create orgs after reconciliation

> Insert the test organization before `authz.Reconcile`, or call `authz.SeedOrgBuiltins` after direct org inserts.

Fixed: all five affected tests in `backfill_test.go` now `insertTestOrganization(t, db)` and `insertTestUser(t, db)` **before** `authz.Reconcile(ctx, db)`, so the org exists when Reconcile iterates `ListActiveOrganizationIDs` and the per-org built-ins are created. `reconcile_test.go` already used the correct ordering — backfill was the outlier.

server/migrations/000052_create_permission_tables.up.sql
```diff
@@ -0,0 +1,173 @@
...
+-- still compare session.Info.Role by string; a custom row named
+-- 'ADMIN', 'SUPER_ADMIN', or 'FIELD_TECH' could be mistaken for the
+-- built-in by name-based code paths until those gates migrate to
+-- builtin_key or permissions. Block the collision at the DB so no
+-- application mistake or future migration can create one.
+ALTER TABLE role
+    ADD CONSTRAINT chk_role_custom_name_not_reserved CHECK (
+        is_builtin = TRUE
+        OR name NOT IN ('SUPER_ADMIN', 'ADMIN', 'FIELD_TECH')
+    );
```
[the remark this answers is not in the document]
mcharles-square: > `chk_role_custom_name_not_reserved` is added before legacy rows are migrated, but preexisting deployments already contain non-built-in `ADMIN` (seeded in `000002_create_core_tables.up.sql`) and often non-built-in `SUPER_ADMIN` rows from onboarding; because this check does not exempt legacy rows (or `deleted_at` rows), adding it will fail validation immediately and abort migration `000051` on upgrade.

Fixed in `14cde48`. The CHECK now exempts `organization_id IS NULL` legacy rows:

```sql
CHECK (
    is_builtin = TRUE
    OR organization_id IS NULL
    OR LOWER(BTRIM(name)) NOT IN ('super_admin', 'admin', 'field_tech')
)
```

The exemption is safe because `CreateCustomRole` always sets `organization_id` (the column is required for any custom-role insert path), so an application caller cannot exploit it. After 000052 soft-deletes the legacy rows the exemption is irrelevant to ongoing operations.

server/migrations/000051_create_permission_tables.down.sql
```diff
@@ -0,0 +1,38 @@
...
+
+-- Custom roles are a per-org concept introduced in this migration; they
+-- have no representation in the pre-up schema. The new partial unique
+-- index uq_role_org_custom_name allowed multiple orgs to share a custom
+-- name, so restoring the global uq_role_name below would fail with a
+-- duplicate-key error on any environment where customs exist. Hard
+-- delete every custom row before the constraint is restored. (PR 1
+-- ships no RPC to create customs, so this is a no-op in production
+-- today; it makes the down safe once custom-role CRUD lands.)
+DELETE FROM role_permission
```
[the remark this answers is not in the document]
mcharles-square: > Running the `000051` down migration currently fails because `role_permission` is dropped at the top of the file, but later this script still executes `DELETE FROM role_permission ...`; PostgreSQL will error with `relation "role_permission" does not exist`

Fixed in `14cde48`. Rewrote the down migration with explicit ordered steps:

1. Custom-role cleanup (DELETE FROM role_permission + DELETE FROM role) — runs FIRST, while the table and the is_builtin/organization_id columns still exist.
2. DROP new tables (user_organization_role, role_permission, permission).
3. DROP role-table columns/constraints/indexes.
4. Restore `uq_role_name UNIQUE(name)`.

Also restricted the role delete to `organization_id IS NOT NULL` so the legacy ADMIN/SUPER_ADMIN rows that 000052 down just restored are preserved — without that, user_organization.role_id would dangle after rollback.

mcharles-square:
## Codex Security Review (round 4) — response

Both findings addressed in `14cde48`. They cover the same two bugs the inline threads flagged.

### HIGH — RBAC migration rejects existing ADMIN/SUPER_ADMIN rows

> `000051` adds `chk_role_custom_name_not_reserved` while existing legacy rows still have `is_builtin = FALSE` and names like `ADMIN` … PostgreSQL validates the new CHECK against existing rows immediately, so the migration fails before `000052` can repoint and soft-delete those legacy rows.

Fixed: the CHECK now exempts `organization_id IS NULL`. Legacy global rows pass; `CreateCustomRole` (which always sets `organization_id`) cannot exploit the exemption.

### MEDIUM — Down migration cannot roll back cleanly

> The down migration drops `role_permission` at line 6, then later tries to `DELETE FROM role_permission`, which will fail with "relation does not exist." It also deletes every `is_builtin = FALSE` role, which includes the legacy `ADMIN`/`SUPER_ADMIN` rows that `000052.down` just restores

Fixed: rewrote the down migration into four explicit steps — custom cleanup first (while the table and the is_builtin/organization_id columns still exist), then table drops, then role-column drops, then `uq_role_name` restored. The custom-role delete is restricted to `organization_id IS NOT NULL` so the legacy globals 000052 restored stay in place and `user_organization.role_id` does not dangle.

Build + lint clean.

server/internal/domain/stores/interfaces/user.go
```diff
@@ -23,10 +23,11 @@ type UserStore interface { //nolint:interfacebloat // GetUserByIDForUpdate is a
 }
 
 // UserManagementStore provides multi-user account management operations
-type UserManagementStore interface {
+type UserManagementStore interface { //nolint:interfacebloat // user mgmt store covers create + lookup + role lookup; splitting would fragment the call sites that need them together
 	CreateUser(ctx context.Context, externalUserID string, username string, passwordHash string, requiresPasswordChange bool) (int64, error)
 	CreateUserOrganizationRole(ctx context.Context, userID int64, organizationID int64, roleID int64) error
 	GetRoleByName(ctx context.Context, roleName string) (Role, error)
```
flesher: this has no callers.  dead code? should we delete?

flesher: and the query is not scoped to org as well.

server/sqlc/queries/role.sql
flesher: see comment in interface, dead code?

server/cmd/fleetd/main.go
```diff
@@ -179,6 +180,10 @@ func start(config *Config) error {
 		return err
 	}
 
+	if err := authz.Reconcile(context.Background(), conn); err != nil {
```
flesher: this has no deadline. wrap with `context.WithTimeout`?

server/internal/domain/authz/reconcile.go
```diff
@@ -0,0 +1,216 @@
...
+//     orgs' ADMIN/FIELD_TECH; operators add them via the role editor
+//     if they want them.
+//
+// Catalog row policy: permissions are always upserted (description
+// text refreshed). A permission key removed from catalog.go is NOT
+// dropped from the permission table — that is a deliberate manual
+// migration because deleting a catalog row would also drop every
+// role_permission referencing it.
+func Reconcile(ctx context.Context, db *sql.DB) error {
+	tx, err := db.BeginTx(ctx, nil)
```
flesher: this directly starts a tx with no retry logic.  Should we add a retry loop?

flesher (approved):

server/internal/domain/stores/interfaces/user.go
mcharles-square: this is 1 of 4 so will be used in subsequent PRs

server/sqlc/queries/role.sql
mcharles-square: will get used in subsequent PR

server/internal/domain/stores/interfaces/user.go
mcharles-square: removing from this PR

server/sqlc/queries/role.sql
mcharles-square: removing from PR

server/cmd/fleetd/main.go
mcharles-square: Added timeout

server/internal/domain/stores/interfaces/user.go
mcharles-square: > this has no callers.  dead code? should we delete?

Confirmed dead and deleted in `334fbc4` — removed from the `UserManagementStore` interface, the `SQLUserStore` impl, the sqlc query, and regenerated the mock. The only callers were the round-2-replaced `auth.Service.CreateUser` (now uses `GetBuiltinRoleForOrg`) and tests against that caller.

server/sqlc/queries/role.sql
mcharles-square: > see comment in interface, dead code?

Same fix in `334fbc4` — deleted the `GetRoleByName` sqlc query along with the interface method and impl.

server/cmd/fleetd/main.go
mcharles-square: > this has no deadline. wrap with `context.WithTimeout`?

Fixed in `334fbc4`. Wrapped with `context.WithTimeout(60s)`. The reconcile holds an advisory lock so concurrent boots serialize; without a deadline a stuck waiter could block boot indefinitely. 60s is generous for the work (catalog upsert + 3 role rows per active org) and short enough that a stuck instance crashes loud.

server/internal/domain/authz/reconcile.go
mcharles-square: > this directly starts a tx with no retry logic.  Should we add a retry loop?

Fixed in `334fbc4`. Refactored `Reconcile` to use `db.WithTransactionNoResult`, which retries on the project's standard retryable Postgres errors with backoff. Moved the advisory lock into a new sqlc query (`AcquireReconcileLock`) so the closure can acquire it inside the wrapped tx — a retry re-acquires it cleanly because the lock is tx-scoped.

mcharles-square:
## Codex Security Review (round 5) — response

Both findings addressed in `334fbc4`.

### HIGH — Non-super-admin sessions can create a new ADMIN user

> `CreateUser` resolves the current org's built-in `ADMIN` role and then creates a user with that role, but this path still does not call `checkCanManageUser` or enforce the new `user:manage` permission before doing so.

Fixed: added `s.checkCanManageUser(ctx, orgID)` at the top of `CreateUser`, before the temp-password generation and before any role/assignment table access. The check runs the same SUPER_ADMIN-only guard that `ResetUserPassword` and `DeactivateUser` already use. An unauthorized caller now fails the permission check before any DB side effects.

This was a pre-existing gap (the bug predates this PR), but you're right that the dual-write I just added makes it worse — escalation would now create rows in both `user_organization` AND `user_organization_role`. Closing the hole in this PR rather than waiting for PR 2's `RequirePermission` swap.

### MEDIUM — Last-SUPER_ADMIN guards count deactivated users

> The new `CountOrgScopeSuperAdminsExcludingAssignment` and `CountOrgScopeSuperAdminsExcludingUser` queries count active assignment rows, but they do not join `"user"` or filter `u.deleted_at IS NULL`. Since deactivation is a soft delete and this PR starts dual-writing assignment rows, a deactivated SUPER_ADMIN can still be counted as preserving the org's last super-admin.

Fixed: both count queries now `JOIN "user" u ON u.id = uor.user_id` and filter on `u.deleted_at IS NULL` plus `r.deleted_at IS NULL`. A deactivated SUPER_ADMIN no longer preserves the count, so the last actually-usable SUPER_ADMIN can't be removed by future flows.

Tying the soft-delete cascade between `"user"` and `user_organization_role` (so deactivating a user also soft-deletes their assignment rows in the same transaction) is the second half of the recommendation — that belongs in PR 2's caller swap where `DeactivateUser` is rewritten to consult the assignment table. Filed as a TODO note for that PR; the count-query fix here is the standalone defense.

Build + lint clean.

PR #290, opened 2026-05-21
fix(ui): ensure unauthenticated miners can be selected and unpaired

### LINKED ISSUE #286: bug: cannot unpair un-authenticated or partially set up miner
### Describe the bug

When a device is added but unable to authenticate, there is no way to unpair the device, since the selector checkbox is disabled.

### Preconditions

_No response_

### Steps to reproduce

1. Change the password or login info on a machine,
2. Add that machine to fleet,
3. Wait for "Authenticate Miners" popup,
4. Try to unpair the device by selecting the checkbox.

### Expected behavior

I should be able to select the device and rather than having the checkbox disabled, nearly all of the control buttons should be disabled.

### Proto Fleet version

v0.2.7

### Environment

Browser: Firefox 151.0 (64-bit)

### Logs

```shell

```

Image

b-rowan:
This is a bit more verbose than I would like, and there may be better ways to do this, but my experience with typescript is pretty limited.  Tested and working, this may affect cases outside the issue specific case, but I think that is generally expected behavior, EG I should always be able to remove a device no matter what state its in, as removing and re-adding may solve some issues.

Fixes: #286

### DIFF
--- a/client/src/protoFleet/features/fleetManagement/components/MinerActionsMenu/MinerActionsMenu.test.tsx
+++ b/client/src/protoFleet/features/fleetManagement/components/MinerActionsMenu/MinerActionsMenu.test.tsx
@@ -3,6 +3,10 @@ import { act, fireEvent, render, screen, waitFor } from "@testing-library/react"
 import { describe, expect, test, vi } from "vitest";
 import { deviceActions, groupActions, performanceActions, settingsActions } from "./constants";
 import MinerActionsMenu from "./MinerActionsMenu";
+import {
+  type MinerStateSnapshot,
+  PairingStatus,
+} from "@/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb";
 
 // Use vi.hoisted to properly hoist mock variable declarations
 const {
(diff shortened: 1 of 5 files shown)


b-rowan:
This may have caused issues with unpairing devices, but I can't tell if that's because of these changes or because they're offline, I'll do more digging.

b-rowan:
> This may have caused issues with unpairing devices, but I can't tell if that's because of these changes or because they're offline, I'll do more digging.

I think this is unrelated, will open a separate issue:

```
Uncaught (in promise) TypeError: crypto.randomUUID is not a function
    t http://10.0.11.202/assets/BulkActions-CZbOy-3-.js:1
    w http://10.0.11.202/assets/BulkActions-CZbOy-3-.js:1
    yd http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:8
    wd http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:8
    _n http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:8
    wd http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:8
    up http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:9
    cp http://10.0.11.202/assets/vendor-react-BYKXDB4x.js:9
BulkActions-CZbOy-3-.js:1:28560

```

rongxin-liu:
Hey @b-rowan! Thank you for opening this PR. Your analysis was right but we landed the same intent in a slightly different shape in #310. The main differences are that we routed the gating through the existing per-action capability pipeline instead of filtering the action list.

PR #294, opened 2026-05-21
fix(fleetmanagement): omit discovery port from unpaired snapshot.Url

### LINKED ISSUE #287: bug: clicking a miner's IP in the table opens a bad URL
### Describe the bug

When clicking a miner's IP in the Miners table, it opens a page with the wrong port number.

### Preconditions

_No response_

### Steps to reproduce

This is reproducible with Whatsminer devices - 
1. Add a Whatsminer,
2. Go to the "Miners" page,
3. Click the IP address in the table,
4. Get redirected to `{ip}:4028`, which is not a valid UI location (contains the CGMiner API).

### Expected behavior

The IP link should always open port 80 and expect the device to redirect as needed from there.

### Proto Fleet version

v0.2.7

### Environment

Browser: Firefox 151.0 (64-bit)

### Logs

```shell

```

_No response_

ankitgoswami:
## Summary

- Server-side fix for #287: clicking an unpaired/discovered miner's IP in the Miners table opened the device's JSON API port (`:4028` for Whatsminer / CGMiner mAPI) instead of its web UI.
- The unpaired branch in [`buildSnapshotsFromUnifiedQuery`](server/internal/domain/fleetmanagement/service.go) was hand-rolling a URL that re-appended `row.Port` whenever it wasn't 80 or 443. `row.Port` is the discovery / control-plane port stored at pairing time, not the web UI port. Routing through the existing `constructWebViewURL` helper gives unpaired devices the same port-omitted, scheme-preserved URL as paired devices — which is what `snapshot.Url` is documented to mean ([service.go:77-81](server/internal/domain/fleetmanagement/service.go#L77-L81)).
- Hoists the shared `IpAddress` / `Url` assignments above the `isPaired` switch so both branches use the same helper, and drops the two now-unused `defaultHTTPPort` / `defaultHTTPSPort` constants.

## Why this instead of #288

This change fixes `snapshot.Url` at the source so the existing client code (`<a href={miner.url}>`) just works. #288 can be closed.

## Test plan

- [x] New `TestConstructWebViewURL` unit test pins the URL semantics: HTTP IPv4, HTTPS scheme preserved, bare IPv6 bracketed, already-bracketed IPv6 not double-bracketed, empty scheme/host return empty string.
- [x] Existing `TestService_ListMinerStateSnapshots_ShouldPopulateCapabilitiesForUnpairedDevices` now asserts `miner.Url == \"http://192.168.1.100\"` against a fixture with `Port=\"4028\"`, pinning the end-to-end behavior reported in #287.
- [x] `go vet` and `golangci-lint` (changed scope) clean.
- [x] CI runs the full server integration suite against the proto-fleet test DB.
- [x] Manual verification on the Whatsminer M30S+ VH60 device that prompted #287 (clicking the IP cell opens `http://<ip>` on port 80, browser reaches the device web UI).

Fixes #287

### DIFF
--- a/server/internal/domain/fleetmanagement/service.go
+++ b/server/internal/domain/fleetmanagement/service.go
@@ -45,10 +45,6 @@ const (
 	// maxPageSize is the maximum number of items that can be returned per page
 	maxPageSize = 1000
 
-	// Standard HTTP ports
-	defaultHTTPPort  = "80"
-	defaultHTTPSPort = "443"
-
 	// concurrentClearAuthKeyLimit bounds the number of parallel ClearAuthKey RPCs
 	// fired in the background after a delete operation
 	concurrentClearAuthKeyLimit = 20
@@ -468,25 +464,18 @@ func (s *Service) buildSnapshotsFromUnifiedQuery(
 
 		snapshot.Name = ComposeDeviceName(row.CustomName.String, snapshot.Manufacturer, snapshot.Model)
 
+		snapshot.IpAddress = row.IpAddress
+		snapshot.Url = constructWebViewURL(row.UrlScheme, row.IpAddress)
+
 		if isPaired {
 			snapshot.MacAddress = row.MacAddress
 			if row.SerialNumber.Valid {
 				snapshot.SerialNumber = row.SerialNumber.String
 			}
-			snapshot.IpAddress = row.IpAddress
-			snapshot.Url = constructWebViewURL(row.UrlScheme, row.IpAddress)
-
 			if row.DeviceStatus.Valid {
 				snapshot.DeviceStatus = convertDeviceStatusStringToProto(string(row.DeviceStatus.DeviceStatusEnum))
 			}
 		} else {
-			snapshot.IpAddress = row.IpAddress
-
-			url := row.UrlScheme + "://" + bracketIPv6Host(row.IpAddress)
-			if row.Port != "" && row.Port != defaultHTTPPort && row.Port != defaultHTTPSPort {
-				url += ":" + row.Port
-			}
-			snapshot.Url = url
 			snapshot.DeviceStatus = pb.DeviceStatus_DEVICE_STATUS_INACTIVE
 		}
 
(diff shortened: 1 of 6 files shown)


server/internal/domain/fleetmanagement/service.go
```diff
@@ -468,25 +464,18 @@ func (s *Service) buildSnapshotsFromUnifiedQuery(
 
 		snapshot.Name = ComposeDeviceName(row.CustomName.String, snapshot.Manufacturer, snapshot.Model)
 
+		snapshot.IpAddress = row.IpAddress
+		snapshot.Url = constructWebViewURL(row.UrlScheme, row.IpAddress)
```
b-rowan, line 468: Curious if `UrlScheme` is even required?  In theory all devices should just be able to use `http://{IP}`, which makes it much easier to maintain since when new miners are added nothing needs to change.  Any miners on https will have a redirect on the http port, since when a user normally accesses them to configure via the interface, they will just enter the IP in the browser.

Also, for future proofing, if there is a plan to eventually swap the backend to be entirely `asic-rs` based, this `UrlScheme` will no longer be tracked.

b-rowan (approved):

server/internal/domain/fleetmanagement/service.go
rongxin-liu, line 468: Both branches now share one helper but the proto godoc on the field this populates (`proto/fleetmanagement/v1/fleetmanagement.proto:108`) still claims port is included for non-default ports. Worth updating it here + `just gen` would keep the contract docs and server behavior aligned across both pairing states.

rongxin-liu (approved):

server/internal/domain/fleetmanagement/service.go
ankitgoswami: Good question. Proto devices in particular are HTTPS-native so stripping the scheme would unnecessarily add the redirect. That said, willing to revisit once we start integrating asic-rs more deeply.

PR #309, opened 2026-05-26
Add Fleet saved view lifecycle E2E coverage

edgars-avotins:
## Summary
- add Fleet miners saved view rename coverage
- add Fleet miners saved view delete coverage
- extend the miners page object with rename/delete view helpers

## Test plan
- ./node_modules/.bin/eslint e2eTests/protoFleet/pages/miners.ts e2eTests/protoFleet/spec/minersFiltersViews.spec.ts
- ./node_modules/.bin/playwright test -c e2eTests/protoFleet/playwright.config.ts e2eTests/protoFleet/spec/minersFiltersViews.spec.ts --project=desktop --no-deps
- ./node_modules/.bin/playwright test -c e2eTests/protoFleet/playwright.config.ts e2eTests/protoFleet/spec/minersFiltersViews.spec.ts --project=mobile --no-deps

### DIFF
--- a/client/e2eTests/protoFleet/spec/minersFiltersViews.spec.ts
+++ b/client/e2eTests/protoFleet/spec/minersFiltersViews.spec.ts
@@ -254,11 +254,12 @@ test.describe("Proto Fleet - Miners filters and saved views", () => {
     });
   });
 
-  test("Saved view can be renamed and persists after reload", async ({ minersPage, commonSteps }) => {
+  test("Saved view can be renamed and persists after reload", async ({ minersPage, commonSteps, page }) => {
     const originalViewName = generateRandomText("miners_view");
     const renamedViewName = generateRandomText("renamed_view");
     let firstMinerIp = "";
     let firstMinerSubnet = "";
+    let activeViewId = "";
 
     await commonSteps.loginAsAdmin();
     await commonSteps.goToMinersPage();
@@ -289,6 +290,11 @@ test.describe("Proto Fleet - Miners filters and saved views", () => {
       await minersPage.validateViewTabNotVisible(originalViewName);
       await minersPage.validateActiveFilterSummary("subnet", firstMinerSubnet);
       await minersPage.validateMinerInList(firstMinerIp);
+
+      const searchParams = new URL(page.url()).searchParams;
+      activeViewId = searchParams.get("view") ?? "";
+      test.expect(activeViewId).not.toBe("");
+      test.expect(searchParams.getAll("subnet")).toEqual([firstMinerSubnet]);
     });
 
     await test.step("Reload and validate the renamed view persists", async () => {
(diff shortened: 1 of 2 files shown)


client/e2eTests/protoFleet/spec/minersFiltersViews.spec.ts
```diff
@@ -253,4 +253,104 @@ test.describe("Proto Fleet - Miners filters and saved views", () => {
...
+      await minersPage.updateSavedView();
+    });
+
+    await test.step("Validate the renamed view is active and the old name is gone", async () => {
+      await minersPage.validateViewTabVisible(renamedViewName);
+      await minersPage.validateViewTabActive(renamedViewName);
+      await minersPage.validateViewTabNotVisible(originalViewName);
+      await minersPage.validateActiveFilterSummary("subnet", firstMinerSubnet);
+      await minersPage.validateMinerInList(firstMinerIp);
+    });
```
rongxin-liu, line 298: Worth checking the URL `?view=<uuid>` param here so if a future regression corrupts the view URL on rename (id swap, param strip, etc.,) we can catch it here. The earlier update/reset tests in the file (lines 246-253) set the bar with explicit `searchParams` checks.

Perhaps we can add `page` to the fixture and assert here (and again in the reload step at lines 294-304):

```ts
const searchParams = new URL(page.url()).searchParams;
test.expect(searchParams.get("view")).not.toBeNull();
test.expect(searchParams.getAll("subnet")).toEqual([firstMinerSubnet]);
```

client/e2eTests/protoFleet/spec/minersFiltersViews.spec.ts
```diff
@@ -253,4 +253,104 @@ test.describe("Proto Fleet - Miners filters and saved views", () => {
...
+      await minersPage.validateViewModalOpened("New view");
+      await minersPage.inputViewName(viewName);
+      await minersPage.saveNewView();
+    });
+
+    await test.step("Return to All miners so the saved view becomes inactive", async () => {
+      await minersPage.clickViewTab("All miners");
+      await minersPage.waitForMinersListToLoad();
+      await minersPage.validateViewTabActive("All miners");
+    });
```
rongxin-liu, line 339: Switching to `All miners` before deleting means this test only exercises the inactive-view delete path. The active-view delete branch `if (activeViewId === idToDelete)` in [client/src/protoFleet/components/ViewsBar.tsx:460-466](https://github.com/block/proto-fleet/blob/main/client/src/protoFleet/components/ViewsBar.tsx#L460-L466), which strips `?view=` from the URL when the deleted view was the active one, isn't exercised by any test in the suite.

Worth adding a second test next to this one (or extend this test with a "delete while active" path) that deletes the view while it remains active, then asserts the same `searchParams.get('view')` is null and `.getAll('subnet')` is empty checks as the existing assertions on lines 341-343. Same shape, just without the intermediate `All miners` switch.

rongxin-liu (approved):

PR #325, opened 2026-05-27
feat(authz): gate ungated handlers with RequirePermission (PR 2d.1)

mcharles-square:
First slice of PR 2d. New-gate work for every currently-ungated authenticated procedure whose target catalog key already exists. Per-miner site-scope narrowing is deferred until DeviceSelector resolution lands.

## What landed

71 procedures move from `ProceduresPendingMigration` → `ProcedurePermissions`:

| Package | # | Mapping |
|---|---|---|
| `fleetmanagement` | 9 | `fleet:read` for aggregates, `miner:read` for per-row reads, matching `miner:*` for mutations (rename, delete, update_worker_names, export_csv) |
| `minercommand` | 15 | one-to-one with catalog `miner:*` keys; stream/batch endpoints on `fleet:read` |
| `deviceset` | 17 | `rack:read` / `rack:manage` |
| `collection` (DeviceCollectionService) | 16 | `rack:read` / `rack:manage` — same mapping as deviceset; the two are the legacy + new wire surfaces for the same domain |
| `errorquery` | 4 | `fleet:read` (org-scoped diagnostics) |
| `telemetry` | 2 | `fleet:read` |
| `networkinfo` | 2 | `site:read` (Get), `site:manage` (Update) |
| `pairing` | 2 | `miner:pair` |
| `foremanimport` | 2 | `miner:pair` (bulk pair flow) |
| `onboarding` | 1 | `fleet:read` |
| `curtailment` reads | 1 | `curtailment:read` (`GetActiveCurtailment`; `ListCurtailmentEvents` was already gated) |

## What this PR does NOT do

- **No per-miner site narrowing.** Every gate runs at org scope (`ResourceContext{}`). Threading `SiteID` per miner requires resolving `DeviceSelector` payloads first — its own slice.
- **Activity (3), Pools (5), Schedule (7) — deferred.** They need new catalog keys (`activity:read`, `pool:read`/`manage`, `schedule:read`/`manage`) plus an ADMIN backfill migration; staying in `ProceduresPendingMigration` with explicit notes.
- **Curtailment Start/Stop/Preview — deferred.** Conditional inline gates that need a redesign (PR 2d.2 scope).
- **fleetnodeadmin Pair/Unpair/ListFleetNodeDevices/DiscoverOnFleetNode — deferred.** Unimplemented stubs.
- **No schema migration in this PR.** The legacy `user_organization.role_id` rename + raising trigger + `middleware/admin.go` deletion are PR 2d.3.

## Test infrastructure

- `testutil.MockAuthContextForTesting` now wires an org-scope `EffectivePermissions` carrying every catalog key. Unit tests that exercise gated handler logic clear the gate without each call site opting in; tests that need to assert `PermissionDenied` build their own `EffectivePermissions` explicitly.
- `errorquery` handler test: swapped direct `authn.SetInfo` for the shared testutil helper. Expected-code assertion now handles both `FleetError` (from `RequirePermission`) and `connect.Error` wrappers.
- `telemetry` tests: added a permissive `GetMinerStateCounts` mock since the live-uptime-bar pass now fires when `orgID != 0` (skipped on the prior no-session test paths).
- `curtailment GetActiveCurtailment` tests use a local `withCurtailmentRead` wrapper to add `PermCurtailmentRead` to the context.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/handlers/middleware/...` green — contract test confirms every registered procedure is classified into one of the four buckets after the swap.
- [x] All non-DB handler unit tests green: activity, buildings, command, curtailment, deviceset, errorquery, firmware, fleetnodegateway, middleware, pairing, sites, telemetry.
- [ ] CI confirms no regression on Postgres-backed integration tests (apikey, auth, interceptors, fleetmanagement ListMinerStateSnapshots, onboarding CreateAdminLogin / GetFleetInitStatus — all pre-existing DB dependencies).

## Rollback

Revert is mechanical: each handler removes its `RequirePermission` call and the procedure moves back to `ProceduresPendingMigration`. The `testutil` helper change is additive (test-only); no production behavior depends on the new permission set in the test context.

### DIFF
--- a/server/internal/handlers/collection/handler.go
+++ b/server/internal/handlers/collection/handler.go
@@ -25,16 +25,9 @@ func NewHandler(svc *collection.Service) *Handler {
 	}
 }
 
-// requirePerm gates a handler at org scope. Per-rack site narrowing
-// is deferred until the request payload surfaces the target site.
-func requirePerm(ctx context.Context, key string) error {
-	_, err := middleware.RequirePermission(ctx, key, authz.ResourceContext{})
-	return err
-}
-
 // CreateCollection creates a new collection.
 func (h *Handler) CreateCollection(ctx context.Context, r *connect.Request[pb.CreateCollectionRequest]) (*connect.Response[pb.CreateCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.CreateCollection(ctx, r.Msg)
@@ -46,7 +39,7 @@ func (h *Handler) CreateCollection(ctx context.Context, r *connect.Request[pb.Cr
 
 // GetCollection retrieves a collection by ID.
 func (h *Handler) GetCollection(ctx context.Context, r *connect.Request[pb.GetCollectionRequest]) (*connect.Response[pb.GetCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackRead); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackRead, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.GetCollection(ctx, r.Msg)
@@ -58,7 +51,7 @@ func (h *Handler) GetCollection(ctx context.Context, r *connect.Request[pb.GetCo
 
 // UpdateCollection updates a collection's label and/or description.
 func (h *Handler) UpdateCollection(ctx context.Context, r *connect.Request[pb.UpdateCollectionRequest]) (*connect.Response[pb.UpdateCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.UpdateCollection(ctx, r.Msg)
@@ -70,7 +63,7 @@ func (h *Handler) UpdateCollection(ctx context.Context, r *connect.Request[pb.Up
 
 // DeleteCollection soft-deletes a collection.
 func (h *Handler) DeleteCollection(ctx context.Context, r *connect.Request[pb.DeleteCollectionRequest]) (*connect.Response[pb.DeleteCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.DeleteCollection(ctx, r.Msg)
@@ -82,7 +75,7 @@ func (h *Handler) DeleteCollection(ctx context.Context, r *connect.Request[pb.De
 
 // ListCollections returns all collections for the organization.
 func (h *Handler) ListCollections(ctx context.Context, r *connect.Request[pb.ListCollectionsRequest]) (*connect.Response[pb.ListCollectionsResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackRead); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackRead, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.ListCollections(ctx, r.Msg)
@@ -94,7 +87,7 @@ func (h *Handler) ListCollections(ctx context.Context, r *connect.Request[pb.Lis
 
 // AddDevicesToCollection adds devices to a collection.
 func (h *Handler) AddDevicesToCollection(ctx context.Context, r *connect.Request[pb.AddDevicesToCollectionRequest]) (*connect.Response[pb.AddDevicesToCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.AddDevicesToCollection(ctx, r.Msg)
@@ -106,7 +99,7 @@ func (h *Handler) AddDevicesToCollection(ctx context.Context, r *connect.Request
 
 // RemoveDevicesFromCollection removes devices from a collection.
 func (h *Handler) RemoveDevicesFromCollection(ctx context.Context, r *connect.Request[pb.RemoveDevicesFromCollectionRequest]) (*connect.Response[pb.RemoveDevicesFromCollectionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.RemoveDevicesFromCollection(ctx, r.Msg)
@@ -118,7 +111,7 @@ func (h *Handler) RemoveDevicesFromCollection(ctx context.Context, r *connect.Re
 
 // ListCollectionMembers returns all members of a collection.
 func (h *Handler) ListCollectionMembers(ctx context.Context, r *connect.Request[pb.ListCollectionMembersRequest]) (*connect.Response[pb.ListCollectionMembersResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackRead); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackRead, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.ListCollectionMembers(ctx, r.Msg)
@@ -130,7 +123,7 @@ func (h *Handler) ListCollectionMembers(ctx context.Context, r *connect.Request[
 
 // GetDeviceCollections returns all collections a device belongs to.
 func (h *Handler) GetDeviceCollections(ctx context.Context, r *connect.Request[pb.GetDeviceCollectionsRequest]) (*connect.Response[pb.GetDeviceCollectionsResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackRead); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackRead, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.GetDeviceCollections(ctx, r.Msg)
@@ -142,7 +135,7 @@ func (h *Handler) GetDeviceCollections(ctx context.Context, r *connect.Request[p
 
 // SetRackSlotPosition sets a device's slot position within a rack.
 func (h *Handler) SetRackSlotPosition(ctx context.Context, r *connect.Request[pb.SetRackSlotPositionRequest]) (*connect.Response[pb.SetRackSlotPositionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.SetRackSlotPosition(ctx, r.Msg)
@@ -154,7 +147,7 @@ func (h *Handler) SetRackSlotPosition(ctx context.Context, r *connect.Request[pb
 
 // ClearRackSlotPosition clears a device's slot position within a rack.
 func (h *Handler) ClearRackSlotPosition(ctx context.Context, r *connect.Request[pb.ClearRackSlotPositionRequest]) (*connect.Response[pb.ClearRackSlotPositionResponse], error) {
-	if err := requirePerm(ctx, authz.PermRackManage); err != nil {
+	if _, err := middleware.RequirePermission(ctx, authz.PermRackManage, authz.ResourceContext{}); err != nil {
 		return nil, err
 	}
 	result, err := h.collectionSvc.ClearRackSlotPosition(ctx, r.Msg)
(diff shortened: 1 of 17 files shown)


server/internal/handlers/networkinfo/handler.go
```diff
@@ -37,7 +42,10 @@ func (h Handler) GetNetworkInfo(ctx context.Context, _ *connect.Request[pb.GetNe
 	}), nil
 }
 
-func (h Handler) UpdateNetworkNickname(context.Context, *connect.Request[pb.UpdateNetworkNicknameRequest]) (*connect.Response[pb.UpdateNetworkNicknameResponse], error) {
+func (h Handler) UpdateNetworkNickname(ctx context.Context, _ *connect.Request[pb.UpdateNetworkNicknameRequest]) (*connect.Response[pb.UpdateNetworkNicknameResponse], error) {
+	if _, err := middleware.RequirePermission(ctx, authz.PermSiteManage, authz.ResourceContext{}); err != nil {
+		return nil, err
+	}
 	// TODO implement me
 	panic("unimplemented")
```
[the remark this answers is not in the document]
mcharles-square: > UpdateNetworkNickname still panics with "unimplemented" …

Addressed in commit e924f5da: replaced the `panic("unimplemented")` with `fleeterror.NewUnimplementedError(...)`, which Connect surfaces as `CodeUnimplemented` so clients get a clean response instead of a 500.

server/internal/handlers/collection/handler.go
```diff
@@ -23,8 +25,18 @@ func NewHandler(svc *collection.Service) *Handler {
 	}
 }
 
+// requirePerm gates a handler at org scope. Per-rack site narrowing
+// is deferred until the request payload surfaces the target site.
+func requirePerm(ctx context.Context, key string) error {
+	_, err := middleware.RequirePermission(ctx, key, authz.ResourceContext{})
```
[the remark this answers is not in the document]
mcharles-square: > When callers only have a site-scoped rack role, this org-scope `ResourceContext{}` makes every migrated rack handler reject them even for racks at their assigned site …

Not addressing in this PR: this is the per-resource site-scope narrowing that the PR description explicitly defers. Threading the target `SiteID` requires resolving each request payload (collection/rack ID, DeviceSelector, etc.) to its site before the gate runs, which is its own slice of work — same shape as the broader DeviceSelector resolution needed for `command/*`. Filed as a follow-up to PR 2d.1.

For now, every gated handler runs at org scope. That preserves the previous "anyone authenticated could call this" → "anyone with the right permission in their org" reduction without regressing site-narrowing behavior, because no production caller currently holds site-scope rack roles (FIELD_TECH carries `rack:read`/`rack:manage` but only after a site-scope assignment is created, which happens in a later UI slice).

mcharles-square:
Re: the [Codex security review](https://github.com/block/proto-fleet/pull/325#issuecomment-3315054343):

> [CRITICAL] Miner command RBAC is not bound to the selected devices

This is the per-miner site-scope narrowing piece that PR 2d.1's description explicitly defers. Quoting the PR body:

> **No per-miner site narrowing.** Every gate runs at org scope (`ResourceContext{}`). Threading `SiteID` per miner requires resolving `DeviceSelector` payloads first — its own slice.

Fixing this properly requires:
1. Resolving `DeviceSelector` (`DeviceList` / `AllDevices` / filters) to a concrete miner set before the gate runs.
2. Verifying every selected miner belongs to the caller's org (cross-org IDOR guard).
3. Deriving target `SiteID`s and gating per-site with `ResourceContext{SiteID: ...}`.

That's substantial enough to warrant its own focused PR. Tracking as a follow-up. The current PR's net effect is still a security improvement: previously any authenticated org member could call these endpoints (no gate at all), now they need the matching catalog permission. Org-scope grants narrowed at a site still pass — that's the gap Codex correctly identifies — but it's an unchanged property carried forward from the previous ungated state, not a new regression.

> [HIGH] Command batch status and log bundle reads are not scoped to the caller's org

This one is a real pre-existing IDOR not introduced by this PR. Confirmed by reading the service:

- `service.go:1527 GetCommandBatchLogBundle(batchUUID string)` takes no `ctx`, no session, no org filter — passes the raw UUID straight to the files service.
- `service.go:1489 StreamCommandBatchUpdates(ctx, msg)` has `ctx` but never passes the caller's org into `statusService.StreamCommandBatchUpdates(ctx, msg.BatchIdentifier)`.
- `service.go:1565 GetCommandBatchDeviceResults(ctx, req)` *does* org-scope via `command_batch_log.organization_id` (comment at line 1559).

So 2 of 3 are leaky. My PR added `fleet:read` / `miner:download_logs` gates, which narrowed the audience but doesn't close the IDOR — a holder of those perms in org A can still read a batch from org B if they learn the UUID.

The proper fix changes service signatures (add `ctx` to `GetCommandBatchLogBundle`; thread org through the batch lookups) and likely needs a new store query for an org-scoped batch header check. That's separate security-focused work, not a 2d.1 mechanical-swap concern. Tracking as a follow-up PR.

Neither of these blocks 2d.1's correctness — both are pre-existing gaps that the gate-swap narrows but doesn't fully close. Calling it out here so the follow-up doesn't get lost.

server/internal/handlers/command/handler.go
```diff
@@ -27,10 +29,21 @@ func NewHandler(commandSvc *command.Service) *Handler {
 	}
 }
 
+// requirePerm gates a handler on the supplied catalog key at org scope.
+// Per-miner site narrowing is deferred until DeviceSelector resolution
+// can surface the target miners' SiteIDs before the gate runs.
+func requirePerm(ctx context.Context, key string) error {
+	_, err := middleware.RequirePermission(ctx, key, authz.ResourceContext{})
```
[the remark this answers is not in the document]
mcharles-square: Will be addressed in followup

server/internal/handlers/collection/handler.go
```diff
@@ -23,8 +25,18 @@ func NewHandler(svc *collection.Service) *Handler {
 	}
 }
 
+// requirePerm gates a handler at org scope. Per-rack site narrowing
+// is deferred until the request payload surfaces the target site.
+func requirePerm(ctx context.Context, key string) error {
```
rongxin-liu: This helper seems to be duplicated in `command/handler.go`, `deviceset/handler.go`, and `fleetmanagement/handler.go`, while 7 other packages (`foremanimport`, `networkinfo`, `onboarding`, `pairing`, `telemetry`, `curtailment`, `errorquery`) call `middleware.RequirePermission(...)` inline.

Worth deciding which one we should go with consistently.

server/internal/handlers/errorquery/handler.go
```diff
@@ -31,21 +33,20 @@ func NewHandler(diagnosticsService *diagnostics.Service) *Handler {
...
-	info, err := session.GetInfo(ctx)
-	if err != nil {
-		return nil, connect.NewError(connect.CodeUnauthenticated, err)
-	}
-	return info, nil
+// requireFleetRead is the read gate for every ErrorQueryService method.
+// Diagnostics are scoped to the caller's organization; the gate runs
+// before any service call to fail closed on missing session / missing
+// effective permissions.
+func requireFleetRead(ctx context.Context) (*session.Info, error) {
```
rongxin-liu: Worth considering/aligning here either parameterize like the other helper packages (`requirePerm(ctx, key)`) or inline `middleware.RequirePermission` at the four call sites in this file. Same shape decision as the `requirePerm` wrappers in `collection`/`command`/`deviceset`/`fleetmanagement`.

server/internal/handlers/curtailment/handler.go
```diff
@@ -168,9 +168,9 @@ func (h *Handler) GetActiveCurtailment(ctx context.Context, _ *connect.Request[p
 	if h.service == nil {
 		return nil, errCurtailmentNotImplemented("GetActiveCurtailment")
 	}
-	info, err := session.GetInfo(ctx)
+	info, err := middleware.RequirePermission(ctx, authz.PermCurtailmentRead, authz.ResourceContext{})
```
rongxin-liu, line 171: Some ordering inconsistency across this file: `GetActiveCurtailment` (this line), `ListCurtailmentEvents` (190), and `AdminTerminateEvent` (209) run `RequirePermission` *before* the `h.service == nil` check, while `UpdateCurtailmentEvent` (108) checks `h.service == nil` *first*.

Perhaps nil-check-first to fail fast on misconfig? Or always gate-first to keep `PermissionDenied` as the dominant signal?

server/internal/handlers/telemetry/handler.go
```diff
@@ -29,6 +31,9 @@ func (h *Handler) GetCombinedMetrics(
 	ctx context.Context,
 	req *connect.Request[telemetryv1.GetCombinedMetricsRequest],
 ) (*connect.Response[telemetryv1.GetCombinedMetricsResponse], error) {
+	if _, err := middleware.RequirePermission(ctx, authz.PermFleetRead, authz.ResourceContext{}); err != nil {
```
rongxin-liu: `RequirePermission` already returns session info and guarantees its presence on success, but here the result is discarded, and lines 42-45 re-fetch via `session.GetInfo` inside a defensive `if err == nil` branch that can no longer fail. `StreamCombinedMetricUpdates` five lines below already does this cleanly it seems.

Perhaps we could do:

```suggestion
	info, err := middleware.RequirePermission(ctx, authz.PermFleetRead, authz.ResourceContext{})
	if err != nil {
		return nil, err
	}
```

Then replace lines 41-45 with `query.OrganizationID = info.OrganizationID` (and subsequently drop the unused `session` import).

rongxin-liu (approved):

server/internal/handlers/collection/handler.go
mcharles-square: > This helper seems to be duplicated in `command/handler.go`, `deviceset/handler.go`, and `fleetmanagement/handler.go`, while 7 other packages (`foremanimport`, `networkinfo`, `onboarding`, `pairing`, `telemetry`, `curtailment`, `errorquery`) call `middleware.RequirePermission(...)` inline.

Addressed in f352dc65: inlined `middleware.RequirePermission(ctx, ..., authz.ResourceContext{})` at every call site and dropped the `requirePerm` wrappers from `collection`, `command`, `deviceset`, and `fleetmanagement` (and `requireFleetRead` from `errorquery`). All eleven packages now follow the same shape.

server/internal/handlers/errorquery/handler.go
mcharles-square: > Worth considering/aligning here either parameterize like the other helper packages (`requirePerm(ctx, key)`) or inline `middleware.RequirePermission` at the four call sites in this file.

Inlined in f352dc65 — same shape decision applied repo-wide. `requireFleetRead` is gone; each call site now does `middleware.RequirePermission(ctx, authz.PermFleetRead, authz.ResourceContext{})` directly, and the `session` import is dropped.

server/internal/handlers/curtailment/handler.go
mcharles-square: > `GetActiveCurtailment` (this line), `ListCurtailmentEvents` (190), and `AdminTerminateEvent` (209) run `RequirePermission` *before* the `h.service == nil` check, while `UpdateCurtailmentEvent` (108) checks `h.service == nil` *first*.

Looked at this and the ordering is the inverse of what's described — `GetActiveCurtailment` (167), `ListCurtailmentEvents` (186), and `UpdateCurtailmentEvent` (107) all run `service == nil` first; `AdminTerminateEvent` is the lone gate-first method, called out at line 207 as "the authoritative RBAC check" for the SessionOnly-paired procedure. The non-admin trio is consistent today, and `TestHandler_NonAdminRPCsReturnUnimplemented` pins them to nil-check-first by asserting `Unimplemented` for sessionless calls. Keeping the single intentional exception for `AdminTerminateEvent`.

server/internal/handlers/telemetry/handler.go
mcharles-square: > `RequirePermission` already returns session info and guarantees its presence on success, but here the result is discarded, and lines 42-45 re-fetch via `session.GetInfo` inside a defensive `if err == nil` branch that can no longer fail.

Applied in f352dc65 — `info` now comes straight from `RequirePermission`, `query.OrganizationID = info.OrganizationID` replaces the re-fetch, and the `session` import is dropped. Matches `StreamCombinedMetricUpdates` below.

mcharles-square:
Acknowledging the Codex security review. These overlap substantially with previously-resolved P1 threads on this PR and intentional deferrals tracked in `middleware/rpc_permissions.go`:

- **Site-scoped miner/rack permissions checked as org-scoped** — same concern as the earlier P1 threads on `command/handler.go:36` and `collection/handler.go:31`. Per-miner/per-rack site narrowing requires resolving target site IDs before the gate runs (selector resolution for command, payload for rack CRUD). Intentionally deferred and called out in the PR description; a follow-up will pass `ResourceContext{SiteID: ...}` per target.
- **Pool configuration RPCs ungated** — `PoolsService` mutations remain in `ProceduresPendingMigration` deliberately. Adding `pool:read` / `pool:manage` keys, backfilling roles, and gating these is the next slice.
- **Schedule create/update ungated** — same status: deferred in `ProceduresPendingMigration`. Schedule mutations need to validate the scheduled action against the underlying miner permission/site scope before storing; planned alongside the site-scoping work above so the validation can use the same selector→site resolver.
- **Command log bundle downloads not tenant-scoped** — separate concern that lives in the `command` service layer (`GetCommandBatchLogBundle` reads by batch UUID without an org check). Out of scope for this gate-only PR; tracking as a follow-up.

PR 2d.1's scope is gating the previously-ungated RPCs that have a catalog key today; the site-scoping and pool/schedule gates are queued as named follow-ups so each can land with its own tests.

server/internal/handlers/foremanimport/handler.go
```diff
@@ -41,6 +46,9 @@ func (h *Handler) ImportFromForeman(ctx context.Context, r *connect.Request[pb.I
 }
 
 func (h *Handler) CompleteImport(ctx context.Context, r *connect.Request[pb.CompleteImportRequest]) (*connect.Response[pb.CompleteImportResponse], error) {
+	if _, err := middleware.RequirePermission(ctx, authz.PermMinerPair, authz.ResourceContext{}); err != nil {
```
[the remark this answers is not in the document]
mcharles-square: > a pairing-only role can still alter rack topology via the Foreman import path; require `rack:manage` for the topology-import branches instead of gating the whole RPC solely on `miner:pair`.

Addressed in 6823287b. `CompleteImport` keeps the `miner:pair` gate as the baseline and adds a second `RequirePermission(PermRackManage, ...)` check when `ImportGroups` or `ImportRacks` is set, so the topology-mutating branches now require the same authority as the direct rack handlers.

mcharles-square:
Acknowledging the updated Codex security review.

- **[HIGH] Site-scoped miner/rack permissions checked as org-scoped** — same concern as the prior P1 threads and the previous Codex round. Already covered in the [earlier acknowledgement](https://github.com/block/proto-fleet/pull/325#issuecomment-4566281388): per-target site resolution is the planned follow-up; this PR ships the org-scope baseline.
- **[MEDIUM] Command batch endpoints not bound to caller's organization** — partially overlaps the prior log-bundle finding and adds `StreamCommandBatchUpdates` (line 245). Both will be addressed by threading `organization_id` through the command service in a separate PR — the gate added here is the procedure-level check; the data-layer org binding is the next slice, alongside the `GetCommandBatchLogBundle` tightening already flagged in the previous reply.

No code changes in this round on those findings; tracking them as named follow-ups. The unrelated P2 (Foreman topology import) was fixed in 6823287b.

PR #332, opened 2026-05-28
feat(fleetnode): device pairing + agent reporting (server) [PR 1/2]

ankitgoswami:
PR 1 of a stack. Adds the server-side surfaces operators need to manage fleet-node device pairings and the gateway endpoint agents call to report devices they discovered on their LAN.

## Summary
- **fleetnodepairing domain** — `Service` + `Store` with `PairDevice`, `UnpairDevice`, `ListPairs`, `ListDevicesForFleetNode`, `UpsertDiscoveredDevices`. Agent reports are validated for RFC1918/RFC4193 IP, port range, and http/https scheme.
- **`UpsertDiscoveredDeviceFromFleetNode`** — `NOT EXISTS` pairing guard prevents fleet node B from overwriting a device already paired with fleet node A.
- **FleetNodeGateway.ReportDiscoveredDevices** — agent-authenticated RPC that ingests discovered devices and routes them through the pairing service.
- **FleetNodeAdmin.PairDeviceToFleetNode / UnpairDevice / ListFleetNodeDevices** — operator endpoints, gated by `fleetnode:manage` / `fleetnode:read` via `middleware.RequirePermission`.
- **Revoke now cascades** — pulled `RevocationCleanupStore` out of `fleetnodeenrollment.Store` so `RevokeFleetNode` deletes `fleet_node_device` rows for the revoked node in the same transaction.

## Deferred to PR 2 (#235)
- `fleetnodecontrol.Registry` (in-memory ControlStream + per-`command_id` event dispatch).
- `FleetNodeGateway.ControlStream` bidi handler.
- `FleetNodeAdmin.DiscoverOnFleetNode` operator-initiated discovery + proto `max_items` caps on `DiscoverRequest` modes.
- The `command_id` correlation hook in `ReportDiscoveredDevices` that fans batches to the operator's waiting stream.

## Test plan
- [x] `cd server && go build ./... && go vet ./...`
- [x] `just lint` clean (buf, eslint, golangci-lint)
- [x] `DB_PASSWORD=fleet go test ./internal/handlers/middleware/... ./internal/handlers/fleetnodeadmin/... ./internal/handlers/fleetnodegateway/... ./internal/domain/fleetnodepairing/... ./internal/domain/fleetnodeenrollment/...` — all green
- [ ] CI green


server/sqlc/queries/fleetnodepairing.sql
```diff
@@ -0,0 +1,67 @@
...
+    ip_address,
+    port,
+    url_scheme,
+    driver_name,
+    model,
+    manufacturer,
+    firmware_version,
+    is_active
+)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, TRUE)
```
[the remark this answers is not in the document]
ankitgoswami: handled in next PR

server/sqlc/queries/fleetnodepairing.sql
```diff
@@ -0,0 +1,67 @@
...
+ON CONFLICT (org_id, device_identifier) WHERE deleted_at IS NULL DO UPDATE SET
```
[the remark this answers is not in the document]
ankitgoswami: fixed

server/internal/domain/fleetnodepairing/service.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("ip_address %q is not in a private range (RFC1918/RFC4193)", r.IPAddress)
+	}
+	port, err := strconv.Atoi(r.Port)
+	if err != nil || port < 1 || port > 65535 {
+		return fmt.Errorf("port %q is not in 1-65535", r.Port)
+	}
+	switch r.URLScheme {
+	case "http", "https", "virtual":
+	default:
+		return fmt.Errorf("url_scheme %q is not allowed (http, https, virtual)", r.URLScheme)
```
rongxin-liu: Should probably add `"virtual"` to the proto `in:[...]` list and regenerate, or, if virtual shouldn't be accepted, revert this to `case "http", "https":`?

server/internal/domain/fleetnodepairing/service.go
```diff
@@ -0,0 +1,138 @@
...
+		rows, upErr := s.store.UpsertDiscoveredDeviceFromFleetNode(ctx, orgID, fleetNodeID, r)
+		if upErr != nil {
+			return accepted, rejectedOwnership, fleeterror.LogInternal(component, "upsert discovered device", clientErrUpsertDiscoveredDevice, upErr)
+		}
+		if rows == 0 {
+			rejectedOwnership++
+			continue
+		}
+		accepted++
+	}
```
rongxin-liu, line 124: This batch loop doesn't seem to be atomic. A mid-batch failure leaves a committed prefix. Each `UpsertDiscoveredDeviceFromFleetNode` auto-commits, and there's no surrounding transaction. Every field-level error is already caught atomically by the validate interceptor before the handler, so the one check that can still fail here is `validateReport`'s `IsPrivate()` test.

Worth validating all reports up front (reject the batch before any write), or wrapping the loop in `s.transactor.RunInTx`.

server/internal/handlers/fleetnodegateway/handler.go
```diff
@@ -73,3 +75,31 @@ func (h *Handler) UploadHeartbeat(ctx context.Context, _ *connect.Request[pb.Upl
...
+			FirmwareVersion:  d.GetFirmwareVersion(),
+		})
+	}
+	accepted, _, err := h.pairing.UpsertDiscoveredDevices(ctx, subject.FleetNodeID, subject.OrgID, reports)
+	if err != nil {
+		return nil, err
+	}
+	return connect.NewResponse(&pb.ReportDiscoveredDevicesResponse{
+		AcceptedCount: accepted,
+	}), nil
```
rongxin-liu: Worth adding `rejected_count` to the response (additive, non-breaking), and/or `slog.Warn` with `fleet_node_id`/`org_id` when `rejectedOwnership > 0`?

server/sqlc/queries/fleetnodepairing.sql
```diff
@@ -0,0 +1,67 @@
...
+    SELECT 1
+    FROM device d
+    LEFT JOIN fleet_node_device fnd
+        ON fnd.device_id = d.id
+       AND fnd.org_id = d.org_id
+       AND fnd.fleet_node_id = $10
+    WHERE d.discovered_device_id = discovered_device.id
+      AND d.org_id = discovered_device.org_id
+      AND d.deleted_at IS NULL
+      AND fnd.fleet_node_id IS NULL
```
rongxin-liu: This guard seems to block more than its comment claims, and freezes discovery refresh after unpair.

The comment (lines 2–4) says it blocks a device *"already paired with a different fleet_node,"* but the predicate (`... WHERE fnd.fleet_node_id IS NULL`) blocks whenever a promoted `device` row isn't paired to the *reporting* node — including one paired to no node. So after `UnpairDevice` (which drops the pairing but keeps the `device` row), the originating node can never refresh that `discovered_device`: `is_active` stays `TRUE` and `last_seen`/IP freeze.

If the goal is only to block hijacking a device pinned to a *different* node, gate on `EXISTS` a `fleet_node_device` with `fleet_node_id <> $10` instead, but either way, worth aligning the comment with the code.

rongxin-liu (approved):

server/internal/domain/fleetnodepairing/service.go
ankitgoswami: Added `"virtual"` to the proto allowlist + regenerated (commit `35e7a64a`). Both layers now accept it.

server/internal/domain/fleetnodepairing/service.go
ankitgoswami: Fixed in commit `35e7a64a`: `UpsertDiscoveredDevices` now validates every report up front (validateReport is pure / O(n)), then wraps the writes in `s.transactor.RunInTx` so a mid-batch failure can't leave a committed prefix. Ownership-rejected rows (0 rows affected) still count toward `rejectedOwnership` without aborting the tx — that's the store's normal 'we refused to overwrite a hijacked row' signal, not an error. New test `TestUpsertDiscoveredDevices_BatchValidationErrorRollsBack` exercises the rollback path.

server/internal/handlers/fleetnodegateway/handler.go
ankitgoswami: Done in commit `35e7a64a`: added `rejected_count` to `ReportDiscoveredDevicesResponse` (additive proto3 field) and the gateway handler now emits `slog.Warn` with `fleet_node_id` / `org_id` / `rejected_count` when `rejectedOwnership > 0`.

server/sqlc/queries/fleetnodepairing.sql
ankitgoswami: Good catch — fixed in commit `35e7a64a`. Rewrote the predicate as `NOT EXISTS (... JOIN fnd ... AND fnd.fleet_node_id <> $10 ...)` so it only blocks when paired with a *different* fleet node, matching the intent in the comment. Added `TestUpsertDiscoveredDevices_RefreshesUnpairedDeviceFromOriginatingNode` to lock in the new behavior (renamed + inverted the previous `RejectsRetargetOfLocallyPairedDevice` test, which was asserting the old over-blocking behavior).

server/internal/domain/fleetnodepairing/service.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, r := range reports {
+			rows, upErr := s.store.UpsertDiscoveredDeviceFromFleetNode(ctx, orgID, fleetNodeID, r)
+			if upErr != nil {
+				return fleeterror.LogInternal(component, "upsert discovered device", clientErrUpsertDiscoveredDevice, upErr)
+			}
+			if rows == 0 {
+				rejectedOwnership++
+				continue
+			}
+			accepted++
```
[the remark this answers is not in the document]
ankitgoswami: fixed in next PR

server/internal/domain/fleetnodepairing/service.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	port, err := strconv.Atoi(r.Port)
+	if err != nil || port < 1 || port > 65535 {
+		return fmt.Errorf("port %q is not in 1-65535", r.Port)
+	}
+	// URL scheme is bounded by proto-validate (max_len 32). The gateway
+	// deliberately matches the source schema's permissive stance rather
+	// than gating on a fixed allowlist — plugins like the virtual driver
+	// emit non-http schemes, and the cloud-facing pairing.Device proto
+	// accepts any string.
```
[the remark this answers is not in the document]
ankitgoswami: fixed in next PR

Issue #172, opened 2026-05-05
Curtailment: frontend


negarn:
Add the Proto Fleet frontend for manual curtailment v1.

Scope:
- Add an Energy route with curtailment status, active event details, event history, filtering/sorting, and stop/update-restore actions.
- Wire the generated `CurtailmentService` client through a typed hook for get/list/preview/start/stop/update.
- Add the start flow for fixed-kW curtailment: target kW, tolerance, normal/emergency priority, duration and restore controls, rack/group/miner scope selection, preview, reason validation, and explicit maintenance override confirmation.
- Surface active curtailment globally in the page header and add activity labels/icons for curtailment audit events.
- Cover the API mapping, planner modal, Energy page, and header pill with focused tests/stories.

Out of scope for this frontend ticket: webhook-triggered curtailment, closed-loop modes, and partial/efficiency curtailment levels.

negarn:
🤖 Created split issue #341 for the mapper extraction PR. GitHub returned a duplicate-sub-issue error when attaching it through the sub-issue API, so linking it here for parent tracking: https://github.com/block/proto-fleet/issues/341

negarn:
🤖 Created split issue #343 and PR #344 for the start modal/control support slice. GitHub still returns inconsistent sub-issue API readback, so linking them here for parent tracking:
- Issue: https://github.com/block/proto-fleet/issues/343
- PR: https://github.com/block/proto-fleet/pull/344

PR #338, opened 2026-05-29
Curtailment: share active event polling data

### LINKED ISSUE #337: Curtailment: share active event polling data
🤖 Sub-issue of #172. Split from the gh-172 curtailment frontend branch as the first under-500-line review slice.

Scope:
- Add a shared active curtailment data cache that coalesces concurrent refreshes.
- Preserve restoring/restored active events when active polling briefly returns empty.
- Move page-header curtailment polling onto the shared cache and poll more frequently while an active event is present.

Out of scope:
- /energy route wiring.
- Energy nav menu entry.
- Main energy page UI changes.

Validation:
- Run the active curtailment data and page-header polling unit tests.

negarn:
🤖

## Summary
- Add a shared active curtailment data cache that coalesces active-event refreshes
- Move page-header curtailment polling onto the shared cache, including faster polling while an active event is present

## Test plan
- [ ] Run the active curtailment data and page-header polling unit tests
- [ ] Confirm this split does not include /energy route or nav menu wiring

Refs #172
Closes #337

### DIFF
--- a/client/src/protoFleet/api/activeCurtailmentData.test.ts
+++ b/client/src/protoFleet/api/activeCurtailmentData.test.ts
@@ -1,0 +1,94 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { create } from "@bufbuild/protobuf";
+
+import {
+  applyActiveCurtailmentEvent,
+  fetchActiveCurtailmentData,
+  getActiveCurtailmentSnapshot,
+  refreshActiveCurtailmentData,
+  resetActiveCurtailmentData,
+} from "@/protoFleet/api/activeCurtailmentData";
+import {
+  type CurtailmentEvent,
+  CurtailmentEventSchema,
+  CurtailmentEventState,
+} from "@/protoFleet/api/generated/curtailment/v1/curtailment_pb";
+
+const { mockGetActiveCurtailment } = vi.hoisted(() => ({
+  mockGetActiveCurtailment: vi.fn(),
+}));
+
+vi.mock("@/protoFleet/api/clients", () => ({
+  curtailmentClient: {
+    getActiveCurtailment: mockGetActiveCurtailment,
+  },
+}));
+
+function curtailmentEvent(eventUuid: string, state = CurtailmentEventState.ACTIVE): CurtailmentEvent {
+  return create(CurtailmentEventSchema, {
+    eventUuid,
+    state,
+  });
+}
+
+describe("activeCurtailmentData", () => {
+  beforeEach(() => {
+    resetActiveCurtailmentData();
+    vi.clearAllMocks();
+  });
+
+  it("keeps newer active curtailment data when an older refresh returns later", async () => {
+    let resolveRefresh: (value: { event: CurtailmentEvent }) => void = () => {};
+    mockGetActiveCurtailment.mockReturnValue(
+      new Promise<{ event: CurtailmentEvent }>((resolve) => {
+        resolveRefresh = resolve;
+      }),
+    );
+
+    const refreshPromise = refreshActiveCurtailmentData();
+    applyActiveCurtailmentEvent(curtailmentEvent("newer-event"));
+    resolveRefresh({ event: curtailmentEvent("older-event") });
+
+    await refreshPromise;
+
+    expect(getActiveCurtailmentSnapshot().event?.eventUuid).toBe("newer-event");
+  });
+
+  it("shares one active curtailment request across concurrent refresh callers", async () => {
+    let resolveRefresh: (value: { event: CurtailmentEvent }) => void = () => {};
+    mockGetActiveCurtailment.mockReturnValue(
+      new Promise<{ event: CurtailmentEvent }>((resolve) => {
+        resolveRefresh = resolve;
+      }),
+    );
+
+    const refreshPromise = refreshActiveCurtailmentData();
+    const pendingRefreshPromise = fetchActiveCurtailmentData();
+
+    expect(mockGetActiveCurtailment).toHaveBeenCalledTimes(1);
+
+    resolveRefresh({ event: curtailmentEvent("shared-event") });
+    const [snapshot, pendingRefresh] = await Promise.all([refreshPromise, pendingRefreshPromise]);
+
+    expect(snapshot.event?.eventUuid).toBe("shared-event");
+    expect(pendingRefresh.event?.eventUuid).toBe("shared-event");
+
+    pendingRefresh.commit();
+
+    expect(getActiveCurtailmentSnapshot().event?.eventUuid).toBe("shared-event");
+    expect(mockGetActiveCurtailment).toHaveBeenCalledTimes(1);
+  });
+
+  it.each([
+    ["restoring", CurtailmentEventState.RESTORING],
+    ["restored", CurtailmentEventState.COMPLETED],
+    ["incomplete restore", CurtailmentEventState.COMPLETED_WITH_FAILURES],
+  ])("preserves a %s curtailment when active polling briefly returns empty", async (eventUuid, state) => {
+    applyActiveCurtailmentEvent(curtailmentEvent(eventUuid, state));
+    mockGetActiveCurtailment.mockResolvedValue({ event: undefined });
+
+    await refreshActiveCurtailmentData();
+
+    expect(getActiveCurtailmentSnapshot().event?.eventUuid).toBe(eventUuid);
+  });
+});
--- a/client/src/protoFleet/api/activeCurtailmentData.test.ts
+++ b/client/src/protoFleet/api/activeCurtailmentData.test.ts
@@ -1,5 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { create } from "@bufbuild/protobuf";
+import { Code, ConnectError } from "@connectrpc/connect";
 
 import {
   applyActiveCurtailmentEvent,
@@ -18,18 +19,10 @@ import {
 const { mockGetActiveCurtailment } = vi.hoisted(() => ({
   mockGetActiveCurtailment: vi.fn(),
 }));
-
-vi.mock("@/protoFleet/api/clients", () => ({
-  curtailmentClient: {
-    getActiveCurtailment: mockGetActiveCurtailment,
-  },
-}));
+vi.mock("@/protoFleet/api/clients", () => ({ curtailmentClient: { getActiveCurtailment: mockGetActiveCurtailment } }));
 
 function curtailmentEvent(eventUuid: string, state = CurtailmentEventState.ACTIVE): CurtailmentEvent {
-  return create(CurtailmentEventSchema, {
-    eventUuid,
-    state,
-  });
+  return create(CurtailmentEventSchema, { eventUuid, state });
 }
 
 describe("activeCurtailmentData", () => {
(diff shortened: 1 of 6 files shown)


client/src/protoFleet/components/PageHeader/useCurtailmentPillData.ts
```diff
@@ -46,23 +50,13 @@ export function useCurtailmentPillData(): UseCurtailmentPillDataResult {
...
         } catch (error) {
           if (isAbortError(error, signal)) {
             return;
           }
 
-          handleAuthErrors({
-            error,
-            onError: () => setActiveEvent(null),
-          });
+          handleAuthErrors({ error });
```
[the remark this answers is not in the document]
negarn: 🤖 Addressed: restored the non-abort error path so `handleAuthErrors` clears the shared active curtailment event, with regression coverage for stale pill removal.

client/src/protoFleet/api/activeCurtailmentData.ts
```diff
@@ -0,0 +1,195 @@
...
+): ActiveCurtailmentSnapshot {
+  if (snapshot.event?.eventUuid && snapshot.event.eventUuid === dismissedEventUuid) {
+    snapshot = initialSnapshot;
+  } else if (snapshot.event?.eventUuid) {
+    dismissedEventUuid = null;
+  }
+
+  if (writeVersion < appliedWriteVersion) {
+    return getActiveCurtailmentSnapshot();
+  }
```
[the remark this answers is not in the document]
negarn: 🤖 Addressed: the write-version guard now runs before dismissal state changes, and the tests cover stale refreshes preserving dismissed-event suppression.

client/src/protoFleet/api/activeCurtailmentData.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  inFlightActiveCurtailmentRequest = request;
+  return request;
+}
+
+function releaseActiveCurtailmentRequestSubscriber(request: InFlightActiveCurtailmentRequest): void {
+  request.subscribers = Math.max(0, request.subscribers - 1);
+  if (request.subscribers === 0 && !request.settled) {
+    request.abortController.abort();
+  }
```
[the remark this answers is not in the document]
negarn: 🤖 Addressed: a zero-subscriber abort now clears the shared in-flight request before aborting, with regression coverage that a later caller starts a fresh fetch.

negarn:
🤖 Addressed the Codex Security Review findings in `cfbace63`:

- Bounded empty active-curtailment preservation: restore-related events survive one transient empty active response, then clear on the next empty response.
- Reset active-curtailment cache on logout and successful login so stale curtailment state does not cross auth sessions.
- Kept the previous zero-subscriber abort fix for canceled in-flight request reuse.

Validation run locally:
- `/Users/negar/Development/proto-fleet/bin/npm exec -- vitest run src/protoFleet/api/activeCurtailmentData.test.ts src/protoFleet/components/PageHeader/useCurtailmentPillData.test.tsx`
- `/Users/negar/Development/proto-fleet/bin/npx tsc --noEmit`
- `/Users/negar/Development/proto-fleet/bin/npm run lint`

client/src/protoFleet/api/activeCurtailmentData.ts
```diff
@@ -0,0 +1,198 @@
...
+  subscribers: number;
+}
+
+const initialSnapshot: ActiveCurtailmentSnapshot = { event: undefined };
+const initialStoreState: ActiveCurtailmentDataStore = {
+  ...initialSnapshot,
+  version: 0,
+};
+
+const useActiveCurtailmentDataStore = createStore<ActiveCurtailmentDataStore>(() => initialStoreState);
```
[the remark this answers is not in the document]
negarn: 🤖 Addressed in `cfbace63`: the shared active curtailment cache now resets on `auth.logout()` and before a successful login stores the new authenticated session, so stale curtailment state cannot carry across auth boundaries.

client/src/protoFleet/api/useLogin.ts
```diff
@@ -47,6 +48,7 @@ const useLogin = () => {
 
           // Session cookie is automatically stored by browser
           // We just track the expiry and user info in state
+          resetActiveCurtailmentData();
```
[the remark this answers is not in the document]
negarn: 🤖 Addressed in `48df9a16`: `useLogin` now resets the active curtailment cache only for normal login flows. Same-session reauth calls that pass `skipLogoutOnError` keep the existing cache, so the header pill does not disappear during settings password verification.

negarn:
🤖 Addressed the updated Codex Security Review reset/write-version finding in `94d401b9`. `resetActiveCurtailmentData()` now advances `appliedWriteVersion` instead of resetting write counters to zero, so any pre-reset pending refresh commit is older than the reset and is rejected by the stale-write guard.

Validation run locally:
- `/Users/negar/Development/proto-fleet/bin/npm exec -- vitest run src/protoFleet/api/activeCurtailmentData.test.ts src/protoFleet/components/PageHeader/useCurtailmentPillData.test.tsx src/protoFleet/features/settings/components/Auth.test.tsx`
- `/Users/negar/Development/proto-fleet/bin/npx tsc --noEmit`
- `/Users/negar/Development/proto-fleet/bin/npm run lint`

client/src/protoFleet/components/PageHeader/useCurtailmentPillData.ts
```diff
@@ -46,23 +54,13 @@ export function useCurtailmentPillData(): UseCurtailmentPillDataResult {
...
         } catch (error) {
           if (isAbortError(error, signal)) {
             return;
           }
 
-          handleAuthErrors({
-            error,
-            onError: () => setActiveEvent(null),
-          });
+          handleAuthErrors({ error, onError: () => applyActiveCurtailmentEvent(undefined) });
```
rongxin-liu, line 63: `resetActiveCurtailmentData()` aborts the shared controller, but this checks `isAbortError(error, signal)` against the hook's own signal, so a reset while the hook stays mounted (login-success, or the upcoming `/energy` consumer) surfaces as `ConnectError(Canceled)` and falls through to `handleAuthErrors`. Benign today since logout unmounts the hook, but maybe worth handling now: perhaps treat `Canceled` as an abort regardless of the signal, or have reset reject in-flight subscribers with a `DOMException` `AbortError` so this branch catches it.

client/src/protoFleet/api/activeCurtailmentData.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  if (snapshot.event?.eventUuid && snapshot.event.eventUuid === dismissedEventUuid) {
+    snapshot = initialSnapshot;
+  } else if (snapshot.event?.eventUuid) {
+    dismissedEventUuid = null;
+  }
+
+  appliedWriteVersion = writeVersion;
+  useActiveCurtailmentDataStore.setState({ event: snapshot.event, version: writeVersion });
```
rongxin-liu: `version` is written here on every commit but never read back (the guard uses the module-level `appliedWriteVersion`), and because each poll brings a new `event` object this `setState` re-renders the pill every 3s even when nothing changed. Perhaps worth dropping `version` from the store and skipping the write when the event is unchanged — e.g. bail out early if `eventUuid` + `state` match the current store value.

client/src/protoFleet/api/activeCurtailmentData.ts
```diff
@@ -0,0 +1,210 @@
...
+  return { event };
+}
+
+export function useActiveCurtailmentEvent(): ProtoCurtailmentEvent | undefined {
+  return useActiveCurtailmentDataStore((state) => state.event);
+}
+
+export function applyActiveCurtailmentEvent(event?: ProtoCurtailmentEvent): ActiveCurtailmentSnapshot {
+  return setActiveCurtailmentSnapshot({ event });
+}
```
rongxin-liu, line 99: Maybe worth zeroing the counter whenever the stored event is cleared — here, or in `setActiveCurtailmentSnapshot` when `snapshot.event` is undefined.

client/src/protoFleet/store/slices/authSlice.ts
```diff
@@ -70,6 +71,7 @@ export const createAuthSlice: StateCreator<FleetStore, [["zustand/immer", never]
 
   logout: () =>
     set((state) => {
+      resetActiveCurtailmentData();
```
rongxin-liu: Worth moving it just before/after `set(...)`; the `useLogin` call site already does it outside `set()`.

client/src/protoFleet/api/activeCurtailmentData.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    eventUuid,
+    state,
+  });
+}
+
+describe("activeCurtailmentData", () => {
+  beforeEach(() => {
+    resetActiveCurtailmentData();
+    vi.clearAllMocks();
+  });
```
rongxin-liu, line 32: The auth-boundary reset and the version guard are the parts the earlier review hardened, but they aren't exercised yet. Maybe worth adding a few here: cache clears on `logout()` / login (`!skipLogoutOnError`) and is kept on reauth; and a stale `commit()` can't overwrite a newer `reset()`, a reset mid-flight discards the in-flight result, and a dismissed event can reappear after reset.

rongxin-liu (approved):

client/src/protoFleet/components/PageHeader/useCurtailmentPillData.ts
negarn: 🤖 Addressed Rongxin: reset-aborted shared requests now normalize the shared-controller `ConnectError(Code.Canceled)` into a `DOMException` `AbortError`, so mounted consumers treat reset cancellation as benign. Added targeted regression coverage for reset-aborted shared requests.

client/src/protoFleet/api/activeCurtailmentData.ts
negarn: 🤖 Addressed Rongxin: removed the unread store `version` field and now skip the Zustand write when the refreshed protobuf event equals the current event. The module-level write version still advances before the no-op return, so stale commits remain rejected.

client/src/protoFleet/api/activeCurtailmentData.ts
negarn: 🤖 Addressed Rongxin: clearing the stored active event now also resets `emptyActiveRefreshCount`, so the one-empty-response preservation window cannot leak across explicit clears or resets.

client/src/protoFleet/store/slices/authSlice.ts
negarn: 🤖 Addressed Rongxin: moved `resetActiveCurtailmentData()` outside the Zustand `set()` mutation in `logout`, matching the login flow side-effect placement.

client/src/protoFleet/api/activeCurtailmentData.test.ts
negarn: 🤖 Addressed Rongxin within the 500-line split cap: added targeted regression coverage for reset-aborted shared requests, which is the highest-risk uncovered path from this follow-up. I kept the broader auth-boundary/write-version test matrix out of this first split to avoid growing the PR over the agreed review size; the implementation changes are covered by the targeted suite plus lint and typecheck.

PR #347, opened 2026-06-01
feat(authz): gate ActivityService with activity:read

mcharles-square:
## Summary

- Adds the `activity:read` catalog key and gates all three `ActivityService` RPCs (`ListActivities`, `ExportActivities`, `ListActivityFilterOptions`) with `middleware.RequirePermission`.
- Migration `000065` seeds the permission row and backfills it onto every existing `ADMIN` role.
- Drops the local `getSessionInfo` helper — `RequirePermission` now covers both the authentication check and the catalog-key check.

## Why

These three procedures were the last "ungated, authenticated read" entry in `ProceduresPendingMigration`. The activity log is the org-wide audit trail; gating it cleans up an ungated read surface in the same shape as the pools/schedules slices.

## Behavior

| RPC | Key |
|-----|-----|
| `ListActivities` | `activity:read` |
| `ExportActivities` | `activity:read` |
| `ListActivityFilterOptions` | `activity:read` |

- **`SUPER_ADMIN`** picks up the key automatically at startup via the full-mode reconciler.
- **`ADMIN`** receives the key: new orgs through `adminSeedPermissions()`, existing orgs through migration `000065`.
- **`FIELD_TECH`** does not receive `activity:read` by design; operators opt in via the role editor.

## Out of scope / follow-ups

- Activity nav guard on the client. The `/activity` route is mounted unconditionally in `navItems.ts:primaryNavItems` — FIELD_TECH and similar sessions will now hit `PermissionDenied` on the initial `ListActivities` call. Same situation as the Schedules/Pools nav gap; the proper fix needs the U9/U10 permission-key plumbing called out in the RBAC plan. Worth a follow-up.

## Test plan
- [x] `go build ./...` clean.
- [x] Catalog tests pass: `TestCatalogCompleteness`, `TestResourceOrder_*`.
- [x] Contract tests pass: `TestRPCContract_EveryRegisteredProcedureIsClassified`, `TestRPCContract_ProcedurePermissionsKeysAreInCatalog`.
- [x] Activity handler tests pass (28 cases; existing `authedCtx` now seeds `activity:read`).
- [ ] DB-backed seed/reconcile tests — run in CI.
- [ ] E2E smoke: role without `activity:read` rejects all three activity RPCs with `PermissionDenied`.

### DIFF
--- a/server/internal/handlers/middleware/rpc_permissions.go
+++ b/server/internal/handlers/middleware/rpc_permissions.go
@@ -45,6 +45,13 @@ import (
 // glance: shrinking ProceduresPendingMigration to zero is the exit
 // criterion for retiring the legacy RequireAdmin middleware.
 var ProcedurePermissions = map[string]string{
+	// Activity log — read-only audit trail. Export is the CSV variant of
+	// the same query; filter options is the lookup endpoint that drives
+	// the UI's filter panel. All three sit on activity:read.
+	activityv1connect.ActivityServiceListActivitiesProcedure:            authz.PermActivityRead,
+	activityv1connect.ActivityServiceExportActivitiesProcedure:          authz.PermActivityRead,
+	activityv1connect.ActivityServiceListActivityFilterOptionsProcedure: authz.PermActivityRead,
+
 	// API key management — gated by RequirePermission(PermAPIKeyManage).
 	apikeyv1connect.ApiKeyServiceCreateApiKeyProcedure: authz.PermAPIKeyManage,
 	apikeyv1connect.ApiKeyServiceListApiKeysProcedure:  authz.PermAPIKeyManage,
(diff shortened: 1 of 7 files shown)


server/internal/handlers/middleware/rpc_permissions.go
```diff
@@ -45,6 +45,13 @@ import (
 // glance: shrinking ProceduresPendingMigration to zero is the exit
 // criterion for retiring the legacy RequireAdmin middleware.
 var ProcedurePermissions = map[string]string{
+	// Activity log — read-only audit trail. Export is the CSV variant of
+	// the same query; filter options is the lookup endpoint that drives
+	// the UI's filter panel. All three sit on activity:read.
+	activityv1connect.ActivityServiceListActivitiesProcedure:            authz.PermActivityRead,
```
rongxin-liu, line 51: No strong feeling, but now that all three `ActivityService` procedures are gated on `activity:read`, worth mirroring it on the client nav as well?

rongxin-liu (approved):
Looks good!

server/internal/handlers/middleware/rpc_permissions.go
mcharles-square: This will be addressed in a subsequent PR

PR #350, opened 2026-06-01
Curtailment: refine page header status pill

### LINKED ISSUE #349: Curtailment: refine page header status pill
Parent: #172

Split out the page-header curtailment pill state mapping so active curtailments can distinguish pending, curtailing, curtailed, and restoring states before the full Energy page route is exposed.

Scope:
- Add shared display-state helpers for curtailment target rollups and observed reduction.
- Map the page-header curtailment pill to the richer active display states.
- Keep the route and menu exposure for the final release split.

negarn:
## Summary
- Add shared curtailment display-state helpers for target rollups and observed reduction
- Map the page-header curtailment pill to pending, curtailing, curtailed, and restoring states without exposing the Energy route

## Test plan
- [ ] Confirm the page-header curtailment pill shows the expected active state label and details
- [ ] Confirm there are no route, nav, or Energy page entry-point changes in this split

Closes #349

### DIFF
--- a/client/src/protoFleet/features/energy/curtailmentDisplayUtils.ts
+++ b/client/src/protoFleet/features/energy/curtailmentDisplayUtils.ts
@@ -112,6 +112,10 @@ interface ActiveCurtailmentDisplayEvent {
   rollups: CurtailmentTargetRollup[];
 }
 
+interface ActiveCurtailmentDisplayOptions {
+  dispatchStartedAsCurtailing?: boolean;
+}
+
 export interface ActiveCurtailmentMinerCompliance {
   curtailedCount: number;
   restoreFailedCount: number;
@@ -358,13 +362,16 @@ function isRestoreIncompleteEventState(state: CurtailmentEventState): boolean {
   return state === "completedWithFailures";
 }
 
-export function getActiveCurtailmentDisplayState(event: ActiveCurtailmentDisplayEvent): ActiveCurtailmentDisplayState {
+export function getActiveCurtailmentDisplayState(
+  event: ActiveCurtailmentDisplayEvent,
+  options: ActiveCurtailmentDisplayOptions = {},
+): ActiveCurtailmentDisplayState {
   if (event.state === "restoring") {
     return "restoring";
   }
 
   if (event.state === "pending") {
-    return hasActiveCurtailmentDispatchStarted(event) ? "curtailing" : "pending";
+    return options.dispatchStartedAsCurtailing && hasActiveCurtailmentDispatchStarted(event) ? "curtailing" : "pending";
   }
 
   if (isRestoreIncompleteEventState(event.state)) {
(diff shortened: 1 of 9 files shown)


client/src/protoFleet/features/energy/curtailmentDisplayUtils.ts
```diff
@@ -135,6 +308,89 @@ export function getCurtailmentTargetKw(event: CurtailmentTargetKwEvent): number
...
+}
+
+export function getActiveCurtailmentDisplayState(event: ActiveCurtailmentDisplayEvent): ActiveCurtailmentDisplayState {
+  if (event.state === "restoring") {
+    return "restoring";
+  }
+
+  if (event.state === "pending") {
+    return hasActiveCurtailmentDispatchStarted(event) ? "curtailing" : "pending";
+  }
```
rongxin-liu, line 375: This `pending` branch behaves differently from the same-named `getActiveCurtailmentDisplayState` in `ActiveCurtailmentStatus.tsx` (around line 307), which always returns `pending`. So once targets start dispatching, the header pill reads as `curtailing` while the Energy page still shows `pending` for the same event.

If that's the intended refinement, perhaps worth making it an explicit option (e.g. a `dispatchStartedAsCurtailing` flag on the shared helper) so it reads as a deliberate choice — otherwise it's easy for someone consolidating these two later to flip the Energy page's behavior by accident.

client/src/protoFleet/features/energy/curtailmentDisplayUtils.ts
```diff
@@ -51,8 +52,92 @@ export const activeCurtailmentEventStates = [
...
+  | "confirmed"
+  | "drifted"
+  | "resolved"
+  | "released"
+  | "restoreFailed";
+
+export interface CurtailmentTargetRollup {
+  state: CurtailmentTargetState;
+  count: number;
+}
```
rongxin-liu, line 67: Looks like most of this new shared block — `CurtailmentTargetState` / `CurtailmentTargetRollup`, the `*TargetStates` arrays, and `getRollupCount` / `getMinerCompliance` / `getProgressPercent` / `getActiveCurtailmentDisplayState` — duplicates logic that already lives in `ActiveCurtailmentStatus.tsx`, which still uses its own local copies.

The Energy route already imports a few helpers from this module, so maybe worth pointing it at these as well so there's a single source of truth. Right now the two `CurtailmentTargetState` / `CurtailmentTargetRollup` definitions only line up structurally (that's what lets the `rollups` assignment in `curtailmentMappers.ts` compile), so they'd silently break if either list gains a state.

If the Energy-route migration is out of scope for this split, perhaps a quick follow-up to consolidate?

rongxin-liu (approved):
Looks good overall! My comments mostly about consolidating codes.

client/src/protoFleet/features/energy/curtailmentDisplayUtils.ts
negarn: 🤖 Thanks, agreed on the direction. I’m keeping the Energy-route component migration out of this split so PR #350 stays under the 500-line review cap and because the route exposure is intentionally being held for the final split. This branch now makes the header-specific pending behavior explicit via `dispatchStartedAsCurtailing`; the parent gh-172 branch is rebased on top of it, so the next Energy-component split can point `ActiveCurtailmentStatus` at these shared helpers without duplicating this work.

PR #361, opened 2026-06-02
Curtailment: share active event API state

### LINKED ISSUE #359: Curtailment: share active event API state
🤖 Split from #172.

## Scope
- Update the curtailment client API hook to use the shared active-curtailment snapshot.
- Add client-side support for repeated history status filters while preserving the legacy single-filter hook API for existing callers.
- Map active events inserted into history with their active display state.

## Review constraint
- Keep this split around the 500-added-line review target using additions-only diff accounting.
- Keep `/energy` route, route prefetch, page export, and menu exposure out of this split.

negarn:
🤖
## Summary
- Share active curtailment API state between the page-header data source and the curtailment management hook
- Add multi-status history filter support in the client hook while preserving the existing single-status hook API for current callers
- Map injected active history rows with their active display state

## Test plan
- [ ] Confirm active curtailment state stays visible in the header and management data after refreshes
- [ ] Confirm history filtering still works for existing single-status callers
- [ ] Confirm multi-status history filters send repeated `state_filters`

Closes #359

### DIFF
--- a/client/src/protoFleet/api/curtailmentMappers.ts
+++ b/client/src/protoFleet/api/curtailmentMappers.ts
@@ -148,7 +148,7 @@ export function mapActiveCurtailmentEvent(event: ProtoCurtailmentEvent): ActiveC
     targetKw: getFixedKwTarget(event),
     observedReductionKw: observedPowerSummary.observedReductionKw,
     remainingPowerKw: observedPowerSummary.remainingPowerKw,
-    restoreBatchSize: event.restoreBatchSize || event.effectiveBatchSize,
+    restoreBatchSize: event.effectiveBatchSize || event.restoreBatchSize,
     restoreBatchIntervalSec: event.restoreBatchIntervalSec,
     rollups: getCurtailmentTargetRollups(event),
   };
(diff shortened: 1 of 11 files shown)


client/src/protoFleet/api/curtailmentMappers.ts
```diff
@@ -152,7 +148,7 @@ export function mapActiveCurtailmentEvent(event: ProtoCurtailmentEvent): ActiveC
     targetKw: getFixedKwTarget(event),
     observedReductionKw: observedPowerSummary.observedReductionKw,
     remainingPowerKw: observedPowerSummary.remainingPowerKw,
-    restoreBatchSize: event.effectiveBatchSize || event.restoreBatchSize,
+    restoreBatchSize: event.restoreBatchSize || event.effectiveBatchSize,
```
rongxin-liu: Should this stay `effectiveBatchSize || restoreBatchSize` instead?

Per the proto, `effective_batch_size = max(restore_batch_size, ceil(0.01 × selected_target_count))` — it's the value the restorer actually uses, and it feeds both the `N miners every Ns` cadence text and the restore-time estimate (`ceil(selected / restoreBatchSize) × intervalSec`) in `ActiveCurtailmentStatus`.

rongxin-liu (approved):

client/src/protoFleet/api/curtailmentMappers.ts
negarn: okay. it's just weird when the user sets it to some value and then they see a different value later on

PR #389, opened 2026-06-04
feat(fleetnode): carry concurrent commands over the ControlStream

ankitgoswami:
## Summary

Groundwork for **sending commands to miners paired via fleet nodes**. Today the
fleet-node `ControlStream` carries exactly one thing: server-initiated discovery
(#235). This PR generalizes the control plane so it can carry many concurrent
commands per node, which is the prerequisite for routing operator miner commands
(reboot, curtail, …) down a node's stream via RFC 0001's remote-node `Miner`
adapter (a later PR).

**No behavior change to discovery.** This is Phase 1 of a 3-phase effort; it ships
no user-visible capability on its own.

## What changed

- **Typed `AgentCommand` envelope** (`proto/pairing/v1`): the `ControlCommand.payload`
  is now a `pairing.v1.AgentCommand` oneof so the node can tell command kinds apart.
  Discovery migrates into the `discover` arm (done exactly once, on both the admin
  send side and the node decode side). Fields `2`/`3` are left for the upcoming
  miner-command and pairing arms, coordinated with the parallel pairing effort so the
  discovery payload is never migrated twice.
- **Registry** (`internal/domain/fleetnode/control`): replaces the
  single-in-flight-command-per-node model with a `command_id`-keyed map.
  *Report-bearing* commands (discovery, and later pairing) stream batches + a terminal
  ack on an `events` channel and admit device reports against the existing scope/quota;
  *ack-only* commands (per-miner commands) take their terminal ack on a per-command
  channel via a new blocking `SendCommand`. The outbound queue is buffered, and teardown
  frees every in-flight command. Channel-ownership invariants are preserved.
- **Node** (`cmd/fleetnode/control.go`): replaces the single command worker with a
  bounded (16) worker pool, so a long (up to 10-min) discovery no longer
  head-of-line-blocks quick commands and many commands run concurrently; `BUSY` is acked
  only at the pool ceiling. `handleCommand` dispatches on envelope kind.

## Test plan

- `go test -race ./internal/domain/fleetnode/control/... ./cmd/fleetnode/...` — green.
  New/updated coverage: many concurrent commands not rejected, ack-routing by kind,
  `SendCommand` block/disconnect/ctx-cancel/teardown, node worker pool runs a quick
  command while a long discovery is in flight, `BUSY` past the pool ceiling, unknown
  envelope kind → `BAD_REQUEST`.
- DB-backed `DiscoverOnFleetNode` integration suite — green (end-to-end envelope
  round-trip through the real handler + registry).
- Full `server` build, `golangci-lint`, and pre-push `tsc` — clean.

## Follow-ups (later PRs)

- Phase 2: `MinerCommand` proto + remote-node `Miner` adapter + `fleet_node_device`
  resolution branch + node-side executor (end-to-end for the `virtual` driver).
- Phase 3: firmware/logs out-of-band transfer, password credential-edit flow, unpair.


server/cmd/fleetnode/control.go
```diff
@@ -43,6 +43,11 @@ const (
 	maxPortsPerIP          = discoverylimits.MaxPortsPerIP
 	// Mirrors the proto cap so a verbose error doesn't fail buf-validate on the ack itself.
 	maxAckErrorMessageBytes = 4096
+	// commandPoolSize bounds quick per-miner commands handled concurrently per
+	// session. Discovery does not draw from this pool; it has its own exclusive,
+	// single-flight slot (see runControlSession). Commands past the ceiling are
+	// acked BUSY.
+	commandPoolSize = 16
```
flesher, line 50: You probably are planning in follow up when you implement non discovery commands, but we should we add test for ack BUSY with 17 commands

proto/pairing/v1/pairing.proto
```diff
@@ -125,6 +125,19 @@ message DiscoverResponse {
   string error = 2;
 }
 
+// AgentCommand is the typed envelope marshaled into ControlCommand.payload so a
+// fleet node can tell command kinds apart rather than assuming a bare
+// DiscoverRequest. command_id stays on the transport-level
+// ControlCommand/ControlAck and is not duplicated here. Field numbers 2 and 3 are
+// reserved for the upcoming miner-command and pairing arms.
+message AgentCommand {
```
flesher, line 133: probably a non issue at the moment since agents have not shipped, but I assume agent nodes and server will be deployed separately?  would contract changes like this require some sort of way for the nodes to tell the server the structure of the response they expect?  maybe in `ControlHello`

flesher (approved):

server/cmd/fleetnode/control.go
ankitgoswami: yea, that test is in the next PR

proto/pairing/v1/pairing.proto
ankitgoswami: for now deploys will be lockstep. I was thinking the first time we have to make a backwards incompatible change we'd add versioning to ControlHello.

PR #393, opened 2026-06-04
Add Edit role flow + require role on add team member

mcharles-square:
Two related Team-page gaps land together:

1. Add an "Edit role" action so an admin can change a member's org-scope role without redoing CreateUser. Server side this is a new auth.v1.UpdateUserRole RPC gated on user:manage. The domain service locks the new role row FOR UPDATE, runs the standard caller-must-dominate parity check against both the target's current and new effective permissions, refuses to drop the last live SUPER_ADMIN seat, and atomically swaps user_organization + user_organization_role inside one tx. The swap is idempotent — assigning the role the user already holds is a no-op.

2. Remove the FIELD_TECH auto-default from the Add team member modal. The role picker starts empty and Save stays disabled until an admin makes an explicit choice; combined with the server's required role_id contract from the previous PR this closes the "accidentally minted a Field Tech / Admin without meaning to" foot-gun.

Client wires `updateUserRole` into useUserManagement, adds an EditRoleModal that preselects the member's current role (hidden on the caller's own row and on SUPER_ADMIN members), and threads the action through Team.tsx's per-row menu. Test coverage:
- server: validation-branch table tests for every pre-resolver rejection path plus a no-op-same-role assertion; handler gate test confirms user:manage is enforced.
- client: AddTeamMemberModal tests updated for the no-default behavior, with new tests for "Save disabled until role picked" and "Admin role flows through to createUser". New EditRoleModal.test.tsx covers preselect, dirty/clean Save state, update payload, error handling, and SUPER_ADMIN filtering.
- e2e: teamAccounts.spec.ts adds an explicit role pick to every create-user step and a new "Edit team member role" test that promotes a Field Tech to Admin and verifies the row reflects the change.


client/src/protoFleet/features/settings/components/Team.tsx
```diff
@@ -247,6 +266,17 @@ const Team = () => {
         onDismiss={() => setDeactivateUserData(null)}
         isSubmitting={isDeactivating}
       />
+      <EditRoleModal
+        open={!!editRoleUser}
+        userId={editRoleUser?.userId ?? ""}
+        username={editRoleUser?.username ?? ""}
+        currentRoleName={editRoleUser?.role ?? ""}
```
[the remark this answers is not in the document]
mcharles-square: > `EditRoleModal` preselects the current role by matching `currentRoleName` against `listRoles().name`, but `editRoleUser.role` comes from `UserInfo.role` (e.g. "ADMIN" / "FIELD_TECH") and won't match the role display names ("Admin" / "Field Tech").

Not addressing: the premise that the two strings differ isn't true in our schema. `UserInfo.role` is the server's `role.name` column (returned by `GetUserRoleName` → `r.name`). `RoleItem.name` from `ListRoles` is the same column (`role.name`). The seed values are `"SUPER_ADMIN"`, `"ADMIN"`, `"FIELD_TECH"` (see `server/internal/domain/authz/builtin.go`), so they round-trip exactly. `formatRole()` is the *display-only* mapper for the user-facing labels (`Owner`/`Admin`/`Field Tech`); the modal never consumes those.

Adjacent fix in 8c3cefc6: clarified the `currentRoleName` prop doc in `EditRoleModal.tsx` to make this explicit ("Role name from `UserInfo.role` (server role's `name` column); do NOT pass the display label") so a future caller can't reintroduce the mismatch.

client/src/protoFleet/features/settings/components/EditRoleModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+  open?: boolean;
+  /** Target user the swap will apply to. */
+  userId: string;
+  username: string;
+  /**
+   * Role NAME the user currently holds (UserInfo.role on the wire is the
+   * server-side name string). We match by name against the listRoles output
+   * to preselect the picker since `role_id` is not on UserInfo.
+   */
+  currentRoleName: string;
```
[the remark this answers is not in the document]
mcharles-square: > The prop comment says `currentRoleName` is the server-side role name string, but in practice `UserInfo.role` is a role key (e.g. "ADMIN" / "FIELD_TECH") and callers need to pass the *display label* that matches `listRoles().name` for preselect to work. Updating this comment avoids future call sites reintroducing the mismatch.

Addressed: the comment-update direction was right, but inverted — `listRoles().name` *is* the same wire string as `UserInfo.role` ("ADMIN"/"FIELD_TECH"/seed names from `role.name`), not the display label. Tightened the prop doc in 8c3cefc6 to say "Role name from `UserInfo.role` (server role's `name` column); do NOT pass the display label here — those come from `formatRole()` and don't match."

server/internal/domain/auth/service.go
```diff
@@ -949,6 +949,181 @@ func (s *Service) DeactivateUser(ctx context.Context, req *authv1.DeactivateUser
...
+
+	s.logActivity(ctx, activitymodels.Event{
+		Category:       activitymodels.CategoryAuth,
+		Type:           "update_user_role",
+		Description:    fmt.Sprintf("Role updated for user: %s", target.Username),
+		UserID:         &info.ExternalUserID,
+		Username:       &info.Username,
+		OrganizationID: &orgID,
+		Metadata:       map[string]any{"target_user_id": target.UserID, "target_username": target.Username, "role_id": req.RoleId},
+	})
```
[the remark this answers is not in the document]
mcharles-square: > `UpdateUserRole` returns early on a no-op (same role) but still logs an `update_user_role` activity event after the tx. This contradicts the docstring ("no audit event") and will create misleading audit noise for idempotent calls.

Addressed in 8c3cefc6: capture a `swapped` bool from the tx callback and skip `logActivity` when the idempotent short-circuit fires. The same role assignment now returns OK without any audit event or store writes.

server/sqlc/queries/user_organization_role.sql
```diff
@@ -37,6 +37,26 @@ FROM user_organization_role
...
+    uor.role_id      AS role_id,
+    r.builtin_key    AS role_builtin_key
+FROM user_organization_role uor
+JOIN role r ON r.id = uor.role_id
+           AND r.organization_id = uor.organization_id
+WHERE uor.user_id = $1
+  AND uor.organization_id = $2
+  AND uor.scope_type = 'org'
+  AND uor.deleted_at IS NULL
+  AND r.deleted_at IS NULL;
```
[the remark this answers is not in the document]
mcharles-square: > `GetOrgScopeAssignmentForUser` is used inside `UpdateUserRole`'s transaction to decide which assignment row to soft-delete, but it doesn't lock the assignment row. Two concurrent role updates on the same target user (to different new roles) can interleave and leave multiple live org-scope assignments (and a last-writer-wins `user_organization.role_id`).

Addressed in 8c3cefc6: added `FOR UPDATE OF uor` to `GetOrgScopeAssignmentForUser` (only caller is `UpdateUserRole`, which is always inside a tx). Two concurrent swaps on the same target now serialize on this read — the second blocks until the first commits, then re-evaluates against the now-soft-deleted assignment (filtered by `deleted_at IS NULL`) and surfaces `InvalidArgument`. Locking just `uor` (not `role`/`user`) keeps the lock surface minimal.

Same commit also fixes the related last-SUPER_ADMIN race flagged by Codex by replacing `CountOrgScopeSuperAdminsExcludingAssignment` with `LockAndCountOrgScopeSuperAdmins` (CTE + `FOR UPDATE OF uor`), so two concurrent SUPER_ADMIN demotions contend on the same row set rather than locking each other's "remaining" rows.

mcharles-square:
> [Codex Security Review](https://github.com/block/proto-fleet/pull/393#issuecomment-3582275713) — 3 findings on `UpdateUserRole`

All three addressed in `8c3cefc6`:

**[HIGH] UpdateUserRole reachable with user API keys.**
Added `authv1connect.AuthServiceUpdateUserRoleProcedure` to `SessionOnlyProcedures` in `server/internal/handlers/interceptors/config.go`. Added `TestAuthInterceptor_UpdateUserRoleRejectsApiKeyAuth` in `config_test.go` asserting both the registry membership and that API-key auth on the procedure returns `PermissionDenied`.

**[HIGH] Last-SUPER_ADMIN guard is racy.**
Replaced `CountOrgScopeSuperAdminsExcludingAssignment` with `LockAndCountOrgScopeSuperAdmins` — a CTE that takes `FOR UPDATE OF uor` on every live org-scope SUPER_ADMIN assignment in the org before aggregating. Two concurrent demotions of different SUPER_ADMINs now contend on the same row set (instead of each locking the *other* tx's "remaining" row), so the second demotion sees count ≤ 1 after the first commits and fails closed. Caller in `UpdateUserRole` compares `total <= 1` (the assignment being dropped is still live at lock time, so count includes it). Aggregates can't appear directly under `FOR UPDATE`, so the CTE-with-lock pattern is the workable shape.

**[MEDIUM] Permission recheck uses lock-free resolver inside the write tx.**
Added `PermissionResolver.LoadEffectiveForUpdateInTx(ctx, userID, orgID)` — fetches tx-bound queries from ctx (via new shared `db.GetTxQueries` / `db.WithTxQueries` helpers; sqlstores's transactor and connection-manager redirected to the shared key) and runs `LoadEffectiveForUpdate`, which takes `FOR UPDATE` on `uor`/`u`/`r`. `UpdateUserRole` now uses this for both the caller and target permission loads, so a concurrent demotion of the caller serializes against the swap write and either fails the recheck or applies after the commit.

Tests: `go test ./internal/domain/auth/... ./internal/handlers/interceptors/... ./internal/handlers/middleware/...` green; `golangci-lint run` clean. DB-dependent integration tests will exercise the lock paths in CI.

mcharles-square:
> [Codex Security Review #2](https://github.com/block/proto-fleet/pull/393#issuecomment-4623727186) — 2 findings + CI e2e failure

All three addressed in `9a6d8240`:

**[CI fail] teamAccounts.spec.ts e2e timeouts.**
Root cause: the Add/Edit modals' Select rendered `role.name` (the seed identifier — `ADMIN`/`FIELD_TECH`/`SUPER_ADMIN`) as the option label, but the page-object helper looked up the *display* label (`Admin`/`Field Tech`). Fixed by wrapping the label in `formatRole(role.name)` in both `AddTeamMemberModal` and `EditRoleModal` — users now see `Admin`/`Field Tech`/`Owner` while the underlying `value` (role_id) and the preselect match key (`role.name`) stay the canonical wire string. Unit-test mocks updated to use canonical names (`ADMIN`/`FIELD_TECH`) so the test exercises the same transform.

**[Codex HIGH] Last-SUPER_ADMIN guard doesn't coordinate with DeactivateUser.**
Moved the soft-delete in `Service.DeactivateUser` into a `RunInTx` block that runs the same `LockAndCountOrgScopeSuperAdmins` guard introduced for `UpdateUserRole`. When the target holds an org-scope SUPER_ADMIN seat, both paths now contend on the same locked row set, so a concurrent demote-A + deactivate-B can no longer both observe `total > 1` and commit. New test pending in a follow-up — this turn's scope is the production fix; the integration suite exercises the lock paths in CI.

**[Codex MEDIUM] Multi-assignment schema allows the swap to miss other org-scope grants.**
Added migration `000074_user_organization_role_single_live_org_scope` — a partial unique index `(user_id, organization_id) WHERE scope_type='org' AND deleted_at IS NULL` that forbids more than one live org-scope assignment per user. Matches the single-role-per-user invariant the application has always assumed (`CreateUserOrganizationRole` and `UpdateUserRole` both maintain it); the index turns an off-path bypass into a DB-enforced error.

server/migrations/000074_user_organization_role_single_live_org_scope.up.sql
```diff
@@ -0,0 +1,16 @@
...
+-- their role_ids differ — the application has always assumed a single
+-- live org-scope row per user (CreateUserOrganizationRole and
+-- UpdateUserRole both maintain that invariant), and an off-path write
+-- that violated it would silently leave UpdateUserRole swapping only
+-- one assignment while other live grants remained.
+--
+-- Constraint shape mirrors the existing partial index pattern: WHERE
+-- scope_type='org' AND deleted_at IS NULL so soft-deleted rows and
+-- site-scope assignments are excluded.
+CREATE UNIQUE INDEX uq_user_org_single_live_org_scope
```
flesher: probably OK since we dont have a userbase, but other indexes like this use CONCURRENTLY

server/internal/domain/auth/service_test.go
```diff
@@ -1072,3 +1072,200 @@ func TestResolveCreateUserRole_ValidationBranches(t *testing.T) {
 		})
 	}
 }
+
```
flesher, line 1076: these cases are not tested for:

- parity-fail-on-current-role: caller’s effective set doesn’t dominate target’s current → PermissionDenied.
- parity-fail-on-new-role: dominates current but not new → PermissionDenied.
- last-SA refusal: target is SUPER_ADMIN, LockAndCountOrgScopeSuperAdmins returns 1 → FailedPrecondition.
- last-SA passes: same setup but count returns 2 → swap proceeds.

server/internal/domain/auth/service.go
```diff
@@ -931,9 +931,32 @@ func (s *Service) DeactivateUser(ctx context.Context, req *authv1.DeactivateUser
...
+	err = s.transactor.RunInTx(ctx, func(ctx context.Context) error {
+		current, err := s.userManagementStore.GetOrgScopeAssignmentForUser(ctx, user.ID, orgID)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return fleeterror.NewInternalErrorf("error getting target assignment: %v", err)
+		}
+		// If the target holds an org-scope SUPER_ADMIN seat, lock every
+		// live SA assignment in the org and require count > 1 (the
+		// target's own seat is still live at this point).
+		if err == nil && current.BuiltinKey == string(authz.BuiltinKeySuperAdmin) {
+			total, countErr := s.userManagementStore.LockAndCountOrgScopeSuperAdmins(ctx, orgID)
```
flesher, line 957: This has no tests.  should we add a TestDeactivateUser table in service_test.go covering:

- Non-SA target → proceeds to SoftDeleteUser.
- SA target, LockAndCountOrgScopeSuperAdmins returns 1 → FailedPrecondition.
- SA target, count returns 2 → proceeds to SoftDeleteUser.

client/src/protoFleet/features/settings/components/Team.tsx
```diff
@@ -247,6 +275,17 @@ const Team = () => {
         onDismiss={() => setDeactivateUserData(null)}
         isSubmitting={isDeactivating}
       />
+      <EditRoleModal
```
flesher, line 278: this is ok for this PR, but we should align how we render modals across the app.  everything in settings seems to always mount it and toggle `open` for visibility.  In other parts of the app we conditionally render the modals.  its about 50/50 from what i can tell

flesher (approved):

server/migrations/000074_user_organization_role_single_live_org_scope.up.sql
mcharles-square: > probably OK since we dont have a userbase, but other indexes like this use CONCURRENTLY

Addressed in 7bcc1132: switched to `CREATE UNIQUE INDEX CONCURRENTLY` (down migration mirrors with `DROP INDEX CONCURRENTLY IF EXISTS`). Matches the pattern from 000055/000056 and avoids the brief ACCESS EXCLUSIVE on user_organization_role during deploy. Comment updated to reference the same recovery procedure (DROP INVALID + `migrate force 73`) the curtailment migrations document.

server/internal/domain/auth/service_test.go
mcharles-square: > these cases are not tested for:
> - parity-fail-on-current-role: caller's effective set doesn't dominate target's current → PermissionDenied.
> - parity-fail-on-new-role: dominates current but not new → PermissionDenied.
> - last-SA refusal: target is SUPER_ADMIN, LockAndCountOrgScopeSuperAdmins returns 1 → FailedPrecondition.
> - last-SA passes: same setup but count returns 2 → swap proceeds.

Addressed in 7bcc1132. Added `TestUpdateUserRole_ParityAndLastSuperAdmin` covering all four cases. The blocker was that `Service.permResolver` was a concrete `*authz.PermissionResolver` with a hard DB dependency — extracted a small interface (`effectivePermissionResolver` with `LoadEffective` + `LoadEffectiveForUpdateInTx`) that the concrete resolver still satisfies, then stubbed it via a `fakeResolver` in tests so each case can hand-craft caller/target effective sets:

- **parity-fail-on-current**: caller has `user:manage + fleet:read`, target current has the same set → caller doesn't strictly dominate → ForbiddenError.
- **parity-fail-on-new**: caller has `user:manage + fleet:read`, target current has only `fleet:read` (dominates), new role grants `user:manage + role:manage + fleet:read` → caller lacks role:manage → ForbiddenError on the new-role check.
- **last-SA refusal**: target's current assignment has `BuiltinKey = SUPER_ADMIN`, `LockAndCountOrgScopeSuperAdmins` mocked to return 1 → FailedPreconditionError, `UpdateUserOrganizationRole` NOT called.
- **last-SA passes**: same SA setup, count returns 2 → swap proceeds (`UpdateUserOrganizationRole` expectation set on the mock).

server/internal/domain/auth/service.go
mcharles-square: > This has no tests. should we add a TestDeactivateUser table in service_test.go covering:
> - Non-SA target → proceeds to SoftDeleteUser.
> - SA target, LockAndCountOrgScopeSuperAdmins returns 1 → FailedPrecondition.
> - SA target, count returns 2 → proceeds to SoftDeleteUser.

Addressed in 7bcc1132. Added `TestDeactivateUser_LastSuperAdminGuard` with all three cases plus the self-deactivate rejection. Uses the same `fakeResolver` introduced for the UpdateUserRole tests so authorizeCallerForUser's parity check is satisfiable without a DB. Each case asserts the expected SoftDeleteUser / LockAndCount call shape via gomock expectations (e.g., the count=1 case sets no SoftDelete expectation, so the mock would fail loudly if the production code accidentally proceeded past the guard).

client/src/protoFleet/features/settings/components/Team.tsx
mcharles-square: > this is ok for this PR, but we should align how we render modals across the app. everything in settings seems to always mount it and toggle `open` for visibility. In other parts of the app we conditionally render the modals. its about 50/50 from what i can tell

Acknowledged — leaving as-is for this PR per your note. The two modals here (AddTeamMemberModal, EditRoleModal, ResetPasswordModal, DeactivateUserDialog) all follow the always-mounted-with-`open` pattern shared by the rest of Settings. Picking one convention is worth doing but it's its own PR with its own choice (always-mount has slightly better animation continuity; conditional-mount has cheaper initial render and avoids stale `useEffect` triggers on hidden modals). Happy to take the cleanup ticket if you want to spin it off.

mcharles-square:
> [Codex Security Review #3](https://github.com/block/proto-fleet/pull/393#issuecomment-4624014023) — Role update can race with user deactivation [LOW]

Addressed in 7bcc1132 by extending `GetOrgScopeAssignmentForUser` to join the `"user"` table, filter on `u.deleted_at IS NULL`, and take `FOR UPDATE OF uor, u`. That covers both directions of the race:

- **DeactivateUser first:** locks `(uor, u)` inside its tx, soft-deletes the user, commits. A concurrent `UpdateUserRole` blocks on `(uor, u)`; once unblocked, its re-read of the WHERE clause sees `u.deleted_at IS NOT NULL` → no rows → `InvalidArgument`. No stale role assignment is created.
- **UpdateUserRole first:** locks `(uor, u)`, swaps, commits. `DeactivateUser` blocks on `(uor, u)`; once unblocked, its in-tx `GetOrgScopeAssignmentForUser` re-evaluates against the new live assignment row, the last-SA guard runs against the fresh state if applicable, then `SoftDeleteUser` proceeds.

The user-row lock is taken via `FOR UPDATE OF uor, u` (not the role row) so the lock surface stays minimal and doesn't contend with unrelated role mutations. SQL comment updated to describe both interleavings.

PR #396, opened 2026-06-04
feat(fleetnode): pair discovered miners (orchestration + operator RPCs)

ankitgoswami:
## Summary

Completes **pairing miners discovered via fleet nodes**, stacked on the foundation in #388 . With this, an operator can take a fleet-node-discovered miner from a `discovered_device` dead-end through to a fully paired, node-owned device.

## What is in here

- **Persist pairing results.** The control registry routes per-device pair results to the waiting operator session (`PublishPairResults`). `ResolvePairTargets` builds the batch from the node's not-yet-paired devices, and `PersistFleetNodePairResult` records each result in one transaction:
  - `PAIRED`: create the device, bind it to the node (`fleet_node_device`) **before** marking `device_pairing` PAIRED so the cloud-paired guard never sees a PAIRED-but-unbound row, store encrypted credentials for basic-auth drivers, seed an ACTIVE status.
  - `AUTH_NEEDED` / `AUTH_FAILED`: persist an AUTHENTICATION_NEEDED device for credential retry.
  - `ERROR`: persist nothing.
  `pairDeviceLocked` is factored out of `PairDevice` so both share the binding logic inside a caller-owned transaction.
- **Operator RPCs.** The gateway `ReportPairedDevices` handler binds the node's results to the in-flight command and routes them up. The server-streaming `PairDiscoveredDevicesOnFleetNode` dispatches a pair `AgentCommand` down the ControlStream and streams per-device outcomes as it persists them (mirroring `DiscoverOnFleetNode`); `ListFleetNodeDiscoveredDevices` enumerates candidates.  `fleetnode:manage`./Gated by `fleetnode:read` (listing) and `fleetnode:manage` (pairing); `PairDiscoveredDevicesOnFleetNode` additionally requires `miner:pair`, consistent with every other miner-onboarding path (e.g. ForemanImport), since it authenticates and onboards miners./ `fleetnode:manage`.

## Testing

- `go build ./...` clean; client `tsc` clean.
- Node tests pass with `-race`.
- Fleetnode domain tests against the live test DB cover `PersistFleetNodePairResult` (asymmetric PAIRED, basic-auth PAIRED with creds stored, AUTH_NEEDED then retry to PAIRED, foreign-node rejection, ERROR persists nothing) and `ResolvePairTargets`.
- Admin handler tests drive the full server-streaming pair path against a simulated node and assert the persisted `device_pairing = PAIRED` + `fleet_node_device` binding, plus no-stream / no-targets / permission cases; gateway tests cover `ReportPairedDevices` routing.

The fake-rig fixtures need no change: the node calls the identical plugin `PairDevice` contract the cloud uses (only the SecretBundle source differs). A full Docker E2E run needs a live `just dev` stack with an enrolled, connected fleet node; the demo script is that manual-verification path.


server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -0,0 +1,235 @@
...
+		ids = identifiers
+		if ids == nil {
+			ids = []string{}
+		}
+	}
+	// A credential-less pair-all can't satisfy AUTHENTICATION_NEEDED rows, so exclude
+	// them: otherwise a capped batch full of unsatisfiable rows starves the
+	// never-attempted devices beyond `limit` on re-issue. Explicit selection and
+	// pair-all WITH credentials still target them (the operator can retry with creds).
+	excludeAuthNeeded := pairAllUnpaired && credentials == nil
```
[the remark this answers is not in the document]
ankitgoswami: Fixed: the exclusion now keys off usable basic-auth credentials, mirroring the node's secretBundleFor — `usableCredentials := credentials != nil && credentials.Password != nil` (a username-only message no longer re-enables AUTHENTICATION_NEEDED rows). Added a regression assertion in TestResolvePairTargets_PairAllExcludesAuthNeededWithoutCredentials.

server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -0,0 +1,233 @@
...
+				return nil
+			}
+		}
+
+		applyReportedIdentity(dd, result)
+		if _, err := s.discoveredDeviceStore.Save(ctx, doi, dd); err != nil {
+			return fleeterror.LogInternal(component, "save discovered device", clientErrPair, err)
+		}
+		if existing == nil {
+			if err := s.deviceStore.InsertDevice(ctx, &dd.Device, orgID, identifier); err != nil {
```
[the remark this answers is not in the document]
ankitgoswami: Fixed: added migration 000076 widening device.device_identifier from VARCHAR(36) to VARCHAR(255) to match discovered_device (synthesized "serial:<...>" identifiers can exceed 36 chars). Increasing a VARCHAR length does not rewrite the table.

server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -0,0 +1,225 @@
...
```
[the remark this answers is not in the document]
ankitgoswami: Looked into this and don't think a change is needed: device.mac_address is VARCHAR(17) NOT NULL with no CHECK/UNIQUE, and Postgres NOT NULL rejects only NULL, not empty string. InsertDevice passes a Go string, so a MAC-less PAIRED report writes "" and the insert succeeds — the transaction does not roll back and the paired miner IS recorded (with an empty mac_address), not left unrecorded. discovered_device has no mac_address column, so no MAC is captured at discovery for fleet-node devices either; populating one is a separate data-quality improvement, out of scope here. Leaving the thread open in case there's a downstream MAC-keyed constraint you had in mind.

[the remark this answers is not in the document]
ankitgoswami: Added a minimal collision guard: InsertDevice/UpdateDeviceInfo now detect a unique-constraint violation (db.IsUniqueViolationError) and return a clean StatusFailed instead of an opaque Internal error — the per-device tx rolls back, the operator sees FAILED for that miner (so it can be reconciled), and the node isn't told to retry (a retry wouldn't clear the conflict). This mirrors how node-reported ERROR outcomes are handled. Full identity reconciliation (attach the auto:* pair to the existing serial/MAC device, like the cloud pairer's reconcileByMAC) is the larger correct fix and is left as a follow-up. Regression test: TestPersistFleetNodePairResult_SerialConflictFailsCleanly.

server/cmd/fleetnode/pair.go
```diff
@@ -224,24 +246,34 @@ func (r *RunCmd) handlePairCommand(ctx context.Context, client gatewayClient, st
 		r.sendAck(stream, commandID, pb.AckCode_ACK_CODE_PARTIAL, fmt.Sprintf("pair supervisor budget exceeded; %d of %d result(s) uploaded", len(results), len(targets)), logger)
```
rongxin-liu, line 258: This `PARTIAL` path can complete the command after `fanOutPairs` returns from the supervisor budget, even if `PairDevice` goroutines that ignore context are still running. A later pair command can then start while old mutating workers may still be changing miners.

Should we keep a pair-specific in-flight guard until workers exit, or otherwise avoid abandoning mutating `PairDevice` calls?

server/internal/handlers/fleetnode/gateway/handler.go
```diff
@@ -157,6 +157,73 @@ func (h *Handler) ReportDiscoveredDevices(ctx context.Context, req *connect.Requ
...
+		case errors.Is(admitErr, control.ErrEmptyReport):
+			return nil, fleeterror.NewInvalidArgumentError("pairing report carried no results")
+		default:
+			return nil, fleeterror.NewFailedPreconditionError("pairing report does not match an in-flight server-issued command")
+		}
+	}
+
+	persisted := make([]*pb.FleetNodePairResult, 0, len(kept))
+	for _, r := range kept {
+		status, err := h.pairing.PersistFleetNodePairResult(ctx, subject.FleetNodeID, meta.OrgID, r, meta.AssignedBy)
```
rongxin-liu, line 188: This persists after `AdmitAndScopePairResults` has already consumed the target from the registry. If `PersistFleetNodePairResult` fails transiently, the response reports the result as rejected, but retrying the same `command_id` cannot persist it because the target has already been deleted.

Do we need to make these persistence failures retryable, or only consume a pair target once persistence succeeds/idempotently recognizes an already-persisted result?

server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -0,0 +1,245 @@
...
+				persisted = StatusPaired
+				return nil
+			}
+		}
+
+		applyReportedIdentity(dd, result)
+		if _, err := s.discoveredDeviceStore.Save(ctx, doi, dd); err != nil {
+			return fleeterror.LogInternal(component, "save discovered device", clientErrPair, err)
+		}
+		if existing == nil {
```
rongxin-liu, line 129: The prior identity-reconciliation feedback still looks unresolved here. When an `auto:*` discovered device reports a serial/MAC that already belongs to an existing device, this path only looks up by the pre-pair `device_identifier` and then inserts.

Perhaps we first reconcile by the reported serial/MAC and bind/update that existing device instead of surfacing the unique conflict as `FAILED`?

server/internal/domain/stores/sqlstores/device.go
```diff
@@ -137,11 +137,8 @@ func (s *SQLDeviceStore) GetDeviceByDeviceIdentifier(ctx context.Context, identi
 
 func (s *SQLDeviceStore) UpdateDeviceInfo(ctx context.Context, device *pb.Device, orgID int64) error {
```
rongxin-liu, line 138: In this function, the sqlc error is still wrapped below with `%v`, so callers cannot recover the underlying Postgres unique violation with `errors.As`. That means an `AUTHENTICATION_NEEDED` retry that updates to a duplicate serial can miss the clean `StatusFailed` conflict path in `PersistFleetNodePairResult` and return an Internal error instead. Should this preserve the cause with `%w` or return the raw DB error, plus a regression test for the update-path duplicate serial case?

server/internal/domain/fleetnode/control/stream.go
```diff
@@ -62,23 +62,70 @@ func (r *Registry) PublishBatch(fleetNodeID int64, commandID string, batch *pair
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	cmd := r.inflightFor(fleetNodeID, commandID)
+	if cmd == nil || !cmd.reportBearing() || cmd.kind != ReportKindPair || cmd.pair == nil {
+		return nil, PairPersistMeta{}, errNoInFlightCommand
+	}
+	if len(results) == 0 {
+		return nil, PairPersistMeta{}, ErrEmptyReport
+	}
+	if cmd.reported+len(results) > cmd.maxReports {
```
rongxin-liu: This quota check charges the raw result count before duplicates or out-of-scope rows are filtered. For a two-target command, a report like `[a, a]` consumes quota for both even though only one target is kept; a later valid `b` report is then rejected as over quota.

Could quota be charged against unique in-scope targets that are actually consumed, with any raw-report abuse tracked separately?

server/internal/domain/fleetnode/pairing/models.go
```diff
@@ -2,6 +2,14 @@ package pairing
 
 import "time"
 
+// device_pairing.pairing_status values this package writes. Mirror the
```
rongxin-liu: This package now has a second hand-maintained copy of the persisted `device_pairing.pairing_status` strings, while `internal/domain/pairing` already owns the canonical constants.

To avoid drift across pairing paths, could this import/use the canonical constants instead of mirroring them locally?

rongxin-liu (approved):

server/cmd/fleetnode/pair.go
ankitgoswami: Good catch — fixed with a pair-specific in-flight gate. handlePairCommand now takes a TryLock on a RunCmd-level pairMu before fanning out and acks BUSY (retryable for the operator) if held; fanOutPairs returns a workersDone channel closed only when every started worker has exited, and the gate is released on that signal. So after a truncated PARTIAL, a re-issued pair command is refused BUSY until the abandoned ctx-ignoring workers actually finish — it can no longer race them. (The receive loop's exclusive lane already serialized concurrent pair commands; the gate covers the post-handler window the lane releases.) Test: TestHandlePairCommand_BusyWhileAbandonedWorkersStillRunning.

server/internal/handlers/fleetnode/gateway/handler.go
ankitgoswami: Fixed: persistence failures are now retryable within the command. The gateway collects identifiers whose PersistFleetNodePairResult returned an error and calls the new Registry.ReinstatePairTargets, which returns them to the command's target set (and re-credits quota), so a retried report for the same command_id is re-admitted and persisted. Deliberate conflicts (serial collision) return StatusFailed with no error and are NOT reinstated — a retry can't clear those. Tests: TestRegistry_ReinstatePairTargets_AllowsRetryAfterPersistFailure.

server/internal/domain/fleetnode/pairing/pair_discovered.go
ankitgoswami: Acknowledged this is still the reconcile-less behavior — that's a deliberate scope call rather than an oversight. Full reconciliation here needs the cloud pairer's machinery (any-device-by-serial/MAC lookup, cross-check guards, discovered_device identifier rewrite/attribution merge, and the identifier mismatch between the auto:* discovery row and the existing device row), which is a substantial addition to an already-large PR. What this PR guarantees instead: the conflict is detected cleanly (unique violation -> StatusFailed, no Internal error, per-device tx rolled back, surfaced to the operator for manual attach/retry) on both the insert and update paths, with regression tests. Reconciling auto:* pairs onto existing devices by reported serial/MAC is queued as the follow-up to this PR. Happy to prioritize it next if you think it should block.

server/internal/domain/stores/sqlstores/device.go
ankitgoswami: Fixed: UpdateDeviceInfo now wraps the sqlc error with %w (FleetError supports Unwrap), so db.IsUniqueViolationError reaches the pgconn.PgError and the update-path duplicate serial takes the same clean StatusFailed conflict path as the insert path. Regression test: TestPersistFleetNodePairResult_UpdateSerialConflictFailsCleanly (existing device updates to a serial owned by another device -> FAILED, serial unchanged, no Internal error).

server/internal/domain/fleetnode/control/stream.go
ankitgoswami: Fixed: quota is now charged per consumed in-scope target, not raw rows. AdmitAndScopePairResults scopes/dedupes first and adds only len(kept) to the command's reported count, so a [a, a] batch consumes one target and a later valid b is still admitted; duplicate/out-of-scope rows are dropped and logged (the existing anomalous-report Warn) rather than starving the command. Consumption itself caps total admissions at the dispatched target count. Tests: TestRegistry_AdmitAndScopePairResults_DuplicatesDoNotStarveLaterTargets.

server/internal/domain/fleetnode/pairing/models.go
ankitgoswami: Fixed: the local constants are now aliases of the canonical internal/domain/pairing constants (StatusPaired = domainpairing.StatusPaired, etc.), so the two pairing paths can't drift; call sites are unchanged. (Kept as aliases rather than direct imports to avoid threading a second package qualifier through this package, but they're compile-time equal.)

server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -0,0 +1,245 @@
...
+			ids = []string{}
+		}
+	}
+	// A pair-all without usable basic-auth credentials can't satisfy
+	// AUTHENTICATION_NEEDED rows, so excluding them stops unsatisfiable rows from
+	// starving never-attempted devices on re-issue. Usable = password set (matches
+	// the node's secretBundleFor); explicit selection still targets them.
+	usableCredentials := credentials != nil && credentials.Password != nil
+	excludeAuthNeeded := pairAllUnpaired && !usableCredentials
+	candidates, err := s.store.ListFleetNodeDiscoveredDevices(ctx, orgID, &fleetNodeID, ids, nil, limit, excludeAuthNeeded)
```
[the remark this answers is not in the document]
ankitgoswami: Fixed: the listing's node-bound and cloud-PAIRED exclusions in ListFleetNodeDiscoveredDevices now match by discovered_device_id OR (device_identifier, org_id), so a paired device whose original discovery row was soft-deleted and re-created by a node (same identifier, diverged linkage) is excluded before dispatch — the node is never sent a target persistence would reject after mutating the miner. Regression test: TestResolvePairTargets_ExcludesPairedDeviceWithDivergedDiscoveryLinkage (covers pair-all and explicit selection).

server/sqlc/queries/fleetnodepairing.sql
```diff
@@ -184,16 +204,41 @@ WHERE dd.org_id = $1
...
+  -- credentials, which can retry them).
+  AND (
+    NOT COALESCE(sqlc.narg('exclude_auth_needed')::bool, FALSE)
+    OR NOT EXISTS (
+      SELECT 1
+      FROM device adn
+      JOIN device_pairing adp ON adp.device_id = adn.id
+      WHERE adn.discovered_device_id = dd.id
+        AND adn.deleted_at IS NULL
+        AND adp.pairing_status = 'AUTHENTICATION_NEEDED'
```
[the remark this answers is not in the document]
ankitgoswami: Fixed: the AUTHENTICATION_NEEDED starvation exclusion now uses the same identifier-level match as the paired/bound guards — (adn.discovered_device_id = dd.id OR (adn.device_identifier = dd.device_identifier AND adn.org_id = dd.org_id)) — so a diverged-linkage auth-needed row no longer refills credential-less pair-all pages. Regression test: TestResolvePairTargets_AuthNeededExclusionSurvivesDivergedLinkage (excluded without creds, still retryable with creds).

PR #422, opened 2026-06-10
feat(buildings): site dropdown in BuildingDetailsModal + global Add building CTA

### LINKED ISSUE #371: Multi-site fleet page — BuildingDetailsModal site dropdown + Buildings-tab Add Building CTA
Follow-up to #368 (Fleet page scaffold).

## Context

The `/fleet/buildings` tab is currently missing an "Add building" CTA. The reason: `BuildingDetailsModal` today is always launched from a parent-site context (the create site dropdown is implicit). The 2026-06-02 redesign requires a global "Add building" entry point — see plan [J3](https://github.com/block/proto-fleet/blob/main/docs/plans/2026-05-05-multi-site-support-plan.md#j3-fleet-page-fleet--tabbed-list-home).

## Scope

1. **Modal change.** Add a **required Site dropdown field** to `BuildingDetailsModal`:
   - Populated from `ListSites`.
   - When opened from `/sites/:id`'s "Add building" CTA, the dropdown is pre-filled and read-only.
   - When opened from the new Buildings-tab CTA, the dropdown is editable with no default.
2. **Buildings tab CTA.** Add a right-aligned "Add building" button in the filter row of `/fleet/buildings`. Opens `BuildingDetailsModal` create mode with no pre-filled site.
3. **Validation.** Submit is disabled until a site is selected.

## Notes

- The empty-state message on `/fleet/buildings` today says "Add a building from a site detail page" — replace with the new CTA once this ships.
- `useBuildingModals.openDetailsCreate` already takes `siteId: bigint` — extend or add a sibling that accepts `siteId | undefined` so the modal can ask.

flesher:
## Summary

- Adds a required Site dropdown to `BuildingSettingsModal` in create mode (populated from `ListSites`); locked when opened from a site context, editable when opened from the new global Buildings-tab CTA.
- `/fleet/buildings` Add building CTA now opens the modal directly with no pre-filled site, replacing the prior `SiteSelectModal` step (deleted).
- Save stays disabled until both a name and a site are chosen.

Closes #371

## Test plan
- [ ] `BuildingSettingsModal` create flow: locked dropdown from `/sites/:id`, editable dropdown from `/fleet/buildings`, Save gated on site + name
- [ ] Verify edit mode is unchanged (no dropdown rendered)
- [ ] `npx vitest run src/protoFleet/features/buildings src/protoFleet/features/sites src/protoFleet/features/fleetManagement`


client/src/protoFleet/features/buildings/components/BuildingSettingsModal/BuildingSettingsModal.tsx
```diff
@@ -151,11 +172,34 @@ const BuildingSettingsModal = (props: BuildingSettingsModalProps) => {
...
+    }
     await props.onSave(values);
-  }, [buildValues, props]);
+  }, [buildValues, props, siteIdText]);
 
   const nameValid = name.trim().length > 0;
-  const primaryDisabled = !nameValid || saving;
+  // Create mode also gates Save on a site selection — buildings can't be
+  // created without a parent. Edit mode ignores the dropdown entirely.
+  const siteValid = !isCreate || siteIdText !== "";
```
rongxin-liu: Should we validate the selected site against the current `siteOptions`, not just `siteIdText !== ""`?

If the global create modal is open, the user selects Site A, and the sites list refreshes without Site A, the `Select` will render a blank Site label while `siteIdText` still contains A's id. That leaves Save enabled and submits a stale/deleted site id. For the editable dropdown, clearing/disabling the stale selection would make this state explicit; for locked site-scoped creates, we may want an explicit stale-site/error state.

client/src/protoFleet/features/buildings/hooks/useBuildingModals.test.ts
```diff
@@ -99,7 +109,7 @@ describe("useBuildingModals", () => {
     act(() => result.current.openDetailsCreate(7n, "North DC"));
 
     await act(async () => {
-      await result.current.detailsCreate(emptyBuildingFormValues());
+      await result.current.detailsCreate(emptyBuildingFormValues(), 7n);
```
rongxin-liu, line 112: Would this test cover the new global create path as well?

Looks like it still opens with `openDetailsCreate(7n, "North DC")`, so it does not prove the `/fleet/buildings` path (no initial site) threads the chosen site through to the mutation. A focused version could call `openDetailsCreate()` with no args, then `detailsCreate(values, 9n)`, and assert `createBuilding` receives `siteId: 9n` before checking refetch/close behavior.

rongxin-liu (approved):

client/src/protoFleet/features/buildings/components/BuildingSettingsModal/BuildingSettingsModal.tsx
flesher: Fixed in 02243467. Now validates `siteIdText` against `siteOptions` (computed from the live `sites` prop), not just non-empty. When the list refreshes and drops the chosen site, Save locks and the Select surfaces "Selected site is no longer available. Refresh and try again." Locked / site-scoped entry hits the same code path — if the pre-filled site disappears, the operator sees the error and can't submit. `handlePrimary` re-checks membership for belt-and-suspenders against a click racing the refresh.

client/src/protoFleet/features/buildings/hooks/useBuildingModals.test.ts
flesher: Added in 02243467. New focused test "global create path: openDetailsCreate() + detailsCreate(values, siteId) threads the chosen siteId to CreateBuilding" — opens with no args, calls `detailsCreate(emptyBuildingFormValues(), 9n)`, then asserts `buildingsClient.createBuilding` was invoked with `siteId: 9n` before checking refetch + state collapse. Proves the modal-collected siteId reaches the mutation when state.siteId is undefined.

PR #443, opened 2026-06-12
Improve curtailment workflow UX

negarn:
## Summary
- Allow curtailment CTAs to surface required-field validation instead of staying disabled for missing local inputs

https://github.com/user-attachments/assets/a08dcb2a-a885-465f-bf9f-c75fd82ec99a

- Restore saved response profile scope when switching profiles, including clearing stale miner selections for whole-fleet profiles

https://github.com/user-attachments/assets/edeea1aa-dd33-4eca-9418-e87655ae9ec0

- Keep readonly inputs from showing focus highlights and update leftover MQTT wording to MaestroOS

https://github.com/user-attachments/assets/8a8b3de3-a7d3-4039-b2c1-020c076fa6bf

- Update curtailment polling so that when restoring has completed, the curtailment history status changes to "completed" instead of remaining at "restoring" until manually refreshing the page

https://github.com/user-attachments/assets/a82e3321-ad9a-43db-91c3-44b28ff1fed6

## Test plan
- [ ] In New curtailment, click Run with required fields empty and confirm field errors appear on the form
- [ ] Select miners, switch to a whole-fleet response profile, and confirm the Apply to selection resets before running
- [ ] Verify MaestroOS source and automation modals surface required-field errors from their Save/Test actions

### DIFF
--- a/client/src/protoFleet/features/settings/components/Curtailment/CurtailmentSettingsPage.tsx
+++ b/client/src/protoFleet/features/settings/components/Curtailment/CurtailmentSettingsPage.tsx
@@ -810,6 +810,9 @@ function SourceModal({
     }
 
     if (!canTestConnection) {
+      if (showSavedPasswordPlaceholder && testConnectionValidationErrors.password) {
+        setPasswordPlaceholderActive(false);
+      }
       setValidationIntent("testConnection");
       return;
     }
@@ -824,7 +827,14 @@ function SourceModal({
     } finally {
       setShowConnectionCallout(true);
     }
-  }, [canTestConnection, isBusy, onTestConnection, values]);
+  }, [
+    canTestConnection,
+    isBusy,
+    onTestConnection,
+    showSavedPasswordPlaceholder,
+    testConnectionValidationErrors,
+    values,
+  ]);
 
   const handleDelete = useCallback(async () => {
     if (!onDelete || isBusy) {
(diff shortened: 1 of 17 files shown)


client/src/protoFleet/features/settings/components/Curtailment/CurtailmentSettingsPage.tsx
```diff
@@ -743,7 +774,12 @@ function SourceModal({
   }, [showSavedPasswordPlaceholder]);
 
   const handleSave = useCallback(async () => {
-    if (!canSave || isBusy) {
+    if (isBusy) {
+      return;
+    }
+
+    if (!canSave) {
```
rongxin-liu, line 781: In edit mode with a saved password, changing a credential field (e.g. changing username) makes `passwordRequired` true, so clicking Save lands here with `saveValidationErrors.password` set — but the input keeps rendering the `......` placeholder, so the field looks filled while the error says it's empty. The Test connection path already clears the placeholder for exactly this case (from the earlier review round);

Save needs the same treatment:

```ts
if (!canSave) {
  if (showSavedPasswordPlaceholder && saveValidationErrors.password) {
    setPasswordPlaceholderActive(false);
  }
  setValidationIntent("save");
  return;
}
```

(plus `saveValidationErrors` / `showSavedPasswordPlaceholder` in the `useCallback` deps).


Image

rongxin-liu (approved):
Looks good overall! Some minor issues worth fixing before merging.

PR #454, opened 2026-06-15
notifications (3/5): notification:read and notification:manage permissions

illegalprime:
## Summary

This PR makes the Notifications surface (channels, alert rules, maintenance windows, delivery history) permission-gated. It introduces a new `notification` authorization resource with two permissions — `notification:read` (view) and `notification:manage` (mutate) — wires every Notifications RPC to the permission it requires, and seeds the permission rows while granting them to existing ADMIN roles. No Notifications handlers are mounted yet (that lands in the next stack PR); this PR establishes the authorization *policy* those handlers will enforce.

It is PR 3 of a 5-PR stack: #452 (proto + generated) → #453 (history-listing DB) → **#454 (this PR, authz)** → #455 (service + handler mounts + wiring) → #456 (client UI).

## How it works

There are two independent paths: the **request-time enforcement policy** (in code) and the **permission data in the database** (rows + role grants).

**Request-time enforcement.** Every authenticated request flows through an auth interceptor that resolves the caller's `EffectivePermissions` (the union of permissions granted by the caller's roles) and stashes it on the request context. A handler gates itself by calling `RequirePermission(key, scope)`, which fails closed: missing context value → `Internal`; no session → `Unauthenticated`; caller lacks the key → `PermissionDenied`. The `ProcedurePermissions` map is the central policy table that says which catalog key each gated procedure needs. This PR adds the Notifications entries to that map: all list/read procedures → `notification:read`; every mutation → `notification:manage`. `TestChannel` is classified as a mutation (`:manage`) because it triggers a real outbound delivery, not just a read.

**Permission data reaching the database.** Two mechanisms populate the `permission` and `role_permission` tables:
1. **Boot reconciler** (`authz.Reconcile`, runs once per server start after migrations): upserts every catalog permission row (refreshing description text), and seeds built-in roles. Crucially, the reconciler seeds ADMIN/FIELD_TECH role grants **only when the role row is first created** — it never re-asserts the seed set onto an already-existing role, so operators' revocations are preserved. That means new catalog keys do NOT auto-propagate onto already-seeded ADMIN roles.
2. **Seed migration** (`000088_seed_notification_permissions`): does the one-time job the reconciler deliberately won't do — backfill the two new permissions onto **existing** ADMIN roles. It also inserts the permission rows itself (`ON CONFLICT` makes this redundant-but-safe alongside the reconciler), scopes the role grant to `builtin_key = 'ADMIN'` so custom roles are untouched, and uses `ON CONFLICT DO NOTHING` so it is replay-safe.

The migration is numbered `000088` so it sorts above the keyset-index migration `000087` from #453 (which merges first), keeping golang-migrate's numeric order aligned with merge order.

## Diagrams

Request authorization flow:

```mermaid
flowchart TD
    client["Client RPC call"] --> interceptor["Auth interceptor"]
    interceptor -->|"resolves caller roles -> EffectivePermissions"| ctx["Request context (stashed permissions)"]
    ctx --> handler["Notifications handler"]
    handler -->|"RequirePermission(key, scope)"| gate{"Caller has key?"}
    policy["ProcedurePermissions map (read -> notification:read, mutations -> notification:manage)"] -.->|"supplies required key"| gate
    gate -->|"yes"| allow["Handler executes (proxies to Grafana)"]
    gate -->|"no"| deny["PermissionDenied"]
    gate -->|"no session"| unauth["Unauthenticated"]
    gate -->|"missing ctx value"| internal["Internal (fail-closed)"]
```

Two ways the permissions reach the database:

```mermaid
flowchart LR
    catalog["catalog.go (in-code permission definitions)"]
    subgraph boot["Every server boot"]
        reconcile["authz.Reconcile"]
    end
    subgraph migrate["One-time migration 000088"]
        seed["seed_notification_permissions.up.sql"]
    end
    permTable[("permission rows")]
    rpTable[("role_permission grants")]

    catalog --> reconcile
    reconcile -->|"upsert all catalog rows"| permTable
    reconcile -->|"seed grants ONLY on first role creation"| rpTable
    seed -->|"insert permission rows (ON CONFLICT safe)"| permTable
    seed -->|"backfill grants onto EXISTING ADMIN roles"| rpTable
```

Lifecycle of a `notification:*` permission on an existing-DB upgrade:

```mermaid
stateDiagram-v2
    [*] --> CatalogDefined: "added to catalog.go"
    CatalogDefined --> RowUpserted: "reconciler upserts permission row on boot"
    RowUpserted --> GrantedToAdmin: "migration 000088 backfills existing ADMIN roles"
    CatalogDefined --> GrantedToAdmin: "migration also inserts rows if reconciler has not run"
    GrantedToAdmin --> [*]: "ADMIN callers now pass RequirePermission"
    note right of RowUpserted
      Reconciler will NOT grant new keys
      onto already-existing roles, hence
      the migration does the backfill.
    end note
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/domain/authz/catalog.go` | New `notification` resource constant; `notification:read` / `notification:manage` permission constants; two `CatalogEntry` rows. | Source of truth for the permission catalog. Confirm the resource is placed in declaration order (the order is contract-tested) and the descriptions match the migration text. |
| `server/internal/handlers/middleware/rpc_permissions.go` | Imports `notificationsv1connect`; adds 13 `ProcedurePermissions` entries (Channel/Rule/MaintenanceWindow/History services) mapping reads → `:read`, mutations → `:manage`. | The actual enforcement policy. Verify each mutation is `:manage`, each read is `:read`, and that `TestChannel` is `:manage` (outbound delivery). |
| `server/migrations/000088_seed_notification_permissions.up.sql` | **New.** Inserts the two permission rows (`ON CONFLICT` upsert) and backfills them onto existing `ADMIN` roles (`ON CONFLICT DO NOTHING`). | The one-time grant onto existing ADMIN roles. Verify scoping to `builtin_key = 'ADMIN'`, `deleted_at IS NULL`, and replay-safety. |
| `server/migrations/000088_seed_notification_permissions.down.sql` | **New.** Deletes the role grants then the permission rows. | Dev-only rollback. Confirm ordering (grants before rows, FK safety). |
| `server/internal/domain/authz/catalog_test.go` | Adds the two new permissions and the `notification` resource to the completeness / grouping / ordering contract fixtures. | Generated/derived expectations — quick check that fixtures match catalog.go, not deep review. |

No DB table for permissions-on-channels exists or is added: channels, rules, and maintenance windows live in Grafana, not in this database.

- **`TestChannel` gated on `:manage`, not `:read`** — chosen because it triggers an outbound delivery (a side-effecting action), over treating "test" as a read-only diagnostic.
- **Migration backfills existing ADMIN roles, even though the reconciler upserts rows** — chosen because the reconciler is intentionally additive only on first role creation and won't grant new keys to already-seeded roles; without the migration, existing orgs' admins would silently lack the new permissions.
- **Permission rows inserted in BOTH the reconciler and the migration** — chosen (with `ON CONFLICT`) over relying solely on the reconciler, so the migration is self-contained and the grant step has its FK target present regardless of boot ordering.
- **Numbered `000088` (above #453's `000087`)** — chosen to keep golang-migrate numeric order aligned with stack merge order, over a lower number that an already-upgraded DB would skip.
- **Grant scoped to `builtin_key = 'ADMIN'` with `DO NOTHING`** — chosen over a broad grant so custom/operator-tuned roles are untouched and re-runs are idempotent.
- **Notifications services intentionally absent from the `registeredServices` RPC-contract fixture** — those entries land in #455 alongside the real `main.go` handler mounts, because the contract test compares `registeredServices` against the live mux; the procedure→permission *policy* stays here.

## Testing & validation

- **Catalog contract tests** (`catalog_test.go`): assert the new permissions and resource appear in the completeness set, are grouped under `notification`, and respect declaration order.
- **RPC contract tests** (`rpc_permissions_test.go`): every key referenced by `ProcedurePermissions` must exist in the catalog, and every registered procedure must be classified in exactly one bucket.
- **Reconciler backfill tests** (`backfill_test.go`) exercise the additive-on-first-creation policy that motivates the migration.

Explicitly NOT covered here:
- **End-to-end handler enforcement** — no Notifications handlers are mounted in `main.go` yet (that is #455), so there is no integration test exercising a real gated Notifications request in this PR.
- **Migration apply/rollback against a live DB** — relies on the standard migration harness; the `down` migration is dev-only and assumes no custom-role hand-grants.
- **Per-scope authorization** — these permissions are org-level; no site/resource-scoped checks are introduced.

### DIFF
--- a/server/generated/sqlc/db.go
+++ b/server/generated/sqlc/db.go
@@ -753,6 +753,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.listNonTerminalCurtailmentEventsStmt, err = db.PrepareContext(ctx, listNonTerminalCurtailmentEvents); err != nil {
 		return nil, fmt.Errorf("error preparing query ListNonTerminalCurtailmentEvents: %w", err)
 	}
+	if q.listNotificationHistoryStmt, err = db.PrepareContext(ctx, listNotificationHistory); err != nil {
+		return nil, fmt.Errorf("error preparing query ListNotificationHistory: %w", err)
+	}
 	if q.listOrganizationsStmt, err = db.PrepareContext(ctx, listOrganizations); err != nil {
 		return nil, fmt.Errorf("error preparing query ListOrganizations: %w", err)
 	}
@@ -2411,6 +2414,11 @@ func (q *Queries) Close() error {
 			err = fmt.Errorf("error closing listNonTerminalCurtailmentEventsStmt: %w", cerr)
 		}
 	}
+	if q.listNotificationHistoryStmt != nil {
+		if cerr := q.listNotificationHistoryStmt.Close(); cerr != nil {
+			err = fmt.Errorf("error closing listNotificationHistoryStmt: %w", cerr)
+		}
+	}
 	if q.listOrganizationsStmt != nil {
 		if cerr := q.listOrganizationsStmt.Close(); cerr != nil {
 			err = fmt.Errorf("error closing listOrganizationsStmt: %w", cerr)
@@ -3423,6 +3431,7 @@ type Queries struct {
 	listMQTTSourceStatesByOrgStmt                         *sql.Stmt
 	listMinerStateSnapshotsStmt                           *sql.Stmt
 	listNonTerminalCurtailmentEventsStmt                  *sql.Stmt
+	listNotificationHistoryStmt                           *sql.Stmt
 	listOrganizationsStmt                                 *sql.Stmt
 	listPermissionsStmt                                   *sql.Stmt
 	listPoolsStmt                                         *sql.Stmt
@@ -3818,6 +3827,7 @@ func (q *Queries) WithTx(tx *sql.Tx) *Queries {
 		listMQTTSourceStatesByOrgStmt:                         q.listMQTTSourceStatesByOrgStmt,
 		listMinerStateSnapshotsStmt:                           q.listMinerStateSnapshotsStmt,
 		listNonTerminalCurtailmentEventsStmt:                  q.listNonTerminalCurtailmentEventsStmt,
+		listNotificationHistoryStmt:                           q.listNotificationHistoryStmt,
 		listOrganizationsStmt:                                 q.listOrganizationsStmt,
 		listPermissionsStmt:                                   q.listPermissionsStmt,
 		listPoolsStmt:                                         q.listPoolsStmt,
--- a/server/generated/sqlc/notification_history.sql.go
+++ b/server/generated/sqlc/notification_history.sql.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"time"
 )
 
 const insertNotificationHistory = `-- name: InsertNotificationHistory :exec
@@ -59,7 +60,6 @@ type InsertNotificationHistoryParams struct {
 	Annotations    json.RawMessage
 }
 
-// Persist one notification delivered by the Grafana alertmanager webhook receiver.
 func (q *Queries) InsertNotificationHistory(ctx context.Context, arg InsertNotificationHistoryParams) error {
 	_, err := q.exec(ctx, q.insertNotificationHistoryStmt, insertNotificationHistory,
 		arg.AlertName,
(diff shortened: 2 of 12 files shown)


illegalprime:
Codex finding **[MEDIUM] services in RBAC contract but not mounted** is a split artifact: this PR adds the permission map + `registeredServices` entries, and #455 mounts the four `notificationsv1connect` handlers in `main.go`. The contract and the mount land in adjacent PRs in the stack; nothing is missing once #455 merges. No change needed here.

illegalprime:
Re the re-review:

- **[MEDIUM] out-of-order migration won't backfill upgraded DBs** — flagging as a **merge-timing** note, not a present bug. `main`'s current max migration is `000086`; this stack's `000087/000088` are the next sequential versions, so they apply cleanly on any main-based DB. Also, the permission **rows** are upserted at every startup by `authz.Reconcile` (`main.go:204` → `upsertCatalog`), so they're never at risk. Migration 087's unique job is the one-time **grant of `notification:*` onto existing ADMIN roles** (the reconciler is additive by design and won't re-assert onto already-seeded roles). That backfill is only skipped if 087 ends up numbered below an already-applied version — i.e. only if other migrations land on `main` numbered ≥087 before this stack merges. **Merge-time action:** confirm 087/088 are still above `main`'s max at merge; renumber if not. Renumbering now would be premature.
- **[MEDIUM] services registered but not mounted** — unchanged disposition: split artifact, #455 mounts the four handlers in `main.go`. No change here.

illegalprime:
codex-connector triage (both fixed):

- **[P1] Renumber the seed after 000088** — fixed: seed is now **000088** (above the keyset index, now 000087 in #453), so numeric order matches merge order and golang-migrate won't skip it.
- **[P2] Drop unmounted services from the contract list** — fixed: the `registeredServices` entries (which `TestRPCContract_RegisteredServicesMatchMainMux` compares against the main.go mux) moved to #455 where the handlers are mounted. The `ProcedurePermissions` policy map stays here. `go test ./internal/handlers/middleware` is green on both PRs.

ankitgoswami (approved):
approving based on the PR description

PR #471, opened 2026-06-15
feat(proto): credentials auth and default-password state core

ankitgoswami:
## Summary

Switches the proto miner plugin from Ed25519 key-based authentication to username/password credentials with factory defaults of `admin`/`proto`. This foundation PR also introduces the server-side `DEFAULT_PASSWORD` pairing state and telemetry signal so the fleet can represent a rig that is paired and reporting metrics but still needs password rotation.

Stack: this PR is the base of the split stack. [#480](https://github.com/block/proto-fleet/pull/480) is stacked on this PR and hardens the server-side command, telemetry, SQL, cache, plugin, and fleet-node boundaries for `DEFAULT_PASSWORD`; [#482](https://github.com/block/proto-fleet/pull/482) is stacked on #480 and contains the hand-written Fleet UI remediation UX; [#477](https://github.com/block/proto-fleet/pull/477) is also stacked on #480 and handles phase-2 fleet-node enrollment cleanup. The diff here is relative to `main`; UI affordances and normal-action gating are intentionally out of scope and live in #482. Phase 3 remains a future cleanup PR for deleting dormant key-auth code and schema.

## How it works

The server continues dispatching miners by `DriverName`, but the proto driver now advertises `basic_auth` instead of asymmetric auth. Pairing builds a `UsernamePassword` secret, encrypts it through the existing credentials path, and reuses those credentials when reconnecting to the miner. If no credentials are supplied, the driver returns the proto defaults and the pairer tries those so factory-default rigs can be adopted without key provisioning.

The proto plugin client lazily logs in to the rig REST API, caches the access token, and retries once after a `401` to handle token expiry or out-of-band credential changes. The public status endpoint reports `default_password_active`; telemetry carries that bit back on `DeviceMetrics`, and the server reconciles it into `PAIRING_STATUS_DEFAULT_PASSWORD`. That state is paired-like for telemetry, snapshots, rollups, CSV, and enrichment, while password-change success clears it back to `PAIRED`.

## Diagrams

```mermaid
flowchart TD
  P["Pairer"] --> C{"proto driver capabilities"}
  C -->|"basic_auth"| S["UsernamePassword secret"]
  S --> E["encrypted miner_credentials"]
  S --> D["proto plugin client"]
  D --> L["POST /api/v1/auth/login"]
  L --> T["cached access token"]
  D --> R["GET /api/v1/system/status"]
  R --> M["DeviceMetrics.default_password_active"]
  M --> Q["device_pairing.DEFAULT_PASSWORD"]
```

```mermaid
stateDiagram-v2
  [*] --> PAIRED: pair with valid credentials
  PAIRED --> DEFAULT_PASSWORD: telemetry sees default_password_active
  DEFAULT_PASSWORD --> PAIRED: password update succeeds
  PAIRED --> AUTHENTICATION_NEEDED: credentials rejected
  AUTHENTICATION_NEEDED --> PAIRED: re-pair with valid credentials
```

## Areas of the code involved

| --- | --- | --- |
| `plugin/proto/internal/driver` | Driver API v2, `basic_auth` capability, default credentials, and credential bundle validation | Main auth-shape migration for proto rigs |
| `plugin/proto/pkg/proto` and `plugin/proto/internal/device` | Login/token cache, 401 retry, status/default-password reads, password-change client update | Runtime auth lifecycle against the rig REST API |
| `server/sdk/v1` and generated SDK outputs | Adds `DeviceMetrics.default_password_active` and capability plumbing | Wire/API contract between server and plugins |
| `proto/fleetmanagement/v1` and `server/migrations/000089_*` | Adds `PAIRING_STATUS_DEFAULT_PASSWORD` and DB enum migration | Persistent representation of default-password remediation state |
| `server/internal/domain/telemetry` | Reconciles the positive default-password signal from telemetry | State source of truth during polling |
| `server/internal/domain/command` | Persists proto password changes and clears default-password state on success | Keeps stored credentials aligned with on-device changes |
| `server/internal/domain/fleetmanagement`, `stores/sqlstores`, `buildings`, `sites`, `collection` | Treats `DEFAULT_PASSWORD` as paired-like for visibility/enrichment/CSV | Keeps metrics visible while remediation is pending |
| `server/fake-proto-rig` and `server/docker-compose.yaml` | Fake rig models credential login and default-password-gated mutation behavior | Test double follows the firmware contract |
| `client/src/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb.ts` | Regenerated from the enum change | Generated; hand-written UI is deferred to #482 |

| Decision | Trade-off |
| --- | --- |
| Replace proto auth in place under the existing `proto` driver name | Existing device rows keep working, but migration fallback has to handle key-era rows |
| Use default credentials as the adoption fallback | Factory rigs pair smoothly, while rigs with changed passwords surface as auth-needed |
| Model default password as a pairing state instead of only a snapshot flag | Requires a migration and paired-like handling, but gives command/server flows a durable remediation state |
| Detect default password from a positive status flag, not a telemetry failure | Preserves telemetry visibility, but requires firmware/plugin support for the status bit |
| Keep hand-written Fleet UI out of this PR | Smaller foundation review, but default-password UX is not fully visible until #482 |

## Testing & validation

- `cd client && npm run lint`
- `cd plugin/proto && ../../bin/go test ./...` partially passed: `internal/device`, `internal/driver`, `pkg/proto`, and `tests/unit` passed; `plugin/proto/tests` failed because Docker closed the image-build connection locally.
- `cd server && ../bin/go test ./internal/domain/plugins ./internal/domain/telemetry ./internal/domain/command` partially passed: `plugins` and `telemetry` passed; the broader `command` package hit local Postgres authentication failures for DB-backed tests.
- `git diff --check`

Not fully covered locally: Docker/testcontainers and DB-backed integration tests need a working local Docker/Postgres setup and should be covered by CI or a local `just dev` smoke before merge.


client/src/protoFleet/features/fleetManagement/utils/pairingRemediation.ts
```diff
@@ -0,0 +1,15 @@
+import { PairingStatus } from "@/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb";
+
+// Pairing states that block telemetry until the operator acts. Both surface as
+// "needs attention" and suppress live metrics, but their remediation differs:
+// AUTHENTICATION_NEEDED needs credentials, DEFAULT_PASSWORD needs a password change.
```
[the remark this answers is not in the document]
ankitgoswami: Good catch — corrected in c66424d8. The comment now states that only AUTHENTICATION_NEEDED suppresses live metrics, while DEFAULT_PASSWORD keeps telemetry flowing and only needs a password change.

client/src/protoFleet/features/fleetManagement/components/MinerList/MinerName.tsx
```diff
@@ -37,10 +40,16 @@ const MinerName = ({
   const name = miner.name || deviceIdentifier;
   const deviceStatus = miner.deviceStatus;
 
-  const needsAuthentication = miner.pairingStatus === PairingStatus.AUTHENTICATION_NEEDED;
+  const needsAuthentication = needsAuthFn(miner.pairingStatus);
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in fc5c6028. MinerName now passes needsCredentialRemediation (true for AUTHENTICATION_NEEDED and DEFAULT_PASSWORD) to SingleMinerActionsMenu, so default-password rigs are restricted to unpair/view. Telemetry display and the alert-to-password-change remediation stay keyed off the AUTHENTICATION_NEEDED-only check, so the rig still shows metrics and the operator is steered to the password change.

client/src/protoFleet/features/fleetManagement/utils/fleetVisiblePairingFilter.ts
```diff
@@ -8,6 +8,7 @@ import {
 export const FLEET_VISIBLE_PAIRING_STATUSES: PairingStatus[] = [
   PairingStatus.PAIRED,
   PairingStatus.AUTHENTICATION_NEEDED,
+  PairingStatus.DEFAULT_PASSWORD,
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in fc5c6028. MinerList disabledMinerIdSet now uses needsCredentialRemediation so DEFAULT_PASSWORD rows are non-selectable, and MinerActionsMenu's bulk guard was broadened to disable mutating actions if a default-password rig is included via select-all. Added a regression test for the bulk-guard case.

client/src/protoFleet/features/fleetManagement/components/MinerList/MinerName.tsx
```diff
@@ -65,7 +79,7 @@ const MinerName = ({
           deviceStatus={deviceStatus}
           minerName={name}
           workerName={miner.workerName}
-          needsAuthentication={needsAuthentication}
+          needsAuthentication={actionsRestricted}
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in b7454b6c. Added an allowSecurityAction prop to SingleMinerActionsMenu so the Manage security (password-change) action survives the restriction for DEFAULT_PASSWORD rigs — MinerName passes needsPasswordChange(status) — while all other mutating actions stay hidden. AUTHENTICATION_NEEDED keeps hiding it (no valid creds to act with). Added regression tests for both cases.

plugin/proto/internal/driver/driver.go
```diff
@@ -324,15 +333,28 @@ func (d *Driver) PairDevice(ctx context.Context, deviceInfo sdk.DeviceInfo, acce
...
+		return sdk.DeviceInfo{}, fmt.Errorf("expected UsernamePassword in secret bundle for pairing, got %T", access.Kind)
+	}
+
+	if err := client.SetCredentials(credentials); err != nil {
+		return sdk.DeviceInfo{}, fmt.Errorf("failed to set credentials: %w", err)
 	}
 
-	if err := client.Pair(ctx, publicKey); err != nil {
+	if err := client.Authenticate(ctx); err != nil {
 		return sdk.DeviceInfo{}, fmt.Errorf("pairing failed: %w", err)
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in d27cca2a. Device.Unpair now tolerates the default-password 403 from DELETE /pairing/auth-key (via the existing isDefaultPasswordError classifier) and completes the local unpair — a credentials-paired rig has no key to clear. Flipped the regression test (TestNew_DefaultPasswordActive_UnpairToleratesDefaultPassword) to assert success while still attempting the DELETE.

client/src/protoFleet/components/StatusModal/StatusModal.tsx
```diff
@@ -166,17 +192,11 @@ const ProtoFleetStatusModal = ({
...
-    const isMinersleeping =
-      (miner?.deviceStatus === DeviceStatus.INACTIVE || miner?.deviceStatus === DeviceStatus.MAINTENANCE) &&
-      !needsAuthentication;
-
     // Build buttons
     const buttons = [];
 
     // Add wake miner button if miner is sleeping
-    if (isMinersleeping) {
+    if (isSleeping) {
```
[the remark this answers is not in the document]
ankitgoswami: Fixed in d27cca2a. isSleeping now also excludes needsPwChange, so the Wake miner button (which runs the firmware-gated StartMining) is no longer offered for default-password rigs; they surface the password-change remediation instead.

ankitgoswami:
🤖 Re: Codex security review: This is addressed in the stacked child PR #480 rather than in the #471 diff: #480 adds paired-like miner resolution for DEFAULT_PASSWORD devices via `include_default_password`, routes UpdateMinerPassword through `GetMinerForCredentialRemediation`, routes telemetry through `GetMinerForTelemetry`, and keeps non-remediation commands explicitly guarded from DEFAULT_PASSWORD devices.

I am leaving #471 unchanged here so the stack stays as currently split: #471 carries the core DEFAULT_PASSWORD/auth state, and #480 carries the remediation/telemetry resolver hardening.

server/internal/domain/command/execution_service.go
```diff
@@ -578,16 +578,24 @@ func (es *ExecutionService) executeCommandOnDevice(ctx context.Context, commandT
...
-		// Proto devices (asymmetric auth) store no password in DB, so always evict.
-		if minerInfo.GetDriverName() != models.DriverNameProto {
-			if dbErr := es.updateMinerPasswordInDB(ctx, message.DeviceID, p.NewPassword); dbErr != nil {
-				slog.Error("device password updated but database sync failed",
-					"device_id", message.DeviceID, "error", dbErr)
-				break
-			}
+		// Surface a persistence failure as a command error: the on-device password
+		// already changed, so a silent success would leave Fleet with stale credentials.
+		if dbErr := es.persistMinerPassword(ctx, message.DeviceID, minerInfo.GetDriverName(), p.NewPassword); dbErr != nil {
```
rongxin-liu, line 583: If `UpdateMinerPassword` succeeds on the rig but `persistMinerPassword` fails, this returns an error before invalidating the cached miner. The cached plugin has already switched to the new in-memory password, so subsequent telemetry can keep succeeding and clear `DEFAULT_PASSWORD` even though durable credentials are stale. It seems worth invalidating on the persistence-failure path and only clearing remediation after the credential write is durable.

server/internal/domain/command/service.go
```diff
@@ -635,6 +638,24 @@ func (s *Service) resolveSelectorIdentifiers(ctx context.Context, selector *pb.D
...
 
+func pairingStatusFiltersForSelector(filter *pb.DeviceFilter, commandType commandtype.Type) []sql.NullString {
+	if filter != nil && len(filter.PairingStatus) > 0 {
+		return []sql.NullString{{
+			String: string(sqlstores.ProtoPairingStatusToSQL(filter.PairingStatus[0])),
+			Valid:  true,
+		}}
+	}
+
+	if commandType == commandtype.UpdateMinerPassword {
```
rongxin-liu, line 649: This now selects `DEFAULT_PASSWORD` devices for password remediation, but the miner resolver still only loads `PAIRED` rows from `server/sqlc/queries/miner_service.sql`. After the cached miner handle expires, `UpdateMinerPassword` and telemetry cannot instantiate the device this selector just admitted. Perhaps make miner resolution command-aware (or paired-like for remediation/telemetry) and regenerate sqlc.

rongxin-liu, line 649: Explicit device selectors bypass the new default-password command policy: they are copied through without checking pairing status, and only the UI hides non-security actions. Because telemetry does not evict cached miner handles when a device enters `DEFAULT_PASSWORD`, a direct API caller can still enqueue non-remediation commands against a cached handle until TTL/plugin rejection. Maybe enforce the `DEFAULT_PASSWORD` lockout in server preflight for every mutating command except password remediation.

server/internal/domain/telemetry/service.go
```diff
@@ -1139,6 +1145,35 @@ func (s *TelemetryService) persistFirmwareVersionIfChanged(ctx context.Context,
...
+	prevActive, _ := prev.(bool)
+	if seen && prevActive == active {
+		return
+	}
+
+	status := pairing.StatusPaired
+	if active {
+		status = pairing.StatusDefaultPassword
+	}
+	if err := s.deviceStore.UpdateDevicePairingStatusByIdentifier(ctx, string(deviceID), status); err != nil {
```
rongxin-liu, line 1169: `default_password_active=false` is not proof that Fleet's stored credentials are valid, but this unconditionally writes `PAIRED`. A stale successful telemetry sample can overwrite `AUTHENTICATION_NEEDED` or another non-paired transition. Worth making this a guarded transition, e.g. only `PAIRED <-> DEFAULT_PASSWORD` via compare-and-set semantics.

server/internal/domain/command/service.go
rongxin-liu, line 649: This selector path includes `DEFAULT_PASSWORD` for `UpdateMinerPassword`, but `CheckCommandCapabilities` still reads devices through queries that require `dp.pairing_status = 'PAIRED'`. That can make remediation report zero supported targets for the devices that need it. Perhaps make capability lookup command-aware so password updates include `DEFAULT_PASSWORD`, while normal commands continue to exclude it, then regenerate sqlc.

plugin/proto/internal/device/device.go
```diff
@@ -283,6 +283,14 @@ func (d *Device) Status(ctx context.Context) (sdk.DeviceMetrics, error) {
 
 	d.refreshFirmwareVersion(ctx, &metrics)
 
+	// Leave unset (nil) on a read failure so the server treats it as undetermined
+	// and doesn't demote a still-default-password device.
+	if defaultPasswordActive, err := d.client.IsDefaultPasswordActive(ctx); err != nil {
```
rongxin-liu, line 288: This adds a separate `/api/v1/system/status` request to every uncached Proto `Status` collection. The telemetry scheduler runs this continuously for paired/default-password rigs, so this is one extra REST call per device per poll at fleet scale. Worth throttling this probe like firmware refresh, piggybacking it on an already-fetched status payload, or limiting it to remediation-relevant states.

rongxin-liu (approved):

server/internal/domain/command/execution_service.go
ankitgoswami: good catch. I'll fix in #480, not #471. right now #480 invalidates after success and auth errors, but not after DB persistence failure. It already only clears remediation after a durable write, only missing piece is invalidating before returning the persistence error.

server/internal/domain/command/service.go
ankitgoswami: this is handled in #480

ankitgoswami: handled in #480 and #482

server/internal/domain/telemetry/service.go
ankitgoswami: handled in #480

server/internal/domain/command/service.go
ankitgoswami: good catch - i think this fix also belongs in #480,

plugin/proto/internal/device/device.go
ankitgoswami: fixed in #480

PR #474, opened 2026-06-16
Improve mobile fleet layout spacing

negarn:
## Summary
- Keep mobile header pills aligned by moving the first visible pill into the top bar and stacking overflow pills with consistent spacing
- Tighten mobile response profile card spacing and preserve curtailment table headers/toggles on small screens

## Test plan
- [ ] Open ProtoFleet on a phone-sized viewport and confirm header pills align with even spacing
- [ ] Confirm curtailment response profile cards and Sources/Automations tables fit without clipping
- [ ] Confirm desktop header and curtailment settings layouts are unchanged

image

### DIFF
--- a/client/src/protoFleet/components/AppLayout/AppLayout.test.tsx
+++ b/client/src/protoFleet/components/AppLayout/AppLayout.test.tsx
@@ -6,6 +6,7 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 import AppLayout from "./AppLayout";
 import type { CurtailmentPillEvent } from "@/protoFleet/components/PageHeader/CurtailmentPill";
 import type { UseSchedulePillDataResult } from "@/protoFleet/components/PageHeader/useSchedulePillData";
+import { useHasPermission } from "@/protoFleet/store";
 
 const mockUseWindowDimensions = vi.fn();
 const mockUseReactiveLocalStorage = vi.fn();
@@ -42,6 +43,10 @@ vi.mock("@/shared/hooks/useReactiveLocalStorage", () => ({
   useReactiveLocalStorage: () => mockUseReactiveLocalStorage(),
 }));
 
+vi.mock("@/protoFleet/store", () => ({
+  useHasPermission: vi.fn(),
+}));
+
 const createSchedulePillData = (overrides: Partial<UseSchedulePillDataResult> = {}): UseSchedulePillDataResult => ({
   hasVisibleSchedules: false,
   pillSchedule: null,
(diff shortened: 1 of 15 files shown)


client/src/protoFleet/components/PageHeader/PageHeader.tsx
```diff
@@ -26,41 +33,74 @@ interface PageHeaderProps {
...
+            ) : null;
+          case "setup":
+            return dismissedSetup ? (
+              <Button
+                key={widget}
+                variant={variants.secondary}
+                size={sizes.compact}
+                text="Continue setup"
+                onClick={onContinueSetup}
+              />
```
rongxin-liu: When setup is the only visible phone widget, this plain `Button` moves into the constrained inline header slot (`min(15rem,45vw)`) and the wrapper clips overflow. Unlike the schedule/curtailment pills, this path does not give the button/label `max-w-full`, `min-w-0`, or `truncate`, so narrow phones or increased text scaling can partially hide the only visible `Continue setup` action. 

Maybe either make this inline setup button shrink/truncate like `PageHeaderPopoverPill`, or keep non-truncating action buttons in the overflow row instead.

client/src/protoFleet/components/AppLayout/AppLayout.test.tsx
```diff
@@ -62,14 +67,36 @@ const activeCurtailmentEvent: CurtailmentPillEvent = {
...
     mockUseCurtailmentPillData.mockReturnValue({ activeEvent: null });
     mockUseSchedulePillData.mockReturnValue(createSchedulePillData());
+    vi.mocked(useHasPermission).mockReturnValue(true);
+  });
+
+  it("keeps the base phone content offset when the only schedule widget fits inline", () => {
+    mockUseSchedulePillData.mockReturnValue(
+      createSchedulePillData({
+        hasVisibleSchedules: true,
+      }),
```
rongxin-liu, line 102: These offset tests set `hasVisibleSchedules: true` while leaving `pillSchedule` as `null`. The real `useSchedulePillData` hook derives `hasVisibleSchedules` from `pillSchedule !== null`, and `PageHeader` only renders the schedule widget when `pillSchedule` exists. That makes these tests exercise a state the app cannot produce, so they can pass even if AppLayout's offset count and the actual rendered widgets drift apart.

Worth providing a non-null `pillSchedule` in the visible-schedule fixture, or derive the flag from the fixture object.

rongxin-liu (approved):

PR #477, opened 2026-06-16
feat(fleetnode): enroll nodes without miner signing keys

ankitgoswami:
Reviewable diff: +56/-23 across 7 files (excludes generated, test, and story files).

## Summary

This PR lets newly enrolled fleet nodes operate in a credentials-only world by no longer requiring each node to generate or register a miner-signing key. It also carries the fleet-node-specific default-password pair-result propagation moved out of #480: when a node pairs a Proto rig and the plugin reports factory credentials are still active, the cloud persists and streams `DEFAULT_PASSWORD`.

**Stack:** This PR is stacked on [#480](https://github.com/block/proto-fleet/pull/480), which hardens server/plugin default-password remediation behavior. The diff here is relative to #480. [#482](https://github.com/block/proto-fleet/pull/482) is a separate UI branch from #480 and is not required for fleet-node enrollment. Removing asymmetric-auth fields entirely is intentionally out of scope and remains a later phase.

## How it works

Fleet-node bootstrap now generates and stores the node identity key only. During registration, the node sends no miner-signing public key, and the server accepts an omitted value by normalizing it to an empty byte slice for persistence and downstream compatibility.

Pairing still supports credential-based drivers through the existing secret bundle path. If a driver still requires asymmetric auth and the node does not have a miner-signing key, the node pair command returns an explicit pair-result error instead of attempting to derive or send a missing key. If a node still has a miner-signing private key in local config, the legacy asymmetric pairing path remains available.

Fleet-node pair results now include an optional `default_password_active` flag. The node copies this from the plugin `DeviceInfo`, the gateway persists `DEFAULT_PASSWORD` for paired results with the flag set, and live operator streams forward the persisted default-password state.

```mermaid
sequenceDiagram
  participant Node as "Fleet node"
  participant Server as "Fleet server"
  participant DB as "Fleet DB"
  Node->>Node: "Pair with credentials"
  Node->>Server: "Report PAIRED + default_password_active"
  Server->>DB: "Persist DEFAULT_PASSWORD + credentials"
  Server-->>Node: "Report accepted"
```

## Areas of the code involved

| --- | --- | --- |
| `proto/fleetnodegateway/v1/fleetnodegateway.proto` | Makes `miner_signing_pubkey` optional for registration and adds `default_password_active` to pair results | Contract change for credentials-only nodes and default-password propagation |
| `server/cmd/fleetnode` | Stops generating/storing miner-signing keys for new enrollments; reports plugin default-password state from pair results | Runtime node behavior reviewers should scrutinize |
| `server/internal/handlers/fleetnode/gateway` | Accepts omitted miner keys and persists/forwards default-password pair state | Server compatibility and authoritative persistence boundary |
| `server/internal/handlers/fleetnode/admin` | Streams default-password pair results as `DEFAULT_PASSWORD` to operators | Keeps live pairing display aligned with DB state |
| `server/internal/domain/fleetnode` | Normalizes nil miner-signing keys and stores default-password pair outcomes | Prevents nil/empty representation bugs and preserves remediation state |
| Generated fleetnodegateway outputs | Regenerated from the proto contract change | Generated files can mostly be skimmed |

| Decision | Trade-off |
| --- | --- |
| Keep miner-signing fields optional instead of deleting them | Preserves compatibility for legacy asymmetric-auth nodes until the removal phase |
| New bootstrap path omits miner-signing keys | Avoids carrying unused secrets for credentials-auth nodes |
| Report default-password state explicitly from the node | Avoids server-side credential guessing for fleet-node paired rigs |
| Explicitly fail asymmetric pairing without a node key | Makes unsupported legacy-driver behavior visible instead of silently producing broken secrets |

## Testing & validation

- `git diff --check`
- `./bin/go test ./server/cmd/fleetnode -run 'TestPluginPairer_DefaultCredentialsReportsUsedCredentials'`
- `./bin/go test ./server/internal/handlers/fleetnode/admin -run 'TestPairResultStatus_DefaultPasswordActive'`
- `./bin/go test ./server/internal/domain/fleetnode/pairing ./server/internal/handlers/fleetnode/gateway -run '^$'`

DB-backed fleet-node persistence/gateway tests were attempted locally but the local Postgres harness failed before test execution with `failed SASL auth` for user `fleet`.


server/internal/domain/fleetnode/pairing/pair_discovered.go
```diff
@@ -83,6 +83,9 @@ func (s *Service) PersistFleetNodePairResult(ctx context.Context, fleetNodeID, o
 	persisted := StatusAuthenticationNeeded
 	if outcome == gatewaypb.PairOutcome_PAIR_OUTCOME_PAIRED {
 		persisted = StatusPaired
+		if result.GetDefaultPasswordActive() {
+			persisted = StatusDefaultPassword
+		}
```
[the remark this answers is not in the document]
ankitgoswami: 🤖
> `nil` = undetermined; consumers must not change remediation state

Addressed in 0998496: paired results now only set `DEFAULT_PASSWORD` from an explicit true, preserve an existing `DEFAULT_PASSWORD` state when the field is absent, and clear it only when the optional is explicitly false. Added regressions for absent and explicit-false reports.

Validation: `git diff --check` and `./bin/go test ./server/internal/domain/fleetnode/pairing -run '^$'` pass. The focused DB-backed tests reached the local Postgres harness but could not run assertions because the local `fleet` database credentials reject auth.

ankitgoswami:
🤖
> Optional default-password state is treated as explicit false

Addressed by 0998496. The persistence path now keeps `default_password_active` as tri-state: explicit true sets `DEFAULT_PASSWORD`, explicit false clears it to `PAIRED`, and an absent value preserves an existing `DEFAULT_PASSWORD` remediation state. Added regression coverage for the absent and explicit-false cases.

Validation: `git diff --check` and `./bin/go test ./server/internal/domain/fleetnode/pairing -run '^$'` pass. The focused DB-backed regression run could not execute assertions locally because the local Postgres `fleet` credentials reject auth.

server/cmd/fleetnode/pair.go
```diff
@@ -145,14 +159,22 @@ func (p *pluginPairer) Pair(ctx context.Context, target *pairingpb.FleetNodePair
 // key for asymmetric-auth drivers, supplied username/password otherwise. ok is
 // false when no credentials apply (the caller falls back to plugin defaults or
 // reports AUTH_NEEDED).
-func secretBundleFor(caps sdk.Capabilities, nodePubKey string, creds *pairingpb.Credentials) (sdk.SecretBundle, bool) {
+func secretBundleFor(caps sdk.Capabilities, nodePubKey string, creds *pairingpb.Credentials) (pairSecretBundle, bool, error) {
 	if caps[sdk.CapabilityAsymmetricAuth] {
-		return sdk.SecretBundle{Version: "v1", Kind: sdk.APIKey{Key: nodePubKey}}, true
+		if nodePubKey == "" {
+			return pairSecretBundle{}, false, errMissingMinerSigningKey
```
rongxin-liu, line 165: Should we handle this missing-key asymmetric case as a non-retryable target rather than a generic `ERROR`?

Since `ERROR` persists nothing, pair-all can probably keep selecting the same legacy asymmetric-only devices ahead of later valid targets. Perhaps we could either record a durable terminal/suppressed state for this case, or filter unsupported asymmetric targets before the pair-all page is selected.

server/internal/handlers/fleetnode/admin/handler_pair.go
```diff
@@ -126,11 +119,25 @@ func (h *Handler) PairDiscoveredDevicesOnFleetNode(ctx context.Context, req *con
...
+// pairResultStatus maps a node pair result to the operator-facing enum, matching
 // what PersistFleetNodePairResult records.
-func pairOutcomeStatus(outcome gatewaypb.PairOutcome) fleetmanagementv1.PairingStatus {
-	switch outcome {
+func pairResultStatus(result *gatewaypb.FleetNodePairResult) fleetmanagementv1.PairingStatus {
+	switch result.GetOutcome() {
 	case gatewaypb.PairOutcome_PAIR_OUTCOME_PAIRED:
+		if result.GetDefaultPasswordActive() {
+			return fleetmanagementv1.PairingStatus_PAIRING_STATUS_DEFAULT_PASSWORD
+		}
```
rongxin-liu, line 140: Perhaps worth updating the admin stream contract docs/generated output alongside this new status. This now lets a pair result surface `PAIRING_STATUS_DEFAULT_PASSWORD`, while the admin proto still says results are always `PAIRED`, `AUTHENTICATION_NEEDED`, or `FAILED`, so API consumers may miss the widened status set.

rongxin-liu (approved):

server/cmd/fleetnode/pair.go
ankitgoswami: this bit is gonna be very short lived - gonna rip it out in the next PR, just needed something here for sequencing

PR #502, opened 2026-06-18
feat: add Breadcrumb shared component with sibling switcher

jmarrxyz:
## Summary

- Adds a shared `Breadcrumb` component (`client/src/shared/components/Breadcrumb/`)
- Ancestor segments render as links; the last segment supports a sibling switcher dropdown for navigating between peers (e.g. switching between racks in the same building)
- Includes Storybook stories covering site, building, rack, and no-switcher variants

## Test plan

- [ ] `npm run storybook` — verify Shared/Breadcrumb stories render (SiteLevel, BuildingLevel, RackLevel, NoSiblings)
- [ ] `npm run typecheck` passes
- [ ] `npm run lint` passes

### DIFF
--- a/client/src/shared/components/Breadcrumb/Breadcrumb.tsx
+++ b/client/src/shared/components/Breadcrumb/Breadcrumb.tsx
@@ -1,0 +1,135 @@
+import { Fragment, useCallback, useRef, useState } from "react";
+import { Link, useNavigate } from "react-router-dom";
+import clsx from "clsx";
+
+import { ChevronDown } from "@/shared/assets/icons";
+import { iconSizes } from "@/shared/assets/icons/constants";
+import { useEscapeDismiss } from "@/shared/hooks/useEscapeDismiss";
+
+export interface BreadcrumbSibling {
+  label: string;
+  to: string;
+  isActive: boolean;
+}
+
+export interface BreadcrumbSegment {
+  label: string;
+  to?: string;
+  siblings?: BreadcrumbSibling[];
+}
+
+export interface BreadcrumbProps {
+  segments: BreadcrumbSegment[];
+  testId?: string;
+}
+
+const Breadcrumb = ({ segments, testId = "breadcrumb" }: BreadcrumbProps) => {
+  const [menuOpen, setMenuOpen] = useState(false);
+  const triggerRef = useRef<HTMLButtonElement>(null);
+  const navigate = useNavigate();
+
+  useEscapeDismiss(menuOpen ? () => setMenuOpen(false) : undefined);
+
+  const handleSelect = useCallback(
+    (to: string) => {
+      setMenuOpen(false);
+      navigate(to);
+    },
+    [navigate],
+  );
+
+  return (
+    <nav className="flex items-center gap-2 text-300" data-testid={testId} aria-label="Breadcrumb">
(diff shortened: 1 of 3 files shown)


client/src/shared/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export interface BreadcrumbProps {
+  segments: BreadcrumbSegment[];
+  testId?: string;
+}
+
+const Breadcrumb = ({ segments, testId = "breadcrumb" }: BreadcrumbProps) => {
+  const [menuOpen, setMenuOpen] = useState(false);
+  const triggerRef = useRef<HTMLButtonElement>(null);
```
rongxin-liu, line 37: `triggerRef` is attached to the switcher button below but never read. If it was it meant for restoring focus to the trigger when the menu closes then perhaps we could wire that up (it'd also help the a11y note below); otherwise we could probably just drop it.

client/src/shared/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,135 @@
...
+      navigate(to);
+    },
+    [navigate],
+  );
+
+  return (
+    <nav className="flex items-center gap-2 text-300" data-testid={testId} aria-label="Breadcrumb">
+      {segments.map((seg, i) => {
+        const isLast = i === segments.length - 1;
+        const hasSiblings = isLast && seg.siblings && seg.siblings.length > 0;
```
rongxin-liu, line 95: Maybe worth constraining the type so only the last segment carries siblings, or at least noting the behavior on `BreadcrumbSegment`?

client/src/shared/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,135 @@
...
+  onSelect: (to: string) => void;
+  onDismiss: () => void;
+  testId: string;
+}
+
+const SiblingMenu = ({ siblings, onSelect, onDismiss, testId }: SiblingMenuProps) => (
+  <>
+    <div className="fixed inset-0 z-20" role="presentation" onClick={onDismiss} />
+    <div
+      role="menu"
```
rongxin-liu: With `role="menu"` / `menuitem` + `aria-haspopup="menu"`, keyboard and screen-reader users will expect arrow-key navigation and focus moving into the menu (then back to the trigger on close). Today only Escape + mouse work. Perhaps we could add roving focus + focus management here — or, if that's more than we want right now, downgrade the roles to something simpler that matches the current behavior.

client/src/shared/components/Breadcrumb/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Breadcrumb } from "./Breadcrumb";
```
rongxin-liu: Other shared components default-export from their barrel (e.g. `import Select from "@/shared/components/Select"`), but this one only exports a named `Breadcrumb`, so it'd be imported differently from the rest. Maybe worth aligning with the convention?

```suggestion
export { default } from "./Breadcrumb";
```

rongxin-liu (approved):
Looks good overall! Left some comments worth taking a look.

PR #549, opened 2026-06-23
Add site detail building summary grid

jmarrxyz:
## Summary

- Populate the site detail buildings module with child `BuildingSummaryCard` tiles.
- Wrap the module in the rounded panel container and use normalized auto-fill grid tracks.
- Add SiteDetailPage Storybook variants for default, single, dense, long-name, empty, and narrow-width states.

## Test Plan

- `npm run lint`
- `npx tsc --noEmit`


flesher (approved):
Looks good! Only thing is its currently getting rendered at the bottom of the page where the performance section should be.

Image

PR #580, opened 2026-06-25
Support multi-scope curtailment targets

negarn:
Reviewable diff: +2585/-556 across 32 files (excludes generated, test, and story files).

## Summary
This PR lets curtailment runs and response profiles target multiple scopes at once, including any mix of selected miners and selected sites. The operator-facing Apply to UI now supports multi-site selection, mixed site/miner copy, response profile persistence, and the single-site case where selecting the only visible site must stay site-scoped instead of collapsing to the whole fleet. It also keeps newly started pending curtailments visible while active/history reads are stale, and lets org-wide curtailment operators see and stop active events whose targets include unassigned miners that cannot be mapped to a site.

https://github.com/user-attachments/assets/adb9755d-9c0e-42c0-abf9-0892ab52ffbf

## How it works
Operators choose site scopes first, then miner scopes. Site selection defaults to none with a `Select` button label; selecting more than one available site can become all-sites, but selecting the only available site remains an explicit site scope. Explicit miners and selected sites can be combined, so `Toronto + 1 miner` stays represented as compact scope entries rather than as a client-expanded list of every miner at Toronto.

The client sends repeated `CurtailmentScope` values for preview, start, response profile create/update, and active-curtailment mapping. Site IDs, site labels, whole-org selection, and explicit device identifiers stay separate across form state and API mapping, which lets the UI preserve labels and prevents the bulk miner table from needing every device ID for large sites.

The server translates legacy oneof scope fields and repeated scopes into one normalized domain `Scope`. Normalization deduplicates site IDs and device identifiers, promotes whole-org scopes when present, validates requested sites against the organization, and serializes the effective scope into `scope_json`. Candidate selection sends compact `site_ids`, `device_identifiers`, and device-set filters into SQL, where matching rows are unioned and deduplicated before dispatch so a miner selected explicitly and by site is only curtailed once.

Response profiles persist the normalized scope in `scope_json` while continuing to populate legacy `site` fields for older clients when possible. Update and delete operations compare the expected scope JSON so scoped profile writes are guarded consistently, and handler authorization derives site resource contexts from repeated scopes, persisted targets, or `scope_json` for targetless events.

For active curtailment visibility, mutation responses now seed the shared active-curtailment snapshot with a short stale-refresh grace period. If a start response returns a pending event but immediate active/history refreshes do not know about it yet, the active card and injected history row remain visible until the server catches up or terminal history confirms a real state change.

For mixed or explicit-miner events with incomplete target-site coverage, list/detail/stop authorization now falls back to org-wide curtailment permission instead of hiding the event from every list API. Site-scoped users still fail closed when an event contains unassigned targets, while org-wide operators can see the event, inspect it, and stop it from the UI.

## Diagrams
```mermaid
flowchart TD
    UI["Apply to UI"] --> ScopeBuilder["Client scope builder"]
    ScopeBuilder --> Request["Preview, start, or profile request"]
    Request --> Handler["Curtailment handler"]
    Handler --> DomainScope["Domain scope normalization"]
    DomainScope --> StoreFilters["SQL filters: site_ids and device_identifiers"]
    StoreFilters --> Targets["Deduped target miners"]
    DomainScope --> ScopeJSON["Persisted scope_json"]
    ScopeJSON --> ProfileRead["Profile and active event readback"]
    ProfileRead --> UI
```

```mermaid
sequenceDiagram
    participant Operator
    participant Client
    participant ActiveAPI
    participant HistoryAPI
    participant SharedState

    Operator->>Client: Start mixed-scope curtailment
    Client->>SharedState: Apply mutation response as active event
    Client->>HistoryAPI: Refresh history page
    Client->>ActiveAPI: Poll active curtailments
    ActiveAPI-->>Client: Stale response without new event
    HistoryAPI-->>Client: Stale page without new event
    Client->>SharedState: Preserve mutation-backed pending event
    SharedState-->>Operator: Active card remains stoppable
```

```mermaid
flowchart TD
    Event["Mixed or explicit-miner event"] --> Coverage["Resolve target site coverage"]
    Coverage --> Complete["All targets map to sites"]
    Coverage --> Incomplete["At least one target is unassigned"]
    Complete --> SiteChecks["Require permission for each site"]
    Incomplete --> OrgWide["Require org-wide curtailment permission"]
    SiteChecks --> Visible["Return event in list/detail"]
    OrgWide --> Visible
    OrgWide --> Stop["Allow stop for org-wide managers"]
```

## Areas of the code involved
| --- | --- | --- |
| `proto/curtailment/v1/curtailment.proto` | Adds composable `CurtailmentScope` entries to events, preview/start requests, and response profiles while keeping legacy oneof fields. | Defines the API contract for mixed miner/site selections and backward-compatible read/write behavior. |
| `client/src/protoFleet/features/energy` | Updates the run curtailment Apply to UI, site selector, preview text, request builders, active card scope labels, and readback mapping. | This is the operator-facing workflow for selecting and understanding mixed curtailment targets. |
| `client/src/protoFleet/api/activeCurtailmentData.ts` | Replaces count-based mutation preservation with a short time-based stale-refresh window for mutation-backed active events. | Prevents newly started pending curtailments from disappearing when active/history polling lags behind the mutation response. |
| `client/src/protoFleet/features/settings/components/Curtailment` | Reuses the same Apply to semantics for response profile create/update and stores multi-site/miner form state. | Keeps response profile targeting aligned with immediate curtailment runs. |
| `client/src/protoFleet/api` | Maps repeated scopes into and out of curtailment events and response profile payloads, including cached site labels. | Prevents label loss such as `Unknown scope` and keeps API payloads compact. |
| `client/src/protoFleet/components/MinerSelectionList.tsx` | Adds site-scoped miner-list filtering support for picker workflows. | Allows miner selection to respect site context without loading whole-site membership into the parent form. |
| `server/internal/handlers/curtailment` | Translates repeated scopes, derives scoped auth contexts, handles targetless scope JSON fallback, and falls back to org-wide permission for incomplete target-site coverage in list/detail/stop paths. | Ensures mixed scopes are authorized, visible, and recoverable consistently across API surfaces. |
| `server/internal/domain/curtailment` | Normalizes scopes, deduplicates IDs, validates site ownership, serializes `scope_json`, and applies scope behavior to runs, response profiles, automations, and reconciliation. | Centralizes whole-org, multi-site, explicit-miner, and legacy compatibility semantics. |
| `server/internal/domain/stores/sqlstores` and `server/sqlc/queries` | Adds compact SQL filters and JSON scope handling for candidate selection, cooldown lookup, response profile writes, and scoped conflict checks. | Keeps performance acceptable for thousands of selected miners or miners under selected sites. |
| `server/migrations/000099_add_response_profile_scope_json.*.sql` | Adds and backfills response profile `scope_json`. | Persists multi-scope response profile selections without dropping legacy site behavior. |
| `client/e2eTests/protoFleet` | Updates curtailment page objects and assertions for scoped request payloads. | Keeps E2E coverage aligned with the new request shape. |
| Generated code | Regenerated Go protobuf, TypeScript protobuf, and sqlc outputs. | Generated from proto/sqlc changes; skip detailed review. |

## Key Technical Decisions & Trade-Offs
| Decision | Trade-off |
| --- | --- |
| Use repeated `CurtailmentScope` entries rather than expanding selected sites into device identifiers in the UI. | Adds backend normalization, but avoids loading or sending thousands of miners per site. |
| Deduplicate in the domain/store path instead of trusting the web client. | Protects API callers outside the UI and handles overlaps between explicit miners and selected sites. |
| Persist response profile selections in `scope_json` while keeping legacy `site` fields populated where possible. | Adds migration and compatibility code, but supports mixed scopes without breaking older readers. |
| Treat whole-org/all-miners as the dominant effective target, except that selecting the only available site remains an explicit site scope. | Simplifies downstream filtering while preserving the operator's intended narrow scope in small or permission-limited fleets. |
| Preserve mutation-backed active curtailments by elapsed time instead of by missing-refresh count. | Avoids UI flicker from multiple refresh consumers, but keeps the grace window bounded so confirmed terminal history can still clear the event. |
| Require org-wide curtailment permission for events with incomplete target-site coverage. | Keeps unassigned-miner events recoverable for fleet operators while continuing to hide them from site-only users. |
| Guard scoped response profile writes with expected scope JSON. | Stricter than checking only legacy `site_id`, but prevents stale writes when mixed scopes are edited. |

## Testing & Validation
- `bin/just gen`
- `git diff --check`
- `cd client && /Users/negar/Development/proto-fleet/bin/npm run lint`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx eslint src/protoFleet/api/activeCurtailmentData.ts src/protoFleet/api/useCurtailmentApi.test.ts src/protoFleet/features/energy/CurtailmentStartModal.tsx src/protoFleet/features/energy/CurtailmentStartModal.test.tsx --max-warnings 0`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx prettier --check src/protoFleet/api/activeCurtailmentData.ts src/protoFleet/api/useCurtailmentApi.test.ts src/protoFleet/features/energy/CurtailmentStartModal.tsx src/protoFleet/features/energy/CurtailmentStartModal.test.tsx`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx tsc --noEmit`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx vite build --mode protoFleet`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx vitest run src/protoFleet/features/energy/CurtailmentStartModal.test.tsx src/protoFleet/features/energy/curtailmentRequestBuilders.test.ts src/protoFleet/features/energy/useCurtailmentPlanPreview.test.ts src/protoFleet/api/useCurtailmentResponseProfiles.test.tsx src/protoFleet/features/settings/components/Curtailment/CurtailmentSettingsPage.test.tsx src/protoFleet/features/energy/CurtailmentManagementPanel.test.tsx src/protoFleet/api/useCurtailmentApi.test.ts`
- `cd client && /Users/negar/Development/proto-fleet/bin/npx vitest run src/protoFleet/api/useCurtailmentApi.test.ts src/protoFleet/features/energy/CurtailmentStartModal.test.tsx`
- `cd server && /Users/negar/Development/proto-fleet/bin/just lint`
- `cd server && /Users/negar/Development/proto-fleet/bin/go test ./internal/domain/curtailment/... ./internal/handlers/curtailment`
- `cd server && /Users/negar/Development/proto-fleet/bin/go test ./internal/handlers/curtailment`
- Post-merge validation: `/Users/negar/Development/proto-fleet/bin/go test ./server/internal/handlers/curtailment ./server/internal/domain/curtailment -count=1`
- Post-merge validation: `cd client && /Users/negar/Development/proto-fleet/bin/npx vitest run src/protoFleet/features/settings/components/SettingsLayout.test.tsx src/protoFleet/features/energy/ActiveCurtailmentStatus.test.tsx src/protoFleet/features/energy/CurtailmentManagementPanel.test.tsx src/protoFleet/features/settings/components/Curtailment/CurtailmentSettingsPage.test.tsx`
- Attempted `cd server && /Users/negar/Development/proto-fleet/bin/go test ./internal/domain/curtailment/... ./internal/domain/stores/sqlstores ./internal/handlers/curtailment`; the sqlstores package could not connect to the local Postgres test database because authentication failed for user `fleet`.
- Full Playwright E2E was not run locally.


server/internal/handlers/curtailment/handler.go
```diff
@@ -372,20 +380,75 @@ func errCurtailmentNotImplemented(rpc string) error {
...
-	return authz.ResourceContext{}
+	return nil, nil
+}
+
+func scopeResourceContexts(scopes []*pb.CurtailmentScope) ([]authz.ResourceContext, error) {
+	scope, err := toCompositeScope(scopes)
+	if err != nil {
+		return nil, err
+	}
+	return siteResourceContextsForScope(scope), nil
```
rongxin-liu: Perhaps worth resolving explicit device identifiers into permission contexts here before returning the site-only contexts. Right now a composite scope such as `site_ids=[allowed] + device_identifiers=[denied-site-miner]` only checks the site selector, while the service still executes the explicit device selector later.

A stronger next step could be to make scope auth use the same normalized selector model as execution: resolve device identifiers to current site IDs, merge those with explicit site IDs, and require org-wide curtailment permission when any explicit device is unassigned or cannot be resolved. That keeps mixed scopes compact without turning auth into a client-side assumption.

server/internal/handlers/curtailment/response_profiles.go
```diff
@@ -142,26 +167,160 @@ func (h *Handler) getResponseProfileWithSitePermission(ctx context.Context, orgI
...
+			if requireKnownDevices {
+				return nil, fleeterror.NewNotFoundError("one or more device identifiers were not found")
+			}
+			continue
+		}
+		if siteID != nil {
+			siteIDs = append(siteIDs, *siteID)
+		}
+	}
+	return siteResourceContextsForScope(domainCurtailment.Scope{SiteIDs: siteIDs}), nil
```
rongxin-liu: Maybe this should treat incomplete device-site coverage as a distinct authorization case instead of silently dropping it from the site-context set. For explicit miner scopes, a missing row or `NULL` site currently means the profile can end up with no site contexts to check, which is risky for site-narrowed curtailment managers.

Could we make the next step explicit here: resolve all requested device identifiers, authorize every resolved site, and require org-wide curtailment permission when any identifier is missing or unassigned. That would also give list/get/delete the same fail-closed behavior as create/update without relying on a workaround at the caller.

server/internal/handlers/curtailment/response_profiles.go
```diff
@@ -307,6 +476,21 @@ func responseProfileFromPayload(
...
+		scopeJSON, err := domainCurtailment.MarshalScopeJSON(scope)
+		if err != nil {
+			return models.ResponseProfile{}, err
+		}
+		if scope.Type == models.ScopeTypeWholeOrg {
+			scopeJSON = []byte(`{"whole_org":true}`)
+		}
+		profile.ScopeJSON = scopeJSON
+		profile.SiteID = legacySiteIDForScope(scope)
+	}
```
rongxin-liu, line 516: Worth distinguishing an omitted scope from an explicit whole-org scope in the update path. As written, an update request that sends neither legacy `site` nor new `scopes` builds a profile with empty `ScopeJSON`; the store normalizes that to `{}`, so an existing mixed/multi-site profile can be broadened to whole-org by a partial or legacy client updating an unrelated field.

Maybe the next step is to carry scope intent through the payload parser: preserve the existing scope when the update omits scope fields, and require clients to send an explicit `whole_org` scope when they intentionally broaden. That keeps backward compatibility without making scope loss the default behavior.

server/migrations/000100_add_response_profile_scope_json.down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE curtailment_response_profile
+    DROP CONSTRAINT IF EXISTS ck_curtailment_response_profile_scope_json_object;
+
+ALTER TABLE curtailment_response_profile
+    DROP COLUMN IF EXISTS scope_json;
```
rongxin-liu, line 45: This rollback path perhaps needs a safety gate before dropping `scope_json`. After this PR ships, multi-site and explicit-miner response profiles may have `site_id = NULL` with their real selector stored only in `scope_json`; rolling back would drop that selector, and legacy code would interpret the row as whole-org.

A useful next step could be to make the down migration preserve only representable single-site scopes back into `site_id`, and fail fast if any rows contain multi-site/device/device-set scopes that legacy schema cannot represent. That gives operators a deliberate cleanup/disable path instead of a rollback that can broaden curtailments.

client/src/protoFleet/api/useCurtailmentResponseProfiles.ts
```diff
@@ -154,6 +215,101 @@ function mapApiResponseProfile(profile: ApiCurtailmentResponseProfile, siteNameB
...
+    switch (scope.scope.case) {
+      case "wholeOrg":
+        return {
+          siteSelection: "allSites",
+          siteId: "",
+          siteName: "",
+          siteIds: [],
+          siteNamesById: {},
+          deviceIdentifiers: [],
+        };
```
rongxin-liu, line 243: Maybe explicit `whole_org` readback should map to the same executable form state as the UI's all-miners selection. This branch returns `siteSelection: "allSites"` with no site IDs and no `minerSelectionMode: "all"`; after a reload, the preview/start builder treats `allSites` as requiring concrete site IDs and rejects the profile.

Could the next step be to normalize this case to `minerSelectionMode: "all"` / whole-fleet form values, or teach the shared scope builder to treat an explicit whole-org profile as whole-org rather than empty all-sites? That should keep API-created profiles round-trippable through the UI.

rongxin-liu (approved):

PR #603, opened 2026-06-26
perf(server-test): clone a migrated template DB per test

ankitgoswami:
## Summary

Cuts the **Server Checks / Test** job's dominant cost. The test step (~363s on CI) is gated by DB-backed packages that each rebuilt their database from scratch: every test ran `CREATE DATABASE` + replayed all ~99 migrations (~0.5-0.7s each). `domain/stores/sqlstores` alone does this ~143 times → ~227s, and it is the single-package floor for the whole job.

This builds **one migrated template database per test process** and provisions each test's database as a fast `CREATE DATABASE ... TEMPLATE` file-copy, skipping the per-test migration replay entirely.

## Mechanism

`testutil.GetTestDB` (the single entry point all DB-backed tests funnel through, directly or via `NewDatabaseService` / `InitializeDBServiceInfrastructure`) now:

1. Builds a fully-migrated template database **once per process** (`sync.Once`), keyed by PID so the concurrent `go test` package processes each own their template on the shared instance. The migrator connection is closed before any clone, since `CREATE DATABASE ... TEMPLATE` requires no sessions on the source.
2. Provisions each test database as `CREATE DATABASE <unique> TEMPLATE <template>` instead of running migrations.

Per-test isolation is **unchanged** — every test still gets its own physical database with a unique name. The lock-heavy TimescaleDB continuous-aggregate DDL now runs once per process instead of once per test, which *removes* the concurrent-migration deadlock surface the existing retry logic guards against. The pinned-`DB_NAME` path (non-default) keeps the migrate-in-place behavior.

## Measured impact (local, warm TimescaleDB)

| Package | Before | After |
|---|---|---|
| `domain/authz` | 14.6s | **3.6s** |
| `domain/stores/sqlstores` | ~80s | **17.7s** |
| `domain/fleetmanagement` | (CI 98s) | 10.4s |
| `domain/fleetnode/pairing` | (CI 113s) | 6.7s |
| `domain/command` | (CI 58s) | 11.7s |

CI's slowest package (`sqlstores`, 227s) should drop well under 100s, pulling the whole Test job down with it. Local numbers are on a fast disk; CI ratios will differ but the per-test migration replay — the dominant cost — is eliminated either way.

## Two pre-existing flakes fixed along the way

Stress-running the full suite (and CI on a sibling branch) surfaced two **pre-existing** races, unrelated to the DB change but worth fixing so the faster suite stays green:

- **`fix(fleetnode)`** — `closeTelemetryDeviceAsync` acquired a slot on the global token channel but its release goroutine re-read the global at exit. Benign in production (the global never changes) but tests swap it, so a slow goroutine from one test could drain the next test's token and break its assertion under load. Now captures the channel locally. Root-caused to the adjacent `TestCloseTelemetryDeviceAsyncDoesNotBlock` → `...WhenCloseWorkersAreExhausted` ordering.
- **`test(timescaledb)`** — `fleet_telemetry_poll_heartbeat` has a 30s background refresh policy (migration 000082); the ingest-stalled-rule test's manual `refresh_continuous_aggregate` can collide with the scheduled one and fail with `SQLSTATE 55P03`. The helper now retries that transient conflict.

## Verification

- Full server suite (`go test ./... -skip ./generated/...`) run **13x** with all three changes: 0 failures, 0 telemetry flakes.
- `go vet`, `golangci-lint` (0 issues), and `golangci-lint fmt --diff` all clean on the changed packages.
- Each fix root-caused from source, not guessed.

## Notes

- Independent of [#602](https://github.com/block/proto-fleet/pull/602) (which caches the TimescaleDB image in the same job) — different files, mergeable in any order. Together they target the build/startup and the test-execution halves of the Server Test job.
- Follow-up (not in this PR): adding `t.Parallel()` to the DB-backed tests now that the template removes the per-test migration cost; with isolated per-test DBs this is safe but touches many files and should be its own change (and needs the `generateTestDBName` entropy widened first).

### DIFF
--- a/server/cmd/fleetnode/telemetry.go
+++ b/server/cmd/fleetnode/telemetry.go
@@ -260,15 +260,18 @@ type telemetryDeviceCloser interface {
 }
 
 func closeTelemetryDeviceAsync(device telemetryDeviceCloser) bool {
+	// Capture the channel so the worker releases the exact slot it acquired,
+	// even if telemetryDeviceCloseTokens is reassigned later (tests swap it).
+	tokens := telemetryDeviceCloseTokens
 	select {
-	case telemetryDeviceCloseTokens <- struct{}{}:
+	case tokens <- struct{}{}:
 	default:
 		slog.Warn("telemetry device close worker capacity is exhausted; closing synchronously")
 		closeTelemetryDevice(device)
 		return false
 	}
 	go func() {
-		defer func() { <-telemetryDeviceCloseTokens }()
+		defer func() { <-tokens }()
 		closeTelemetryDevice(device)
 	}()
 	return true
--- a/server/internal/infrastructure/timescaledb/ingest_stalled_rule_test.go
+++ b/server/internal/infrastructure/timescaledb/ingest_stalled_rule_test.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


ankitgoswami:
Closing for now — holding off on the template-DB test-speedup change. Branch `perf/server-test-template-db` is preserved if we revisit.

PR #609, opened 2026-06-29
fix(fleetnode): support cooling mode reads

ankitgoswami:
Reviewable diff: +62/-2 across 3 files (excludes generated, test, and story files).

## Summary
Fleetnode-paired miners can now report their current cooling mode, so the single-miner cooling-mode flow can prepopulate from remote devices instead of falling back through an unsupported remote adapter path. The read is bounded and validates the returned enum at the gateway boundary, closing the read-side parity gap with combined/direct plugin mode while keeping the existing set-cooling command path unchanged.

## How it works
When the fleet-management service asks a remote fleetnode miner for its cooling mode, the remote miner sends a `GetCoolingModeAction` through the existing fleetnode gateway control command envelope with a short command timeout. The fleet node reconstructs the local SDK device from the command target, calls `GetCoolingMode`, converts the SDK cooling enum back into the shared proto enum, and returns a validated `GetCoolingModeResult` in the ack payload. The cloud-side remote miner validates and unmarshals that payload, then returns the mode to the existing fleet-management caller.

## Diagrams
```mermaid
flowchart LR
  UI["ProtoFleet cooling modal"] --> API["FleetManagement GetMinerCoolingMode"]
  API --> Remote["RemoteFleetNodeMiner"]
  Remote --> Gateway["Fleetnode control command"]
  Gateway --> Node["fleetnode runMinerAction"]
  Node --> Miner["SDK device GetCoolingMode"]
  Miner --> Node
  Node --> Gateway
  Gateway --> Remote
  Remote --> API
```

```mermaid
sequenceDiagram
  participant Cloud as "fleetd cloud server"
  participant Stream as "ControlStream"
  participant Node as "fleetnode"
  participant Device as "miner SDK device"
  Cloud->>Stream: MinerCommand get_cooling_mode
  Stream->>Node: deliver command
  Node->>Device: GetCoolingMode()
  Device-->>Node: sdk.CoolingMode
  Node-->>Stream: ack payload GetCoolingModeResult
  Stream-->>Cloud: common.v1.CoolingMode
```

## Areas of the code involved
| --- | --- | --- |
| `proto/fleetnodegateway/v1/fleetnodegateway.proto` | Adds `GetCoolingModeAction` and a validated `GetCoolingModeResult`. | This is the node/cloud contract for the new read command and returned enum boundary. |
| `server/internal/domain/miner/remotenode` | Replaces the unsupported `GetCoolingMode` stub with bounded command dispatch, ack validation, and result decoding. | This is the cloud-side behavior that makes fleetnode-paired miners match direct plugin reads without holding a node command slot indefinitely. |
| `server/cmd/fleetnode` | Handles the new command by calling the SDK device and converting SDK cooling modes back to proto values. | This is the node-side execution path and enum boundary. |
| `server/generated/grpc/fleetnodegateway/v1/fleetnodegateway.pb.go` | Generated — skip. | Regenerated from the gateway proto. |
| `client/src/protoFleet/api/generated/fleetnodegateway/v1/fleetnodegateway_pb.ts` | Generated — skip. | Regenerated from the gateway proto. |
| Tests | Adds focused remote-adapter and fleetnode-handler coverage, including invalid enum payloads and bounded read timeout. | Verifies command encoding, payload decoding, SDK invocation, enum conversion, validation, and timeout behavior. |

- Reused the existing fleetnode miner-command ack payload pattern instead of adding a new RPC, keeping reads on the established ControlStream mechanism.
- Bounded the remote cooling-mode read with the same style as quick status/error reads so a hung node/plugin cannot hold the per-node command gate until the caller context expires.
- Returned `COOLING_MODE_UNSPECIFIED` for unknown SDK cooling values on the node side, matching the direct plugin adapter's tolerant mapping, while rejecting undefined enum values received in the cloud ack payload.
- Kept SDK-to-proto cooling-mode conversion local to each boundary instead of adding a shared helper package for a four-case mapping.

## Testing & validation
- `bin/buf lint`
- `go test ./internal/domain/miner/remotenode ./cmd/fleetnode`
- `golangci-lint run -c .golangci.yaml ./internal/domain/miner/remotenode ./cmd/fleetnode`
- `git diff --check`

`just gen` was attempted first, but local client dependency installation failed with npm tarball integrity errors in this temporary worktree. The committed gateway Go and TS generated outputs were regenerated with the repo-pinned Buf/tooling path, using an existing sibling worktree's installed `protoc-gen-es` for the TS plugin.

- Watch fleetnode command ack logs for `get_cooling_mode` failures, invalid payload validation errors, or `ACK_CODE_UNIMPLEMENTED` responses after deploy.
- Healthy signal: single-miner cooling-mode reads for fleetnode-paired miners return a concrete `common.v1.CoolingMode` or `UNSPECIFIED` without surfacing unsupported-operation errors.
- Failure signal: increased `GetMinerCoolingMode` errors for fleetnode-paired miners, especially invalid payload, unmarshal, or command timeout errors.
- Mitigation: the UI already tolerates read failure by opening without a preselected mode; revert this PR if the command path causes unexpected fleetnode ack failures.
- Validation window: first deploy after merge; owner: fleetnode feature owner/on-call.


server/cmd/fleetnode/minercommand.go
```diff
@@ -305,6 +305,12 @@ func runMinerAction(ctx context.Context, client gatewayClient, commandID, firmwa
 			return nil, err
 		}
 		return nil, dev.SetCoolingMode(ctx, mode)
+	case *pb.MinerCommand_GetCoolingMode:
+		mode, err := dev.GetCoolingMode(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return getCoolingModeResultPayload(toProtoCoolingMode(mode))
```
rongxin-liu, line 313: Maybe worth adding a node-side supervisor or dedupe for this read. The cloud wait is bounded, but if `dev.GetCoolingMode(ctx)` ignores context, this handler keeps a `cmdSem` lane after the cloud gate releases. Should we make `GetCoolingMode` release/ack on a node-side timeout, or throttle duplicate reads per device, so UI prefill cannot starve unrelated commands?

rongxin-liu (approved):

server/cmd/fleetnode/minercommand.go
ankitgoswami: good callout and something which applies to other actions as well. I'd like to come back to this in a separate hardening PR that also does a pass for other node actions.

ankitgoswami: https://github.com/block/proto-fleet/issues/615

PR #635, opened 2026-07-01
feat(alerts): deliver org Slack/webhook alerts from fleet-api

illegalprime:
## Summary

Org alert delivery moves **out of Grafana and into fleet-api**. Previously each org's Slack/webhook destination was a native Grafana contact point, and Grafana delivered the notification directly. That message couldn't be cleaned up — Grafana's native Slack notifier hardcodes the attachment `title_link` to the Grafana rule URL and the `footer_icon` to the Grafana favicon, its attachment text isn't mrkdwn, and its DB role can't read the `device` table for a friendly name.

Now fleet-api owns delivery: it renders a fully-controlled **Slack Block Kit** message (no Grafana branding, a header linking to the proto-fleet instance, friendly device name + MAC) or a clean JSON payload for webhook channels. Grafana keeps what it's good at — rule evaluation, silences (rule pause / maintenance windows), and the single internal history webhook.

## How it works

- Org channel destinations move from Grafana contact points into a new `alert_channel` table. The Slack URL / webhook bearer is the secret, stored **encrypted** (AES-GCM via the existing `encrypt.Service`) because Grafana redacts secrets on read and fleet-api needs the real value to POST.
- Grafana still evaluates rules and POSTs every alert batch to the existing internal Alertmanager webhook (which already persisted `notification_history`). After the history write, the handler hands the parsed batch to a new `Deliverer`.
- The `Deliverer` groups the batch by org (dropping operator-only `proto_fleet_scope=internal` and org-less alerts), loads that org's channels, resolves device name/MAC in one query (the same `device`/`discovered_device` join notification-history uses), renders per channel kind, and POSTs — re-checking each destination against the SSRF policy at send time.
- Rule pause and maintenance windows still work end-to-end because a Grafana silence suppresses delivery to the internal webhook *before* fleet-api ever sees the alert.

## Diagrams

```mermaid
flowchart LR
  G[Grafana: rules + silences] -->|POST batch| W[/internal/alertmanager-webhook/]
  W --> H[(notification_history)]
  W --> D[Deliverer]
  D -->|group by org, drop internal/org-less| R{channel kind}
  R -->|slack| S[Block Kit → org Slack webhook]
  R -->|webhook| J[clean JSON → org webhook]
  D -.decrypt.-> C[(alert_channel<br/>encrypted config)]
  D -.resolve name/MAC.-> DB[(device + discovered_device)]
```

## Areas of the code involved

- `server/migrations/000106_alert_channel.*`, `server/sqlc/queries/alert_channel.sql` — new encrypted channel table + queries (channel CRUD and `GetDeviceIdentities`).
- `server/internal/domain/stores/sqlstores/alert_channel.go` — store implementing `ChannelStore` + `DeviceIdentityLookup`.
- `server/internal/domain/alerts/{store,deliver,render}.go` — domain interfaces, the `Deliverer` (fan-out, SSRF re-check, `SendTest`), and the Block Kit / webhook renderers.
- `server/internal/domain/alerts/service.go`, `grafana_client.go` — channel CRUD repointed to the encrypted store; notification-tree/reconcile and Grafana contact-point/receiver-test code removed. Rules + maintenance (Grafana silences) untouched.
- `server/internal/handlers/alertmanagerwebhook/handler.go`, `cmd/fleetd/{main,config}.go`, docker-compose alert overlays — wire the deliverer + `FLEET_PUBLIC_URL`; drop the reconcile loop.

- **Secret storage.** Grafana redacts contact-point secrets on read, so delivery from fleet-api requires a local copy. Stored encrypted with the existing service master key; reads never return the secret (only `HasSecret` + a host-only webhook URL). Secret-carry-on-rename is preserved and never carries a credential onto a changed destination.
- **Inline delivery.** Delivery runs synchronously before the `204`, bounded by a 25s total / 10s per-send timeout, so it inherits Grafana's grouping/dedup and is at-least-once. Trade-off: a very slow fan-out could cross Grafana's webhook timeout and trigger a retry (duplicate messages). Mitigated with a panic-recover, a per-send cap, and a Block Kit size cap; full async decoupling was deliberately deferred.
- **mrkdwn safety.** Device names are user-controlled, so all dynamic text is escaped (`& < >`) to prevent link injection / broken rendering, and the message is capped under Slack's 50-block / 3000-char / 150-rune limits.
- **Migration note.** Existing org channels can't be auto-migrated (their secrets are redacted in Grafana); orgs re-add destinations. Operators should also clear any stale Grafana per-org contact points / dynamic notification tree so Grafana doesn't keep double-delivering the old-format message. Acceptable because alerts are still beta / `FLEET_ALERTS_ENABLED`-gated.

## Testing & validation

- `just lint` — buf, eslint, golangci-lint all **0 issues**.
- Targeted `go test` (with DB) for `internal/domain/alerts/...`, `internal/domain/stores/sqlstores/...`, `internal/handlers/alertmanagerwebhook/...`, `cmd/fleetd/...` — **all pass**, including the `alert_channel` store integration test (CRUD + encryption round-trip + soft-delete reuse + device-identity resolution) and the webhook handler integration tests against a live database.
- New unit tests: Block Kit renderer (no Grafana leakage, instance link, name/MAC, mrkdwn escaping, block-cap overflow), deliverer (httptest capture, per-org fan-out, internal-scope exclusion, SSRF refusal, test-send), and channel-service tests (fake store + real cipher + stub tester).
- Generated sqlc code committed in sync; only a new migration (`000106`) added.
- Not run: E2E (docker-compose). Manual follow-up worth doing — trigger a firing alert with `FLEET_ALERTS_ENABLED=true` and `FLEET_PUBLIC_URL` set and confirm the Slack message shape, then pause the rule and confirm delivery stops.


mcharles-square (approved):
worth addressing the 🤖 comment `[MEDIUM] Large alert batches are accepted but persisted one row at a time`

PR #645, opened 2026-07-02
fix(protoos): keep KPI tab navigation inside embedded miner view

mcharles-square:
Reviewable diff: +116/-6 across 2 files (excludes generated, test, and story files). The load-bearing code change is +9/-6 in a single file (`TabMenuWrapper.tsx`); the remaining +107 is a `docs/solutions/` note.

## Summary

Clicking the **Hashrate / Efficiency / Power Usage / Temperature** stat cards in the embedded single-miner view (`/miners/:id/*`) kicked the user back to the fleet dashboard instead of switching KPI tabs. The KPI tab strip navigated with absolute paths (`/hashrate`), which escape the embed. This wires the tab strip's base path to `minerRoot` so navigation stays inside the hosted miner view. Standalone ProtoOS is unaffected (`minerRoot` is `""` there).

## How it works

ProtoOS runs in two shells: standalone (routes live at `/hashrate`, `/efficiency`, …) and embedded inside ProtoFleet's single-miner view at `/miners/:id/*`. In-view navigation must be relative to whichever shell is hosting, so `MinerHostingContext` exposes `minerRoot` — `/miners/:id` when fleet-hosted, `""` when standalone.

The KPI tab strip (`TabMenuWrapper` → shared `TabMenu`) was the one navigation that didn't prefix `minerRoot`: `KpiLayout` rendered `<TabMenu />` with no `basePath`, so the shared component defaulted it to `""` and called `navigate("/hashrate")`.

- **Standalone:** `/hashrate` is a real route → fine.
- **Fleet-embedded:** the target should be `/miners/:id/hashrate`, but it navigated to `/hashrate`. ProtoFleet's router matches that against its `/:siteScope` route (treating `"hashrate"` as a site slug), fails to resolve it, and `SiteScopeLayout` renders `<Navigate to="/" replace />` → `appEntryLoader` → the dashboard. A quieter side effect: the active-tab check (`basePath + path === pathname`) never matched, so no tab appeared selected in fleet mode.

The fix reads `minerRoot` from `useMinerHosting()` and passes it as `basePath`, matching the idiom already used by `Navigation`, `NoPoolsCallout`, and `PoolStatusWrapper`. The dead `basePath` prop on `TabMenuWrapper` (declared, never passed) is removed.

## Diagrams

```mermaid
flowchart TD
    Click["Click KPI card in /miners/:id/*"] --> Wrapper["TabMenuWrapper"]
    Wrapper --> Root{"basePath source"}
    Root -->|"before: none, defaults to empty"| Abs["navigate('/hashrate')"]
    Root -->|"after: minerRoot"| Scoped["navigate('/miners/:id/hashrate')"]
    Abs --> SiteScope["ProtoFleet '/:siteScope' route treats 'hashrate' as a site"]
    SiteScope --> Redirect["Unknown site -> Navigate to '/' -> dashboard"]
    Scoped --> Stay["Hashrate tab renders inside embedded view"]
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/protoOS/features/kpis/components/TabMenu/TabMenuWrapper.tsx` | Read `minerRoot` from `useMinerHosting()`, pass as `basePath`; drop unused `basePath` prop | The whole fix. Confirm `minerRoot` is the right base-path source and standalone (`""`) is preserved |
| `client/src/protoOS/features/kpis/components/TabMenu/TabMenuWrapper.test.tsx` | New test | Covers standalone (`/hashrate`) and fleet-hosted (`/miners/miner-1/efficiency`) navigation |
| `docs/solutions/logic-errors/kpi-tab-navigation-needs-minerroot-prefix-2026-07-02.md` | New solution note | Documents the "ProtoOS navigation must prefix `minerRoot`" convention so the next embedded consumer doesn't repeat it |

- **Fix in `TabMenuWrapper`, not shared `TabMenu`.** `TabMenuWrapper` is the ProtoOS-specific wrapper that already has access to `useMinerHosting()`; the shared `TabMenu` stays presentational and shell-agnostic. Alternative — defaulting `basePath` inside shared `TabMenu` — was rejected because `shared/` cannot import from `protoOS/` (component-boundary rule) and shouldn't know about hosting.
- **Reuse `minerRoot` rather than a new prop drill.** Matches the three existing sibling navigations, so there's one convention to reason about.

## Testing & validation

- New `TabMenuWrapper.test.tsx`: asserts `/hashrate` in standalone and `/miners/miner-1/efficiency` when fleet-hosted. Both pass.
- Existing shared `TabMenu` tests still pass.
- `tsc --noEmit` clean; `eslint` clean on changed files; lefthook pre-commit (format) and pre-push (typecheck) passed.
- Not covered: no E2E added; verified via unit tests and reasoning about the ProtoFleet `/:siteScope` redirect path rather than a live click-through.

### DIFF
--- a/client/src/protoOS/features/kpis/components/TabMenu/TabMenuWrapper.test.tsx
+++ b/client/src/protoOS/features/kpis/components/TabMenu/TabMenuWrapper.test.tsx
@@ -1,0 +1,56 @@
+import { useNavigate } from "react-router-dom";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, it, type Mock, vi } from "vitest";
+
+import TabMenuWrapper from "./TabMenuWrapper";
+import { useMinerHosting } from "@/protoOS/contexts/MinerHostingContext";
+
+const mockNavigate = vi.fn();
+
+vi.mock("react-router-dom", async () => {
+  const actual = await vi.importActual("react-router-dom");
+  return {
+    ...actual,
+    useNavigate: vi.fn(),
+    useLocation: vi.fn(() => ({ pathname: "/", search: "", hash: "", state: null, key: "default" })),
+  };
+});
+
+vi.mock("@/protoOS/contexts/MinerHostingContext", () => ({
+  useMinerHosting: vi.fn(),
+}));
+
+vi.mock("@/protoOS/store", () => ({
+  useMiner: vi.fn(() => undefined),
+  useTemperatureUnit: vi.fn(() => "F"),
+  convertAndFormatMeasurement: vi.fn(() => "--"),
+  convertValueUnits: vi.fn(() => ({ value: null })),
+}));
+
+describe("TabMenuWrapper", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    (useNavigate as Mock).mockReturnValue(mockNavigate);
+  });
+
+  it("navigates to absolute KPI paths in standalone protoOS (empty minerRoot)", () => {
+    (useMinerHosting as Mock).mockReturnValue({ minerRoot: "" });
+
+    render(<TabMenuWrapper />);
+    fireEvent.click(screen.getByText("Hashrate").closest("button") as HTMLButtonElement);
+
+    expect(mockNavigate).toHaveBeenCalledWith("/hashrate");
+  });
+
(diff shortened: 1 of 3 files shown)


negarn (approved):
super cool!

PR #654, opened 2026-07-03
feat(rack): assign miners by scanning their QR sticker (MAC or serial)

mcharles-square:
Reviewable diff: +888/-0 across 12 files (excludes generated, test, and story files).

## Summary

Lets an operator on a mobile browser assign a miner to a rack slot by **scanning the miner's QR sticker** instead of searching for it by hand. During rack creation/edit, tapping a slot now offers a third option — **"Scan QR code"** — alongside "Select from list" / "Search miners". The scanned code is resolved to a paired miner, shown for confirmation, and dropped into the selected slot via the existing assignment path.

The miner sticker carries **both** a MAC address and a serial number, and the QR payload can be either, so the lookup is **format-agnostic**: it accepts a MAC *or* a serial and routes accordingly.

## How it works

1. **Scan** — The slot popover opens `ScanMinerQrModal`. In a secure context (HTTPS or `localhost`) it opens a live rear-camera stream and polls frames with a QR decoder. On plain HTTP — where `getUserMedia` is unavailable — it falls back to a photo-capture button (`<input capture>`), which needs no secure context. This **progressive enhancement** matches the default install (HTTP over a LAN IP), so scanning still works there, just with a per-miner photo instead of a live viewfinder.
2. **Parse** — `parseScannedIdentifier` strips a label prefix (`SN:`, `S/N`, `SERIAL …`, `MAC:`, `MAC ADDRESS …`) and whitespace. For prefix-less payloads it detects a MAC by shape (colon/dash-separated hex pairs, or 12 bare hex digits) and otherwise leaves the type unspecified. It returns `{ value, type }`.
3. **Resolve** — The client calls `LookupMinerByIdentifier` with the value and a type hint. The server routes a MAC to the MAC lookup (normalizing separators/case) and a serial to the serial lookup; when the type is `UNSPECIFIED` it infers from the value shape. It returns a fully hydrated `MinerStateSnapshot` (name/model/status/placement) for operator confirmation; NotFound becomes a friendly "no paired miner" message.
4. **Confirm & assign** — On confirmation the miner is added to the rack and placed in the selected slot, reusing the exact code path the "Search miners" flow already uses. A miner already in a *different* rack is blocked.

### QR engine & on-prem

The decoder is `barcode-detector` (native `BarcodeDetector` on Android; ZXing-C++ WASM fallback on iOS/Safari/Firefox). The WASM binary is **self-hosted as a same-origin, content-hashed Vite asset** via a `locateFile` override — the library otherwise fetches it from a public CDN, which would break air-gapped installs. It is lazy-loaded only into the rack-modal route chunk.

## Diagrams

```mermaid
flowchart LR
    A["Tap rack slot"] --> B["Slot popover"]
    B --> C["Scan QR code"]
    C --> D{"Secure context?"}
    D -->|"HTTPS / localhost"| E["Live camera stream"]
    D -->|"plain HTTP"| F["Photo capture (input capture)"]
    E --> G["Decode QR (barcode-detector)"]
    F --> G
    G --> H["parseScannedIdentifier: strip prefix, detect MAC vs serial"]
    H --> I["LookupMinerByIdentifier RPC (value + type)"]
    I -->|"found"| J["Confirm miner, then assign to slot"]
    I -->|"NotFound"| K["Unknown-identifier message"]
    J --> L["Existing add-to-rack + slot-assign path"]
```

```mermaid
flowchart TD
    S["identifier + type"] --> T{"type?"}
    T -->|"MAC_ADDRESS"| M["GetPairedDeviceByMACAddress (normalized)"]
    T -->|"SERIAL_NUMBER"| N["GetPairedDeviceBySerialNumber (verbatim)"]
    T -->|"UNSPECIFIED"| U{"MAC-shaped?"}
    U -->|"yes"| M
    U -->|"no"| N
    M --> H["Hydrate snapshot"]
    N --> H
```

```mermaid
stateDiagram-v2
    [*] --> scanning
    scanning --> lookingUp: code decoded
    lookingUp --> found: identifier resolved
    lookingUp --> notFound: no paired miner
    lookingUp --> error: transport / server error
    found --> scanning: "Scan another"
    notFound --> scanning: "Try again"
    error --> scanning: "Try again"
    found --> [*]: "Assign to slot"
```

## Areas of the code involved

| --- | --- | --- |
| `proto/fleetmanagement/v1/fleetmanagement.proto` | New `LookupMinerByIdentifier` RPC + `MinerIdentifierType` enum | The API contract; identifier + type hint, validation caps |
| `server/.../fleetmanagement/service.go` | Route MAC/serial/UNSPECIFIED to the right store lookup; infer by shape | Core server logic; reuses two existing store methods + hydration helper |
| `server/.../handlers/fleetmanagement/handler.go` | Connect handler with `miner:read` guard | Authz boundary for the new RPC |
| `server/generated/**`, `client/.../generated/**` | Regenerated from proto | generated — skip |
| `client/.../api/lookupMinerByIdentifier.ts` | RPC wrapper; forwards value + type; found / notFound / error | Maps NotFound to an expected UI state, not an error |
| `client/.../utils/parseScannedIdentifier.ts` | Prefix stripping + MAC-vs-serial shape detection | Encodes both observed label formats; edge cases in tests |
| `client/.../hooks/useQrScanner.ts` | Camera lifecycle + decode loop; secure-context detection | getUserMedia teardown, decode polling, error mapping |
| `client/.../utils/initBarcodeScanner.ts` | Self-host WASM via `locateFile` | The on-prem-critical bit; avoids CDN fetch at runtime |
| `client/.../ManageRackModal/ScanMinerQrModal.tsx` | Container: camera + lookup state machine | Async race guarding (`lookupSeq`), phase transitions |
| `client/.../ManageRackModal/ScanMinerQrModalView.tsx` | Presentational shell for all phases | Pure view; storyable without camera/backend |
| `client/.../ManageRackModal/{ManageRackModal,RackPane}.tsx` | Wire "Scan QR code" into the popover + assign path | Reuses `handleSearchMinerConfirm`-style placement |
| `client/package.json` | Add `barcode-detector` | **New runtime dependency — needs sign-off** |

- **Single format-agnostic RPC** (`identifier` + `MinerIdentifierType`) over one RPC per key — the sticker carries both a MAC and a serial and the QR may encode either, so the flow must not assume. The type hint is sent from the parsed prefix; `UNSPECIFIED` lets the server infer from shape (robust when a QR carries no prefix). Both underlying store lookups already existed.
- **`barcode-detector`** over `jsQR` (unmaintained since 2021) or native-only (`BarcodeDetector` is flag-gated on iOS Safari, so unusable for iPhone operators).
- **Self-hosted WASM** over the library's default CDN fetch — required for air-gapped / on-prem sites.
- **Progressive enhancement** over gating the feature behind HTTPS — the default install serves plain HTTP over a LAN IP, so a hard HTTPS requirement would break scanning there. Install flow intentionally untouched.
- **Container / presentational split** so the visual states are covered by Storybook + tests without a camera or backend.

## Testing & validation

- **Server:** unit tests for empty/whitespace identifier (InvalidArgument), store NotFound propagation, input trimming, explicit MAC lookup, both `UNSPECIFIED` auto-detect branches (MAC-shaped → MAC, else serial), and a hydrated found-path. `go build` + `golangci-lint` clean.
- **Client:** unit tests for the parser (SN:/MAC: prefixes, shape inference, the "SN-prefixed value" and "16-digit not a MAC" edge cases), the lookup API (found/notFound/NotFound-error/other-error + type forwarding), and the modal state machine (found → assign, not-found, in-other-rack block, HTTP photo fallback, error). `tsc` + eslint clean; Storybook build succeeds; production build confirms same-origin WASM; all existing fleet-management tests pass.
- **Not covered / follow-ups:** real-device iOS Safari camera + `capture` behavior (worth a manual pass); a "scan → auto-advance to next empty slot" rapid-entry loop; the handler-level DB integration test needs the docker-compose Postgres (unit tests are mock-based).

> Note: unrelated pre-existing `server/go.mod` change (promoting `golang.org/x/text` to a direct dep) was deliberately left out of this branch.


mcharles-square:
IMG_7136
IMG_7137
IMG_7138

mcharles-square:
🤖 Addressing the Codex Security Review (commit 34318972e):

> **[MEDIUM] QR scan can assign miners outside the existing paired-only eligibility rule** … Operators can place auth-needed/default-password miners into rack membership via QR scan, bypassing the existing rack assignment UI eligibility rule.

Confirmed — `LookupMinerByIdentifier`'s store lookups match `pairing_status IN ('PAIRED','AUTHENTICATION_NEEDED','DEFAULT_PASSWORD')`, while the list/search assignment flows allow only `FLEET_SELECTABLE_PAIRING_STATUSES` (`PAIRED`). Fixed on the client, matching how the other flows enforce this: `ScanMinerQrModalView` now disables **Assign to slot** for any non-`PAIRED` scanned miner (checked against `FLEET_SELECTABLE_PAIRING_STATUSES`, so it tracks the canonical rule) and shows a "this miner isn't fully paired yet" callout, mirroring the existing already-in-another-rack block.

I did not add the server-side `LookupMinerByIdentifier` guard: rack membership is persisted by `SaveRack` (not the lookup), no flow enforces pairing status server-side today, and `DEFAULT_PASSWORD` is intentionally treated as "paired-like" elsewhere — so a server invariant here would be a broader change than restoring parity with the existing UI rule. Happy to file a follow-up for server-side `SaveRack` enforcement if we want defense-in-depth across all flows.

> **[LOW] Async QR decode polling can overlap and amplify browser CPU use** … decode calls can overlap [if `detect()` is slow].

Fixed — added an in-flight `detectingRef` guard in `useQrScanner` so a slow `detect()` (the WASM fallback on mobile) can't let 250ms interval ticks stack up.

Added a test asserting a non-`PAIRED` scanned miner blocks assignment; existing scan tests updated to use a `PAIRED` snapshot for the happy path.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -0,0 +1,181 @@
...
+            const value = results[0]?.rawValue;
+            if (value && !detectedRef.current) {
+              detectedRef.current = true;
+              onDetectedRef.current(value);
+            }
+          } catch {
+            // Transient decode failures (e.g. a blurry frame) are expected;
+            // keep polling.
+          }
+        }, SCAN_INTERVAL_MS);
```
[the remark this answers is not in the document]
mcharles-square: 🤖

> The async `setInterval` scan loop can still call `detector.detect()` (and then `onDetected`) after the effect has been cleaned up … because the interval callback doesn't check `cancelled` after the `await`.

Fixed in 824dbf796 — the loop returns early when `cancelled` is set after the awaited `detect()`, so a closed/unmounted modal can no longer fire `onDetected` (and its lookup/state update) post-teardown. The overlap point you also raised is handled by the `detectingRef` in-flight guard added in 34318972e.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -0,0 +1,181 @@
...
+            const results = await detector.detect(el);
+            const value = results[0]?.rawValue;
+            if (value && !detectedRef.current) {
+              detectedRef.current = true;
+              onDetectedRef.current(value);
+            }
+          } catch {
+            // Transient decode failures (e.g. a blurry frame) are expected;
+            // keep polling.
+          }
```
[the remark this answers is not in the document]
mcharles-square: 🤖

> the catch block swallows every failure while leaving `status` as `"scanning"` … users see an endless live viewfinder with no way to recover except closing the modal. Please transition to an error state after [persistent failures].

Fixed in 824dbf796 — a blurry frame resolves empty, but a genuine decoder failure (e.g. the self-hosted WASM asset can't load on-prem) throws. The loop now counts consecutive throws and, after ~4s, stops and sets the error state; `ScanMinerQrModalView` renders the photo-capture fallback on `cameraError`, so the user can recover without closing the modal.

client/package.json
```diff
@@ -33,6 +33,7 @@
     "@connectrpc/connect-web": "2.1.2",
     "@dnd-kit/core": "6.3.1",
     "@dnd-kit/sortable": "10.0.0",
+    "barcode-detector": "^3.2.0",
```
flesher: should we explicitly pin this dep?

server/internal/handlers/fleetmanagement/handler.go
```diff
@@ -76,6 +76,18 @@ func (h *Handler) ExportMinerListCsv(ctx context.Context, r *connect.Request[pb.
 	})
 }
 
+func (h *Handler) LookupMinerByIdentifier(ctx context.Context, r *connect.Request[pb.LookupMinerByIdentifierRequest]) (*connect.Response[pb.LookupMinerByIdentifierResponse], error) {
+	if _, err := middleware.RequirePermission(ctx, authz.PermMinerRead, authz.ResourceContext{}); err != nil {
```
flesher, line 80: With an empty resource context this lookup is only org scoped.  A user could technically scan a miner from a site they dont have access to and assign it another rack.

flesher (approved):

client/package.json
mcharles-square: > should we explicitly pin this dep?

Yes — done in ef6fd24ce. Pinned `barcode-detector` to exact `3.2.0` (the already-resolved version, so no dependency-tree change) to match the exact-pin convention every other client dependency uses.

server/internal/handlers/fleetmanagement/handler.go
mcharles-square: > With an empty resource context this lookup is only org scoped. A user could technically scan a miner from a site they dont have access to and assign it another rack.

Good catch. You're right the lookup is org-scoped (`ResourceContext{}`). Flagging that this matches the current posture across rack/miner management — `SaveRack` (the actual assignment), `CreateDeviceSet`/`UpdateDeviceSet`, and the other miner-read handlers all use an empty `ResourceContext` too. So closing this properly means site-scoping the assignment path (at minimum the lookup **and** `SaveRack`), not just this endpoint — otherwise the same cross-site assignment is still reachable through the list/search flow. Leaving this thread open: it needs a call on whether we introduce site-scoped enforcement for rack management now (and how broadly), since it's a shared authz change rather than something isolated to the QR-scan feature.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  // Keep the latest callback without retriggering the start/stop effect.
+  useEffect(() => {
+    onDetectedRef.current = onDetected;
+  }, [onDetected]);
+
+  const getDetector = useCallback((): BarcodeDetector => {
+    if (!detectorRef.current) {
+      initBarcodeScanner();
+      detectorRef.current = new BarcodeDetector({ formats: ["qr_code"] });
```
b-rowan, line 82: Should this support barcodes, not just QR codes?  Bitmain for example has their miner SN on a 1D barcode rather than a 2D QR.

EG:

Image

image

mcharles-square: @b-rowan thanks for these. We definitely want to support as many of these format as possible. I'll look to add barcode support in a follow up PR

server/internal/handlers/fleetmanagement/handler.go
mcharles-square: Filed #674 to track site-scoped RBAC enforcement for rack/miner management. Since the gap is the existing org-scoped posture across `SaveRack` and the sibling handlers (not something the QR lookup introduces), merging #654 as-is and handling the scoping holistically in #674. Resolving here — follow-up lives in that issue.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -0,0 +1,209 @@
...
+          await video.play().catch(() => {
+            // Autoplay can reject if the element isn't visible yet; the
+            // interval loop below still reads frames once it plays.
+          });
+        }
+
+        const detector = getDetector();
+        setStatus("scanning");
+
+        intervalRef.current = setInterval(async () => {
```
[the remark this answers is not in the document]
mcharles-square: > startup resumes here and installs a new scanner interval despite `cancelled` being true. That interval [is never cleared].

Fixed in 14a7091dc — re-check `cancelled` after the awaited `play()` and bail before installing the interval, so a teardown during startup can no longer leave an orphaned scan loop. All three await points in the start path (`getUserMedia`, `play()`, and the in-loop `detect()`) now guard on `cancelled`.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
mcharles-square: 🤖 Filed #675 to track 1D barcode support (broaden the detector formats + scan-modal copy; the serial resolves through the existing lookup path). Keeping it out of this PR as discussed.

PR #678, opened 2026-07-07
feat(rack): scan 1D barcodes (not just QR) for miner slot assignment

### LINKED ISSUE #675: Support 1D barcodes (not just QR) in miner scan for rack assignment
### Summary

The scan-to-assign flow currently decodes **QR codes only**. Some miners print their serial number on a **1D barcode** rather than a 2D QR — e.g. Bitmain — so those labels can't be scanned today.

Raised by @b-rowan in review of #654:

> Should this support barcodes, not just QR codes? Bitmain for example has their miner SN on a 1D barcode rather than a 2D QR.

Deferred to this follow-up (per the PR thread) to keep #654 scoped.

### Where it's constrained

- `client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts` — the detector is built as `new BarcodeDetector({ formats: ["qr_code"] })`; the live-scan loop and `detectFromBlob` (photo fallback) both use it.
- UI copy assumes QR (`ScanMinerQrModalView`: "Scan miner QR code", "Point the camera at the miner's QR label", not-found copy).

### Suggested direction

- Broaden the detector `formats` to include the relevant 1D symbologies (e.g. `code_128`, `code_39`, and whatever Bitmain/other vendors actually use — worth confirming against real labels like the ones in the PR thread). `barcode-detector` (already a dependency) supports these.
- The resolved value flows through `parseScannedIdentifier`, which already infers serial-vs-MAC by shape, so a serial-bearing 1D barcode should resolve via the existing `LookupMinerByIdentifier` path with no server change.
- Update the scan modal copy to say "barcode" (or "QR / barcode") so the affordance matches.
- Watch for false positives / decode cost from enabling more formats; keep the first-hit-wins behavior.

### Context

- Follow-up from #654 (QR-scan rack assignment).

mcharles-square:
## What

Extends the scan-to-assign flow (added in #654) to read **1D barcodes** in addition to QR. Some vendors — notably Bitmain/Antminer — print the miner serial on a 1D barcode rather than a QR, so those labels couldn't be scanned before.

Closes #675 (follow-up raised by @b-rowan on #654).

## Changes

- **`useQrScanner`** — the `BarcodeDetector` now decodes `["qr_code", "data_matrix", "code_128", "code_39", "code_93"]` (was QR-only), for both the live-camera loop and the photo-capture `detectFromBlob`. Antminer serials are Code 128.
- **`useQrScanner`** — request an `ideal` 1920×1080 camera stream (was defaulting to ~640×480). A dense 1D barcode needs enough horizontal pixels across its bars to decode; QR survives a lower-res frame, 1D often doesn't. `ideal` degrades gracefully on lesser cameras.
- **Copy** (`ScanMinerQrModalView`, `RackPane` popover) — made format-agnostic: "Scan miner barcode", "QR code or barcode", "code" instead of "QR code".

No server change: a decoded serial flows through the existing `parseScannedIdentifier` → `LookupMinerByIdentifier` path unchanged, regardless of symbology.

## Testing

- Verified the engine + these formats decode a real Antminer S21 label end-to-end: ran the actual label photo through `zxing-wasm` (the lib `barcode-detector` wraps) with the app's exact format list → `Code128: "DGAHFKABDJGJD01G1"`. Also round-tripped the serial through Code 128/39/93.
- `ScanMinerQrModal` tests updated for the copy change; 6/6 passing.
- `tsc --noEmit` ✅, `eslint --max-warnings 0` ✅ (format literals validated as `BarcodeFormat`s).

## Notes

- The self-hosted ZXing reader WASM (`initBarcodeScanner`) already supports all these symbologies, so no asset change was needed.
- Symbology set chosen for alphanumeric equipment serials (Code 128/39/93) + 2D (QR/Data Matrix); easy to extend (e.g. `itf`/`codabar`) if a vendor label needs it.

### DIFF
--- a/client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
+++ b/client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
@@ -29,27 +29,32 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
   // a result). Toggling this tears the stream down between scans.
   const cameraActive = show && liveCamera && phase.kind === "scanning";
 
-  const runLookup = useCallback(async (rawValue: string) => {
-    const { value, type } = parseScannedIdentifier(rawValue);
-    if (!value) {
-      setPhase({ kind: "not-found", identifier: rawValue.trim() });
+  // A single frame/photo can decode more than one barcode (e.g. a serial plus a
+  // model/asset code), and the detector's ordering isn't guaranteed — so try
+  // each decoded value against the lookup and only report not-found once none
+  // resolve, rather than committing to the first.
+  const runLookup = useCallback(async (rawValues: string[]) => {
+    const seq = ++lookupSeq.current;
+    const candidates = rawValues.map((raw) => ({ raw, ...parseScannedIdentifier(raw) })).filter((c) => c.value);
+    if (candidates.length === 0) {
+      setPhase({ kind: "not-found", identifier: rawValues[0]?.trim() ?? "" });
       return;
     }
-    const seq = ++lookupSeq.current;
-    setPhase({ kind: "looking-up", identifier: value });
-    const result = await lookupMinerByIdentifier(value, type);
-    if (seq !== lookupSeq.current) return; // superseded
-    switch (result.status) {
-      case "found":
+    setPhase({ kind: "looking-up", identifier: candidates[0].value });
+    let lastError: string | null = null;
+    for (const { value, type } of candidates) {
+      const result = await lookupMinerByIdentifier(value, type);
+      if (seq !== lookupSeq.current) return; // superseded by a newer scan / close
+      if (result.status === "found") {
         setPhase({ kind: "found", snapshot: result.snapshot });
-        break;
-      case "notFound":
-        setPhase({ kind: "not-found", identifier: value });
-        break;
-      case "error":
-        setPhase({ kind: "error", message: result.message });
-        break;
+        return;
+      }
+      if (result.status === "error") lastError = result.message;
+      // notFound → keep trying the remaining decoded values
     }
+    setPhase(
+      lastError ? { kind: "error", message: lastError } : { kind: "not-found", identifier: candidates[0].value },
+    );
   }, []);
 
   const { videoRef, status, errorMessage, detectFromBlob } = useQrScanner({
@@ -76,9 +81,9 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
       if (!file) return;
       setPhase({ kind: "looking-up", identifier: "" });
       try {
-        const rawValue = await detectFromBlob(file);
-        if (rawValue) {
-          await runLookup(rawValue);
+        const rawValues = await detectFromBlob(file);
+        if (rawValues.length) {
+          await runLookup(rawValues);
         } else {
           setPhase({ kind: "not-found", identifier: "" });
         }
--- a/client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
+++ b/client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
@@ -1,6 +1,7 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 
 import ScanMinerQrModalView, { type ScanPhase } from "./ScanMinerQrModalView";
+import { MinerIdentifierType } from "@/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb";
 import { lookupMinerByIdentifier } from "@/protoFleet/api/lookupMinerByIdentifier";
 import { canUseLiveCamera, useQrScanner } from "@/protoFleet/features/fleetManagement/hooks/useQrScanner";
 import { parseScannedIdentifier } from "@/protoFleet/features/fleetManagement/utils/parseScannedIdentifier";
@@ -24,6 +25,9 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
   const fileInputRef = useRef<HTMLInputElement | null>(null);
   // Guards the async lookup against a modal close / rescan mid-flight.
   const lookupSeq = useRef(0);
+  // Aborts in-flight lookups (a multi-candidate loop issues several) on rescan
+  // or unmount, so a mid-scan dismiss doesn't keep hitting the server.
+  const abortRef = useRef<AbortController | null>(null);
 
   // Only run the live camera while we're actively scanning (not while showing
   // a result). Toggling this tears the stream down between scans.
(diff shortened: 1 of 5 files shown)


client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -122,7 +128,11 @@ export function useQrScanner({ onDetected, active }: UseQrScannerOptions): UseQr
     const start = async () => {
       try {
         const stream = await navigator.mediaDevices.getUserMedia({
-          video: { facingMode: { ideal: "environment" } },
+          // Request a high resolution: a dense 1D barcode (e.g. an Antminer
+          // serial in Code 128) needs enough horizontal pixels across its bars
+          // to decode, where a QR would survive a lower-res frame. `ideal` so
+          // devices without a 1080p camera gracefully fall back.
+          video: { facingMode: { ideal: "environment" }, width: { ideal: 1920 }, height: { ideal: 1080 } },
```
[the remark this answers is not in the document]
mcharles-square: > a 1D barcode in the hidden side margins can be decoded and assigned while the operator is aiming at a different visible label. Please either show the full stream aspect ratio or crop the frame passed to the detector.

Fixed in afca24dd0 — cropping the frame rather than changing the preview. Each frame is now drawn to its visible center square (matching the aspect-square/object-cover crop) before `detect()`, so only what the operator sees is decoded; barcodes in the hidden side margins are no longer read. Kept the square preview UI unchanged.

client/src/protoFleet/features/fleetManagement/hooks/useQrScanner.ts
```diff
@@ -79,7 +82,13 @@ export function useQrScanner({ onDetected, active }: UseQrScannerOptions): UseQr
   const getDetector = useCallback((): BarcodeDetector => {
     if (!detectorRef.current) {
       initBarcodeScanner();
-      detectorRef.current = new BarcodeDetector({ formats: ["qr_code"] });
+      // QR is the common case, but some vendors (e.g. Bitmain) print the serial
+      // on a 1D barcode, so accept the alphanumeric linear symbologies used on
+      // equipment labels alongside the 2D formats. The decoded value flows
+      // through parseScannedIdentifier the same way regardless of symbology.
+      detectorRef.current = new BarcodeDetector({
+        formats: ["qr_code", "data_matrix", "code_128", "code_39", "code_93"],
```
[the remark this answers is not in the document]
mcharles-square: > both scan paths still inspect only `results[0]` ... please try each decoded `rawValue` before showing the not-found state

Fixed in 94c9a19a0. `useQrScanner` now surfaces *every* decoded value (live loop + `detectFromBlob` return `string[]`), and the container tries each candidate against the lookup, only showing not-found once none resolve — so a model/asset code sorting ahead of the serial no longer discards the matching code. Added a test covering a multi-barcode frame where the first value misses and the serial resolves.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
```diff
@@ -29,27 +29,32 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
...
+      if (result.status === "found") {
         setPhase({ kind: "found", snapshot: result.snapshot });
-        break;
-      case "notFound":
-        setPhase({ kind: "not-found", identifier: value });
-        break;
-      case "error":
-        setPhase({ kind: "error", message: result.message });
-        break;
+        return;
```
rongxin-liu: Following up on the earlier Codex security note about multi-barcode frames. Trying every candidate is a big improvement over stopping at `results[0]`, but the loop still commits to the *first* candidate the server resolves, and the photo path decodes the full frame (the new center-square crop only guards the live-camera path).

Three hardenings might be worth folding in here:

1. Maybe de-dupe `candidates` by `value` before the loop — the same label can decode more than once in a single frame/photo.
2. Perhaps try explicitly typed candidates (`SN:`/`MAC:` prefixed, i.e. `type !== UNSPECIFIED`) ahead of unspecified ones, so a stray model/asset code that happens to resolve can't out-race the intended serial.
3. Should we collect *all* `found` results and prompt the operator to choose (or rescan) when more than one distinct miner resolves, rather than silently assigning whichever value the detector happened to list first? A photo of a dense rack row with two resolvable labels currently picks one nondeterministically, and the confirmation screen doesn't signal that the match was ambiguous.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
```diff
@@ -29,27 +29,32 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
...
+      }
+      if (result.status === "error") lastError = result.message;
+      // notFound → keep trying the remaining decoded values
     }
+    setPhase(
+      lastError ? { kind: "error", message: lastError } : { kind: "not-found", identifier: candidates[0].value },
+    );
```
rongxin-liu: Two smaller thoughts on the failure path:

- When a lookup returns `status: "error"` from a transport/server failure, the remaining candidates are almost certainly doomed too, so it could be worth breaking out early (or running the lookups concurrently, first-found-wins) so an operator with an unreachable server isn't waiting through N sequential failed round-trips before seeing any feedback.
- The final not-found message always names `candidates[0].value`, which may not be the value the operator actually aimed at when several codes parsed. Perhaps prefer a typed (`SN:`/`MAC:`) candidate as the headline identifier, or word the callout to cover all attempted values.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
```diff
@@ -29,27 +29,32 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
...
-    const seq = ++lookupSeq.current;
-    setPhase({ kind: "looking-up", identifier: value });
-    const result = await lookupMinerByIdentifier(value, type);
-    if (seq !== lookupSeq.current) return; // superseded
-    switch (result.status) {
-      case "found":
+    setPhase({ kind: "looking-up", identifier: candidates[0].value });
+    let lastError: string | null = null;
+    for (const { value, type } of candidates) {
+      const result = await lookupMinerByIdentifier(value, type);
```
rongxin-liu: `lookupMinerByIdentifier` already accepts an `AbortSignal`, but it isn't passed here. Since `ManageRackModal` unmounts this component on dismiss and `lookupSeq` is only bumped when the modal (re)opens, dismissing mid-scan lets this loop keep issuing up to N lookups after the modal is gone (pre-PR it was a single stray call at most). Maybe wire an `AbortController` per `runLookup` invocation, aborted from an unmount cleanup effect and from `rescan()` in addition to the seq bump, so the loop stops at the next await.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
```diff
@@ -76,9 +81,9 @@ export default function ScanMinerQrModal({ show, currentRackLabel, onDismiss, on
       if (!file) return;
       setPhase({ kind: "looking-up", identifier: "" });
       try {
-        const rawValue = await detectFromBlob(file);
-        if (rawValue) {
-          await runLookup(rawValue);
+        const rawValues = await detectFromBlob(file);
+        if (rawValues.length) {
```
rongxin-liu, line 116: Tiny copy nit just below this hunk (the `catch` at L91, untouched by this diff): the message still reads "Try again with **the QR code** centered" but for this PR I think we want something more generic like  "Try again with **the code** centered."

rongxin-liu (approved):

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
mcharles-square: > de-dupe `candidates` by `value` ... try explicitly typed candidates ... ahead of unspecified ones ... collect *all* `found` results and prompt the operator to choose

(1) and (2) done in fad10603d — candidates are de-duped by value, and explicitly-typed (`SN:`/`MAC:`) values are tried before unspecified ones so a stray model/asset code can't out-race the intended serial. (3): I'd rather keep the ambiguous-multi-match chooser as a separate follow-up than fold it into this PR — it needs a new phase + selection UI, and with de-dupe + typed-priority the common wrong-pick paths are already closed. Leaving this thread open pending that call.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
mcharles-square: > break out early (or running the lookups concurrently, first-found-wins) ... The final not-found message always names `candidates[0].value`

Done in fad10603d — a transport/server `error` now breaks the loop immediately instead of running N sequential failed round-trips. And since typed candidates are sorted first, `candidates[0].value` (the not-found headline) is the typed `SN:`/`MAC:` value whenever one was scanned.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
mcharles-square: > `lookupMinerByIdentifier` already accepts an `AbortSignal`, but it isn't passed here ... wire an `AbortController` per `runLookup` invocation, aborted from an unmount cleanup effect and from `rescan()`

Done in fad10603d — each `runLookup` uses a fresh `AbortController`, passes its signal to `lookupMinerByIdentifier`, and aborts on `rescan()` and on unmount (and aborts the prior scan's controller when a new one starts), so a mid-scan dismiss stops the loop at the next await.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
mcharles-square: > the message still reads "Try again with **the QR code** centered" but for this PR I think we want something more generic like "Try again with **the code** centered."

Fixed in fad10603d.

client/src/protoFleet/features/fleetManagement/components/ManageRackModal/ScanMinerQrModal.tsx
mcharles-square: I think this is fine for now. sn/mac should always be 1:1 mapping. and givent the form factor of rigs I dont see the dense rack being an issue.

PR #706, opened 2026-07-08
fix(ci): require releases and -rc. tags to be on main (allow test builds off branches)

mcharles-square:
## Problem

RC tags could be cut from non-mainline commits without the release workflow complaining. This happened in practice: `v0.2.9-rc.2` was tagged on an unmerged host-monitoring feature branch, 8 commits ahead of `main`, so the versioned RC line silently diverged from mainline.

## Root cause

The release workflow's on-main ancestry guard was gated behind a condition that only matched full releases:

```yaml
- name: Require full releases to be on main
  if: ${{ !contains(github.ref_name, '-') }}
```

Every prerelease tag contains a `-` (e.g. `v0.2.9-rc.2`), so `!contains(..., '-')` is false and the check was skipped for all of them — including versioned release candidates.

## Fix

**1. Gate the ancestry check by tag name**, not by whether it's a prerelease:

```yaml
if: \${{ !contains(github.ref_name, '-') || contains(github.ref_name, '-rc.') }}
```

- **Full releases** (`vX.Y.Z`) and **release candidates** (`vX.Y.Z-rc.N`) must be cut from `main`.
- **All other prerelease suffixes** (`-alpha.N`, `-beta.N`, `-<label>`) stay unrestricted, so the team can still cut throwaway test/branch builds off feature branches. They just can't use the `-rc.` name, which is reserved for mainline release candidates.

**2. Reserve `-rc.` naming to the canonical form.** The on-main gate keys on the literal `-rc.`, so a malformed RC (`v1.2.3-rc1`, `v1.2.3-RC.1`) would miss the gate and be treated as an unrestricted test build. The format-validation step now requires any tag whose prerelease starts with `rc` (any case) to be exactly `vX.Y.Z-rc.N`:

| Tag | Result |
|-----|--------|
| `v1.2.3`, `v1.2.3-rc.1`, `v1.2.3-rc.24` | accepted |
| `v1.2.3-rc1`, `v1.2.3-RC.1`, `v1.2.3-rc.1.2`, `v1.2.3-rc-1` | rejected (malformed RC) |
| `v1.2.3-alpha.4`, `v1.2.3-beta-2`, `v1.2.3-search.1` | accepted (not an RC) |

Together these keep the intended workflow — test releases off branches are fine — while preventing versioned RCs from being conflated with, or cut from, non-mainline work.

## Caveat (not addressed here)

This gate runs on tag *push*, so it fails the release **build** rather than preventing the tag from being created. It stops artifacts/releases from being produced off a non-main commit, but a stray tag would still need manual cleanup. Truly blocking the push would require tag protection rules / a server-side hook, which is out of scope.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,20 +17,14 @@ jobs:
         env:
           TAG_NAME: ${{ github.ref_name }}
         run: |
-          # General shape: vX.Y.Z with an optional prerelease suffix
-          # (e.g. v1.2.3, v1.2.3-rc.1, v1.2.3-alpha.4, v1.2.3-<label>.1).
+          # vX.Y.Z with an optional prerelease suffix (e.g. v1.2.3-alpha.4).
           if [[ ! "$TAG_NAME" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9._-]+)?$ ]]; then
             echo "::error::Tag '$TAG_NAME' does not match expected format (e.g., v1.2.3, v1.2.3-rc.1, or v1.2.3-<label>.1)"
             exit 1
           fi
-          # Reserve release-candidate naming: any tag that looks like an RC must be
-          # exactly vX.Y.Z-rc.N. This covers both a leading "rc" token (a typo like
-          # v1.2.3-rc1 or v1.2.3-RC.1) and an "-rc." appearing anywhere in the tag
-          # (e.g. v1.2.3-hostmon-rc.1). The on-main gate below keys on the literal
-          # "-rc." (matched case-insensitively by contains()), so both layers must
-          # agree on what counts as an RC — otherwise such a tag would pass here and
-          # then fail confusingly at the gate. Rejecting now keeps "-rc." fully
-          # reserved for canonical RCs.
+          # Reserve RC naming: any RC-looking tag must be canonical vX.Y.Z-rc.N.
+          # Matches the on-main gate below (which keys on "-rc." anywhere,
+          # case-insensitively) so a near-miss can't slip past it as a test build.
           if [[ "$TAG_NAME" == *-* ]]; then
             prerelease="${TAG_NAME#*-}"
             first_token="${prerelease%%[.-]*}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -17,23 +17,44 @@ jobs:
...
-      - name: Require full releases to be on main
-        if: ${{ !contains(github.ref_name, '-') }}
+      # Versioned releases (vX.Y.Z) and release candidates (vX.Y.Z-rc.N) must be
+      # cut from main so the release line can't silently diverge. Other
+      # prerelease suffixes (e.g. -alpha.N, -beta.N, -<label>) are intentionally
+      # left unrestricted so the team can cut throwaway test builds off feature
+      # branches — just not under the -rc. name, which is reserved for mainline
+      # release candidates.
+      - name: Require releases and release candidates to be on main
+        if: ${{ !contains(github.ref_name, '-') || contains(github.ref_name, '-rc.') }}
```
rongxin-liu, line 44: This gate matches `-rc.` anywhere in the tag (and `contains()` is case-insensitive), but the reservation check above only looks at the first prerelease token. So a test build like `v1.2.3-hostmon-rc.1` passes validation, then fails here — with an error suggesting the suffix form it already used. Fails closed, but confusing.

Maybe make the two layers agree, e.g. have the reservation check also reject any non-canonical tag containing `-rc.`:

```bash
tag_lc=$(printf '%s' "$TAG_NAME" | tr 'A-Z' 'a-z')
if { [[ "$first_token_lc" == rc* ]] || [[ "$tag_lc" == *-rc.* ]]; } && [[ ! "$TAG_NAME" =~ ... ]]; then
```

(Or, if mid-label `-rc.` should stay allowed for test builds, gate on a step output computed in the validate step instead.)

.github/workflows/release.yml
```diff
@@ -17,23 +17,44 @@ jobs:
...
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           AHEAD_BY=$(gh api "repos/${{ github.repository }}/compare/main...${{ github.sha }}" --jq '.ahead_by')
           if [[ "$AHEAD_BY" -ne 0 ]]; then
```
rongxin-liu: `-ne` coerces a non-numeric `ahead_by` (e.g. `null`) to 0, silently passing the gate. Since this is the enforcement point, worth failing closed:

```suggestion
          if [[ "$AHEAD_BY" != "0" ]]; then
```

rongxin-liu (approved):

.github/workflows/release.yml
mcharles-square: Good catch. The reservation check now also rejects any non-canonical tag containing `-rc.` (matched case-insensitively via `tag_lc`), so it agrees with the gate — `v1.2.3-hostmon-rc.1` is now rejected at validation instead of passing through and failing confusingly at the gate. `-rc.` is fully reserved for canonical `vX.Y.Z-rc.N`.

.github/workflows/release.yml
mcharles-square: Done — switched to `[[ "$AHEAD_BY" != "0" ]]` so a non-numeric/null `ahead_by` fails closed instead of coercing to 0. Also surfaced the value in the error message for debuggability.

ankitgoswami (approved):

PR #717, opened 2026-07-08
fix(server): allow recreating soft-deleted users and pools

### LINKED ISSUE #423: Cannot recreate a user with the same username after deleting an existing user
### Describe the bug

If a user is soft deleted, it still prevents the creation of a new user with an identical username

### Preconditions

A user is soft deleted

### Steps to reproduce

1. Go to settings - team
2. Soft delete a user
3. Try creating a new user with the same username
4. Fails

### Expected behavior

It should allow reusing usernames

### Proto Fleet version

v0.2.8-rc.4

### Environment

_No response_

### Logs

```shell

```

_No response_

### LINKED ISSUE #696: [Bug] Deleted mining pools still block recreating the same URL and username
### Describe the bug

After deleting a mining pool, Proto Fleet still appears to treat it as a duplicate when trying to create a new pool with the same values.

The deleted pool no longer appears in the current list, but creating a new pool with the same URL and username is blocked as if the old pool still exists. This appears to be a regression.

Previously "released" related Linear ticket: `DASH-1186`, therefore this is a regression.

### Steps to reproduce

1. Navigate to `Settings -> Pools`
2. Click `Add pool`
3. Enter a title, URL, and username
4. Save the new pool
5. Open the three-dots menu for that pool row
6. Click `Delete pool`
7. Click `Add pool` again
8. Enter the same title, URL, and username as before
9. Click `Save`

### Expected behavior

The pool should be saved, because there is no duplicate in the current list of pools.

### Actual behavior

The pool cannot be saved. The message shown to the user is vague, and network behavior suggests the pool is not being treated consistently as a duplicate.

### Proto Fleet version

`nightly-20260703-a7b00e867a12-13-g04a0c8c5-dirty`

### Environment

- OS: macOS
- Browser: Chrome

ankitgoswami:
Reviewable diff: +78/-1 across 9 files (excludes generated, test, and story files).

## Summary
Deleted users and deleted mining pools no longer reserve their old unique keys. Operators can recreate a pool with the same URL and username after deletion, and admins can recreate a username after deactivating the prior user, while live duplicates remain rejected.

The migration is split so every `CREATE UNIQUE INDEX CONCURRENTLY` and `DROP INDEX CONCURRENTLY` operation is the only statement in its migration file. That matches the existing golang-migrate/Postgres pattern, documents the dirty-migration recovery path for each concurrent build, and avoids dirty migrations from concurrent index builds being submitted in a multi-command query.

Fixes #696.
Fixes #423.

## How it works
The database invariant moves from full-table unique constraints to partial unique indexes covering only rows where `deleted_at IS NULL`. New inserts still collide with other live rows, so active usernames and active pool keys remain unique. Once a row is soft-deleted, it leaves the live uniqueness set and the same key can be inserted again as a new row.

The migration runs in four ordered steps. First, it builds the live-row `user.username` index concurrently as a single-statement migration, then swaps the old user constraint for that index name in a regular transactional migration. It repeats the same pattern for `pool(org_id, url, username)`: build the live-row index concurrently, then drop the old constraint and rename the new index. Down migrations reverse the sequence, with rollback expected to fail if historical reused keys now exist.

Pool creation now wraps database insert failures with `%w`, so callers and tests can inspect the underlying Postgres error through `errors.As` while preserving the existing error message text. The regression test asserts both duplicate paths fail on the expected unique constraint/index names before proving soft-deleted rows can be replaced.

## Diagrams
```mermaid
flowchart TD
    A["Create live user or pool"] --> B["Live-row unique index enforces key"]
    B --> C["Soft delete sets deleted_at"]
    C --> D["Row leaves live uniqueness set"]
    D --> E["Create replacement with same key"]
    E --> F["Replacement becomes the only live row for that key"]
```

```mermaid
flowchart TD
    U1["000116: create user live index concurrently"] --> U2["000117: drop old user constraint and rename index"]
    U2 --> P1["000118: create pool live index concurrently"]
    P1 --> P2["000119: drop old pool constraint and rename index"]
```

## Areas of the code involved
| --- | --- | --- |
| `server/migrations/000116_scope_soft_delete_uniqueness.*.sql` | Builds/drops the user live-row unique index concurrently as a single-statement migration, with recovery notes for dirty/INVALID-index deploy failures. | Keeps concurrent index operations outside multi-statement migration execution and documents operator recovery. |
| `server/migrations/000117_swap_user_soft_delete_uniqueness.*.sql` | Swaps the original `user.username` constraint for the partial live-row index name, and restores it on rollback. | This is where user uniqueness semantics change from all rows to live rows. |
| `server/migrations/000118_create_live_pool_key_index.*.sql` | Builds/drops the pool live-row unique index concurrently as a single-statement migration, with recovery notes for dirty/INVALID-index deploy failures. | Keeps the pool concurrent index operation isolated for safe migration execution and documents operator recovery. |
| `server/migrations/000119_swap_pool_soft_delete_uniqueness.*.sql` | Swaps the original pool key constraint for the partial live-row index name, and restores it on rollback. | This is where pool key uniqueness semantics change from all rows to live rows. |
| `server/internal/domain/stores/sqlstores/pool.go` | Wraps create-pool insert failures with `%w`. | Preserves access to the underlying Postgres error for constraint-aware callers/tests without changing the user-facing error text. |
| `server/internal/domain/stores/sqlstores/soft_delete_uniqueness_integration_test.go` | Adds DB-backed regression coverage for live duplicate rejection and post-delete reuse, sharing one migrated test database across the focused pool and user subtests. | Proves both flows match the intended soft-delete semantics and pins the expected unique constraint/index names. |

- Use partial unique indexes instead of application-side duplicate filtering so all insert paths share one database invariant.
- Split concurrent index DDL into single-statement migration files instead of combining them with constraint swaps, matching existing migration safety patterns.
- Preserve the old index/constraint names for the live-row indexes so unique-violation metadata remains stable for callers.
- Wrap pool creation insert failures with `%w` instead of string-matching the error so constraint assertions use typed Postgres errors.
- Keep sqlc queries unchanged because inserts, soft deletes, and list filters already express the correct live-row semantics once the database invariant changes.

## Testing & validation
- `just lint`
- `DB_PASSWORD=fleet go test ./internal/domain/stores/sqlstores -run 'TestSQLStores_AllowKeyReuseAfterSoftDelete' -count=1` from `server/`
- Verified `000116` and `000118` concurrent migration up/down files each contain exactly one SQL statement.
- `git diff --check`

### DIFF
--- a/server/internal/domain/stores/sqlstores/soft_delete_uniqueness_integration_test.go
+++ b/server/internal/domain/stores/sqlstores/soft_delete_uniqueness_integration_test.go
@@ -5,56 +5,68 @@ import (
 
 	poolspb "github.com/block/proto-fleet/server/generated/grpc/pools/v1"
 	"github.com/block/proto-fleet/server/internal/domain/stores/sqlstores"
+	"github.com/block/proto-fleet/server/internal/infrastructure/encrypt"
 	"github.com/block/proto-fleet/server/internal/testutil"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
-func TestSQLPoolStore_AllowsPoolKeyReuseAfterSoftDelete(t *testing.T) {
+func TestSQLStores_AllowKeyReuseAfterSoftDelete(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping database integration test in short mode")
 	}
 
 	ctx := t.Context()
-	_, poolStore := setupTransactorTest(t)
-	config := &poolspb.PoolConfig{
-		PoolName: "Reusable Pool",
-		Url:      "stratum+tcp://reuse.pool.example.com:3333",
-		Username: "wallet",
-		Password: wrapperspb.String("secret"),
-	}
+	poolStore, userStore := setupSoftDeleteUniquenessStores(t)
 
-	firstID, err := poolStore.CreatePool(ctx, config, 1)
-	require.NoError(t, err)
+	t.Run("pool key", func(t *testing.T) {
+		config := &poolspb.PoolConfig{
+			PoolName: "Reusable Pool",
+			Url:      "stratum+tcp://reuse.pool.example.com:3333",
+			Username: "wallet",
+			Password: wrapperspb.String("secret"),
+		}
 
-	_, err = poolStore.CreatePool(ctx, config, 1)
-	require.Error(t, err, "duplicate live pool keys must still be rejected")
+		firstID, err := poolStore.CreatePool(ctx, config, 1)
+		require.NoError(t, err)
 
-	require.NoError(t, poolStore.SoftDeletePool(ctx, 1, firstID))
+		_, err = poolStore.CreatePool(ctx, config, 1)
+		require.Error(t, err, "duplicate live pool keys must still be rejected")
 
-	secondID, err := poolStore.CreatePool(ctx, config, 1)
-	require.NoError(t, err)
-	require.NotEqual(t, firstID, secondID)
+		require.NoError(t, poolStore.SoftDeletePool(ctx, 1, firstID))
+
+		secondID, err := poolStore.CreatePool(ctx, config, 1)
+		require.NoError(t, err)
+		require.NotEqual(t, firstID, secondID)
+	})
+
+	t.Run("username", func(t *testing.T) {
+		firstID, err := userStore.CreateUser(ctx, "external-user-1", "reuse@example.com", "hash", false)
+		require.NoError(t, err)
+
+		_, err = userStore.CreateUser(ctx, "external-user-2", "reuse@example.com", "hash", false)
+		require.Error(t, err, "duplicate live usernames must still be rejected")
+
+		require.NoError(t, userStore.SoftDeleteUser(ctx, firstID))
+
+		secondID, err := userStore.CreateUser(ctx, "external-user-2", "reuse@example.com", "hash", false)
+		require.NoError(t, err)
+		require.NotEqual(t, firstID, secondID)
+	})
 }
 
-func TestSQLUserStore_AllowsUsernameReuseAfterSoftDelete(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping database integration test in short mode")
-	}
+func setupSoftDeleteUniquenessStores(t *testing.T) (*sqlstores.SQLPoolStore, *sqlstores.SQLUserStore) {
+	t.Helper()
 
-	ctx := t.Context()
 	db := testutil.GetTestDB(t)
-	userStore := sqlstores.NewSQLUserStore(db)
-
-	firstID, err := userStore.CreateUser(ctx, "external-user-1", "reuse@example.com", "hash", false)
+	config, err := testutil.GetTestConfig()
 	require.NoError(t, err)
 
-	_, err = userStore.CreateUser(ctx, "external-user-2", "reuse@example.com", "hash", false)
-	require.Error(t, err, "duplicate live usernames must still be rejected")
-
-	require.NoError(t, userStore.SoftDeleteUser(ctx, firstID))
-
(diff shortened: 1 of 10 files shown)


server/migrations/000116_scope_soft_delete_uniqueness.up.sql
```diff
@@ -0,0 +1,8 @@
+-- Scope legacy uniqueness checks to live rows so soft-deleted pools and users
+-- do not reserve their old keys forever. This first migration builds the user
+-- replacement index before 000117 swaps out the old full-table constraint.
+--
+-- CONCURRENTLY must be the sole statement and cannot run in a transaction.
+CREATE UNIQUE INDEX CONCURRENTLY uq_user_username_live
```
rongxin-liu, line 13: Worth adding the operator-recovery runbook comment that `000074_user_organization_role_single_live_org_scope.up.sql` documents for this exact pattern? If the concurrent build fails mid-deploy, `schema_migrations` is left `dirty=true` (possibly with an INVALID index), and since migrations run at startup under `restart: always`, the container crash-loops without self-healing.

And then maybe copy the recovery paragraph from 000074, drop the INVALID index, then `migrate force 115`, so on-call doesn't have to rediscover the procedure mid-incident.

server/migrations/000118_create_live_pool_key_index.up.sql
```diff
@@ -0,0 +1,7 @@
+-- Build the live-row pool key replacement index before 000119 swaps out the
+-- old full-table constraint.
+--
+-- CONCURRENTLY must be the sole statement and cannot run in a transaction.
+CREATE UNIQUE INDEX CONCURRENTLY uk_pool_org_url_username_live
```
rongxin-liu, line 12: Same suggestion as 000116, perhaps add the recovery runbook comment here too (drop the INVALID index, then `migrate force 117`).

server/internal/domain/stores/sqlstores/soft_delete_uniqueness_integration_test.go
```diff
@@ -0,0 +1,72 @@
...
+			Url:      "stratum+tcp://reuse.pool.example.com:3333",
+			Username: "wallet",
+			Password: wrapperspb.String("secret"),
+		}
+
+		firstID, err := poolStore.CreatePool(ctx, config, 1)
+		require.NoError(t, err)
+
+		_, err = poolStore.CreatePool(ctx, config, 1)
+		require.Error(t, err, "duplicate live pool keys must still be rejected")
```
rongxin-liu: Both duplicate-rejection subtests assert only `require.Error`, so an unrelated failure (FK violation, encryption error) would pass the test that guards this PR's core invariant.

Should we assert it's actually a unique violation?

- **username subtest**: `CreateUser` returns the raw error, so this works today:
  ```go
  var pgErr *pgconn.PgError
  require.ErrorAs(t, err, &pgErr)
  require.Equal(t, pgerrcode.UniqueViolation, pgErr.Code)
  require.Equal(t, "uq_user_username", pgErr.ConstraintName)
  ```
- **pool subtest**: `CreatePool` wraps with `%v` via `fleeterror.NewInternalErrorf` (`pool.go:80`), so the `PgError` is unreachable via `errors.As`. Should we change that `%v` to `%w`, or — if that's out of scope — at minimum assert the message contains `uk_pool_org_url_username`?

This would also pin the constraint-name stability that the 000117/000119 rename comments promise to callers.

rongxin-liu (approved):

server/migrations/000118_create_live_pool_key_index.up.sql
ankitgoswami: 🤖
> Same suggestion as 000116, perhaps add the recovery runbook comment here too (drop the INVALID index, then `migrate force 117`).

Addressed: added the matching recovery runbook comment for dirty=true at version 118, a possible INVALID `uk_pool_org_url_username_live`, and recovery via dropping that index before `migrate force 117`.

PR #719, opened 2026-07-08
fix(ci): prevent stale review policy publishes

ankitgoswami:
Reviewable diff: +56/-8 across 2 files (excludes generated, test, and story files).

## Summary

Review Policy now avoids stale or misleading side effects without reopening the stale-green hole. Ignored/no-op events use their own per-run concurrency group, so they cannot cancel or replace queued meaningful evaluations for the same head SHA; real policy evaluations still serialize by head SHA. Cancelled evaluations publish a pending Review Policy status instead of a red fallback failure, non-cancelled evaluation failures still fail closed, PR label sync remains serialized per head SHA with the permission GitHub requires, and the workflow regression tests now run in PR Gate.

## How it works

The workflow distinguishes meaningful policy events from ignored/no-op triggers at the workflow concurrency layer. Meaningful events such as review dismissals, PR edits, workflow completions, and external status/check completions continue to use the PR head SHA as their concurrency group. Ignored events, such as GitHub Actions check events, managed Review Policy status events, rerequested check events, and comment-only review submissions, use `github.run_id` in a per-run group so they cannot replace a pending meaningful evaluation under GitHub's default single-pending-run concurrency behavior.

When an evaluation resolves a PR but is cancelled, the status publisher still runs behind the stale-head and newer-run guards and writes `Review Policy` as `pending`. That blocks a future required status check without showing a false policy failure. Normal evaluator failures keep the existing fail-closed defaults: `needs-human-review`, `passed=false`, and `enforced=true`.

The AI classifier step has its own timeout, so a hung classifier becomes a normal `continue-on-error` failure that flows through the existing fail-closed fallback instead of cancelling the whole evaluate job. Label sync still skips cancelled evaluations, serializes by head SHA, and keeps exactly one managed Review Policy label.

## Diagrams

```mermaid
flowchart TD
  A["Review Policy event"] --> B{"Ignored/no-op event"}
  B --> C["Use per-run concurrency group"]
  B --> D["Use head-SHA concurrency group"]
  C --> E["Skip evaluation without replacing queued policy runs"]
  D --> F["Resolve PR and head SHA"]
  F --> G{"Evaluate cancelled"}
  G --> H["Publish pending Review Policy status"]
  G --> I["Use evaluator outputs or fail-closed defaults"]
  I --> J["Publish success or failure for current head"]
  J --> K["Sync one managed PR label"]
```

## Areas of the code involved

| --- | --- | --- |
| `.github/workflows/review-policy.yml` | Routes ignored/no-op events to a per-run concurrency group, keeps real evaluations serialized by head SHA, bounds the AI classifier step, publishes pending status for cancelled evaluations, and keeps label sync skipped on cancellation. | This is the production workflow behavior that prevents stale green statuses and confusing red failures while preserving fail-closed behavior for real evaluation failures. |
| `.github/workflows/pr-gate.yml` | Adds a Review Policy Tests job and includes it in the Gate dependency list. | Ensures workflow regression guards run in CI instead of only locally. |
| `.github/scripts/evaluate_review_policy_test.py` | Adds structural workflow assertions for PR Gate wiring, ignored-event concurrency behavior, publish/label job conditions, permissions, and classifier timeout. | Test file — protects the YAML guardrails that caused the observed #716/#717 confusion and later stale-status review feedback. |

- Give ignored events unique per-run concurrency groups instead of `queue: max`, preserving stale-run cancellation for real same-head policy evaluations.
- Publish `pending` for cancelled evaluations instead of skipping status publishing entirely, so required checks block without showing a false policy failure.
- Keep label sync skipped for cancelled evaluations, since labels are advisory PR metadata and should only reflect completed decisions.
- Add a step-level classifier timeout so hung AI calls fail closed through the normal fallback path.
- Parse workflow YAML in tests with Ruby stdlib YAML from Python, avoiding a new PyYAML dependency while making assertions structural instead of string-count based.

## Testing & validation

- `python3 .github/scripts/evaluate_review_policy_test.py`
- `ruby -ryaml -e 'data=YAML.load_file(%q(.github/workflows/review-policy.yml)); puts data.fetch(%q(concurrency)).fetch(%q(group)); puts data.fetch(%q(concurrency)).fetch(%q(cancel-in-progress))'`
- `ruby -ryaml -e 'ARGV.each { |path| YAML.load_file(path); puts "parsed #{path}" }' .github/workflows/review-policy.yml .github/workflows/pr-gate.yml`
- `git diff --check`

Not covered: the updated PR Gate job and Review Policy workflow still need to complete in GitHub Actions on this pushed commit.

---

GPT-5

### DIFF
--- a/.github/scripts/evaluate_review_policy_test.py
+++ b/.github/scripts/evaluate_review_policy_test.py
@@ -11,29 +11,24 @@ import evaluate_review_policy as policy
 
 
 class ReviewPolicyTest(unittest.TestCase):
-    def test_workflow_publishes_only_completed_policy_decisions(self):
+    def test_workflow_skips_cancelled_publish_side_effects(self):
         workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
             encoding="utf-8"
         )
 
-        self.assertIn(
-            "publishable: ${{ steps.evaluate_policy.outputs.publishable || "
-            "steps.fallback_policy.outputs.publishable || 'false' }}",
-            workflow,
-        )
-        self.assertEqual(workflow.count("needs.evaluate.outputs.publishable == 'true'"), 2)
         self.assertEqual(workflow.count("needs.evaluate.result != 'cancelled'"), 2)
+        self.assertNotIn("needs.evaluate.outputs.publishable", workflow)
+        self.assertNotIn("publishable:", workflow)
+        self.assertNotIn('echo "publishable=true"', workflow)
 
-    def test_workflow_publishes_unsupported_base_decisions(self):
+    def test_workflow_keeps_fail_closed_non_cancelled_failures(self):
         workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
             encoding="utf-8"
         )
 
-        self.assertIn('echo "unsupported_base=true" >> "$GITHUB_OUTPUT"', workflow)
-        self.assertIn("id: fallback_policy", workflow)
-        self.assertIn('if [ "$UNSUPPORTED_BASE" = "true" ]; then', workflow)
-        self.assertIn("enforced=true", workflow)
-        self.assertIn('echo "enforced=${enforced}" >> "$GITHUB_OUTPUT"', workflow)
+        self.assertIn('exit 1', workflow)
+        self.assertIn("got base ref ${BASE_REF}.", workflow)
+        self.assertIn("ENFORCED: ${{ needs.evaluate.outputs.enforced || 'true' }}", workflow)
 
     def test_workflow_serializes_and_authorizes_label_sync(self):
         workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
--- a/.github/scripts/evaluate_review_policy_test.py
+++ b/.github/scripts/evaluate_review_policy_test.py
@@ -18,11 +18,23 @@ class ReviewPolicyTest(unittest.TestCase):
 
         self.assertIn(
             "publishable: ${{ steps.evaluate_policy.outputs.publishable || "
+            "steps.unsupported_base_policy.outputs.publishable || "
             "steps.bootstrap_policy.outputs.publishable || 'false' }}",
             workflow,
         )
         self.assertEqual(workflow.count("needs.evaluate.outputs.publishable == 'true'"), 2)
         self.assertEqual(workflow.count("needs.evaluate.result != 'cancelled'"), 2)
+
+    def test_workflow_publishes_unsupported_base_decisions(self):
+        workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
+            encoding="utf-8"
+        )
+
+        self.assertIn('echo "unsupported_base=true" >> "$GITHUB_OUTPUT"', workflow)
+        self.assertIn("id: unsupported_base_policy", workflow)
+        self.assertIn("steps.unsupported_base_policy.outputs.enforced", workflow)
+        self.assertIn('echo "enforced=true" >> "$GITHUB_OUTPUT"', workflow)
+        self.assertIn("steps.policy_source.outputs.unsupported_base != 'true'", workflow)
 
     def test_workflow_serializes_and_authorizes_label_sync(self):
         workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
--- a/.github/scripts/evaluate_review_policy_test.py
+++ b/.github/scripts/evaluate_review_policy_test.py
@@ -49,8 +49,11 @@ class ReviewPolicyTest(unittest.TestCase):
 
     def test_workflow_ignored_events_do_not_cancel_active_evaluations(self):
         workflow = load_workflow("review-policy.yml")
+        group_expression = workflow["concurrency"]["group"]
         cancel_expression = workflow["concurrency"]["cancel-in-progress"]
 
+        self.assertIn("format('ignored-{0}', github.run_id)", group_expression)
+        self.assertIn("github.event.pull_request.head.sha || github.event.workflow_run.head_sha", group_expression)
         self.assertIn("github.event.check_run.app.slug == 'github-actions'", cancel_expression)
         self.assertIn("github.event.action == 'rerequested'", cancel_expression)
         self.assertIn("github.event.check_suite.app.slug == 'github-actions'", cancel_expression)
--- a/.github/scripts/evaluate_review_policy_test.py
+++ b/.github/scripts/evaluate_review_policy_test.py
@@ -18,8 +18,7 @@ class ReviewPolicyTest(unittest.TestCase):
 
         self.assertIn(
             "publishable: ${{ steps.evaluate_policy.outputs.publishable || "
-            "steps.unsupported_base_policy.outputs.publishable || "
-            "steps.bootstrap_policy.outputs.publishable || 'false' }}",
+            "steps.fallback_policy.outputs.publishable || 'false' }}",
             workflow,
         )
         self.assertEqual(workflow.count("needs.evaluate.outputs.publishable == 'true'"), 2)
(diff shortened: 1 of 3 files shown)


.github/workflows/review-policy.yml
```diff
@@ -419,7 +422,11 @@ jobs:
   publish-status:
     name: Publish Review Policy Status
     needs: evaluate
-    if: always() && needs.evaluate.outputs.found == 'true'
+    if: >-
+      always() &&
+      needs.evaluate.outputs.found == 'true' &&
+      needs.evaluate.outputs.publishable == 'true' &&
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > retargeting the PR away from the default branch makes `Select policy source` exit before either `bootstrap_policy` or `evaluate_policy` sets `publishable`

Fixed in f738ade8be by making unsupported-base retargets an explicit publishable fail-closed decision. Cancelled/interrupted evaluations still leave `publishable=false`, but non-default-base PRs now publish enforced `needs-human-review` so they overwrite any stale green `Review Policy` status on the same head SHA.

.github/workflows/review-policy.yml
```diff
@@ -419,7 +447,11 @@ jobs:
...
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > non-cancelled evaluation failures no longer fail closed

Fixed in dcffe40911 by removing the `publishable` gate entirely. The publisher and label sync now skip only when the evaluate job was cancelled, so checkout/fetch/preflight/policy-source failures still publish the existing fail-closed `needs-human-review` default instead of leaving a stale green status behind.

.github/workflows/review-policy.yml
```diff
@@ -419,7 +419,10 @@ jobs:
   publish-status:
     name: Publish Review Policy Status
     needs: evaluate
-    if: always() && needs.evaluate.outputs.found == 'true'
+    if: >-
+      always() &&
+      needs.evaluate.outputs.found == 'true' &&
+      needs.evaluate.result != 'cancelled'
```
rongxin-liu: This removes the fail-closed backstop for a real case: the concurrency group keys only on head SHA, so routine same-SHA events (e.g. `check_run` completions that never run `evaluate`) can cancel a downgrade re-evaluation with no successor publish, leaving a stale green status, and cancelled runs never retry.

Should we publish `state: pending` here instead of skipping entirely (still behind the existing newer-run guard)? That blocks merge without the misleading red.

.github/workflows/review-policy.yml
```diff
@@ -506,12 +509,18 @@ jobs:
   sync-label:
     name: Sync Label
     needs: evaluate
-    if: always() && needs.evaluate.outputs.found == 'true'
+    if: >-
+      always() &&
+      needs.evaluate.outputs.found == 'true' &&
+      needs.evaluate.result != 'cancelled'
```
rongxin-liu, line 542: GitHub also reports a job exceeding `timeout-minutes` as `cancelled`, so a hung AI-classifier call (the one unbounded step in `evaluate`) now publishes nothing where it previously failed closed. Maybe worth a step-level `timeout-minutes` on that step. It already has `continue-on-error: true` and a fail-closed fallback.

.github/scripts/evaluate_review_policy_test.py
```diff
@@ -11,6 +11,38 @@
 
 
 class ReviewPolicyTest(unittest.TestCase):
+    def test_workflow_skips_cancelled_publish_side_effects(self):
```
rongxin-liu: These tests don't seem to run anywhere in CI, worth wiring them into pr-gate so the regression guard actually fires?

.github/scripts/evaluate_review_policy_test.py
```diff
@@ -11,6 +11,38 @@
...
+        workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
+            encoding="utf-8"
+        )
+
+        self.assertEqual(workflow.count("needs.evaluate.result != 'cancelled'"), 2)
```
rongxin-liu: A whole-file `count(...) == 2` can't tell whether each job actually carries the guard. Perhaps `yaml.safe_load` the workflow and assert `jobs['publish-status']['if']` and `jobs['sync-label']['if']` individually? That would also make the exact-indentation permissions assertion below reformat-proof.

.github/scripts/evaluate_review_policy_test.py
```diff
@@ -11,6 +11,38 @@
...
+        self.assertNotIn('echo "publishable=true"', workflow)
+
+    def test_workflow_keeps_fail_closed_non_cancelled_failures(self):
+        workflow = (Path(__file__).resolve().parents[1] / "workflows" / "review-policy.yml").read_text(
+            encoding="utf-8"
+        )
+
+        self.assertIn('exit 1', workflow)
+        self.assertIn("got base ref ${BASE_REF}.", workflow)
+        self.assertIn("ENFORCED: ${{ needs.evaluate.outputs.enforced || 'true' }}", workflow)
```
rongxin-liu: All three strings exist verbatim in the pre-PR workflow, so this test still passes if the new guard is reverted. Maybe assert the actual `if:` expressions on both jobs instead (`always()` + `found == 'true'` + `result != 'cancelled'`)?

rongxin-liu (approved):

.github/workflows/review-policy.yml
ankitgoswami: 🤖
> Should we publish `state: pending` here instead of skipping entirely (still behind the existing newer-run guard)?

Fixed in e8efc5de5c. Cancelled evaluations that have resolved a PR now publish `Review Policy` as `pending` behind the same stale-head and newer-run guards. I also changed workflow-level concurrency so ignored/no-op same-SHA events no longer cancel an active evaluation in the first place.

.github/workflows/review-policy.yml
ankitgoswami: 🤖
> Maybe worth a step-level `timeout-minutes` on that step.

Fixed in e8efc5de5c. The AI classifier step now has its own 10-minute timeout while keeping `continue-on-error: true`, so a hung classifier flows through the existing fail-closed classifier fallback instead of cancelling the whole evaluate job.

.github/scripts/evaluate_review_policy_test.py
ankitgoswami: 🤖
> These tests don't seem to run anywhere in CI, worth wiring them into pr-gate so the regression guard actually fires?

Fixed in e8efc5de5c. PR Gate now has a `Review Policy Tests` job that runs `python3 .github/scripts/evaluate_review_policy_test.py`, and the Gate job depends on it.

.github/scripts/evaluate_review_policy_test.py
ankitgoswami: 🤖
> A whole-file `count(...) == 2` can't tell whether each job actually carries the guard.

Fixed differently in e8efc5de5c. The tests now parse the workflow structurally using Ruby's stdlib YAML from Python, so they assert `publish-status` and `sync-label` fields directly, including label permissions, without adding a PyYAML dependency.

.github/scripts/evaluate_review_policy_test.py
ankitgoswami: 🤖
> Maybe assert the actual `if:` expressions on both jobs instead (`always()` + `found == 'true'` + `result != 'cancelled'`)?

Fixed differently in e8efc5de5c. The tests now assert the actual intended job conditions: `publish-status` runs for resolved PRs and separately verifies cancelled evaluations publish `pending`, while `sync-label` keeps the `needs.evaluate.result != 'cancelled'` guard.

.github/workflows/review-policy.yml
```diff
@@ -16,7 +16,17 @@ on: # zizmor: ignore[dangerous-triggers] review policy runs from trusted workflo
 
 concurrency:
   group: review-policy-${{ github.event.pull_request.head.sha || github.event.workflow_run.head_sha || github.event.check_run.head_sha || github.event.check_suite.head_sha || github.event.sha || github.sha || github.run_id }}
-  cancel-in-progress: true
+  cancel-in-progress: >-
+    ${{
+      !(
+        (github.event_name == 'check_run' && github.event.check_run.app.slug == 'github-actions') ||
+        (github.event_name == 'check_run' && github.event.action == 'rerequested') ||
```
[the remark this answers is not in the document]
ankitgoswami: 🤖
> a subsequent ignored GitHub Actions check/status/comment event for the same head can replace that pending meaningful evaluation

Fixed in 5adc10b4fd. Ignored/no-op events now use a per-run concurrency group via `format('ignored-{0}', github.run_id)`, while meaningful policy evaluations still use the resolved head SHA group. That prevents skipped ignored runs from replacing queued review-dismissal/workflow_run evaluations without giving up stale-run cancellation for real evaluations.

PR #726, opened 2026-07-09
fix(ui): allow scrolling back through Server Logs and Miner Logs

illegalprime:
## Problem

Logs in the Server Logs UI had no scrollback — you could only ever see the tail of the log. The shared `Logs` component (originally built for Miner Logs) rendered the log list in a fixed-height container with `overflow-y-hidden`, which `scrollIntoView` can still scroll programmatically (so auto-follow worked) but blocks user wheel/trackpad scrolling entirely.

On top of that, the pin-to-bottom detection listened only to `window` scroll. That matched the Miner Logs layout (protoOS scrolls the document), but the fleet app scrolls an inner `overflow-auto` container, so on Server Logs the component always believed it was pinned to the bottom.

## Fix

`Logs` now takes an explicit `scrollMode` prop declared by the call site:

- `"document"` (default, Miner Logs): the pane grows with its content and the page scrolls, as before.
- `"container"` (Server Logs): the pane is a fixed-height box that scrolls itself (`overflow-y-auto`).

Scroll listening and pin-to-bottom detection follow the declared mode — the pane in `"container"` mode, the document in `"document"` mode.

## Verification

- `tsc`, `eslint`, and the existing vitest suite are clean.
- Drove both pages in the local dev app with Playwright (API mocked): initial load pins to the bottom; wheel-up scrolls back to the first log line; new poll entries do not yank the view down while scrolled up; auto-follow resumes after returning to the bottom. Verified on both `/settings/server-logs` and `/miners/:id/logs`, re-run after the `scrollMode` refactor.

### DIFF
--- a/client/src/shared/components/Logs/Logs.tsx
+++ b/client/src/shared/components/Logs/Logs.tsx
@@ -46,6 +46,12 @@ interface LogsProps {
   logsData?: LogsData;
   fetchMaxLogs: () => Promise<LogsData | undefined>;
   downloadFilename?: string;
+  /**
+   * "document": the pane grows with its content and the page scrolls (Miner Logs).
+   * "container": the pane is a fixed-height box that scrolls itself, for layouts
+   * where the page itself never scrolls (Server Logs).
+   */
+  scrollMode?: "document" | "container";
 }
 
 const LEVEL_TO_LOG_TYPE: Record<StructuredLogEntry["level"], logType> = {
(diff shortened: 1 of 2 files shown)


client/src/shared/components/Logs/Logs.tsx
```diff
@@ -246,23 +247,35 @@ const Logs = ({ logsData, fetchMaxLogs, downloadFilename = "miner-logs" }: LogsP
   }, [searchValue, filterByLogType]);
 
   const handleScroll = useCallback(() => {
-    const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
-    const scrollHeight = document.documentElement.scrollHeight;
-    const clientHeight = document.documentElement.clientHeight;
-    const distanceFromBottom = scrollHeight - scrollTop - clientHeight;
+    // The log pane scrolls itself when a fixed-height layout constrains it
+    // (Server Logs); otherwise the document scrolls (Miner Logs).
```
ankitgoswami: maybe the call site should declare a mode instead of using a layout side-effect?
So something like:
```
<Logs
  ...
  scrollMode="scroll"
/>
```

illegalprime: much better yeah

ankitgoswami (approved):

Issue #727, opened 2026-07-09
Bulk action menu gets cut off on shorter viewports


flesher:
## Problem

The bulk actions popover (`BulkActionsPopover`) gets clipped / cut off on shorter viewports. The popover opens **upward** from the bottom-anchored action bar, so on short screens the top of the menu runs off-screen and its items become unreachable.

### Where

- `client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx`
  - Default `position = positions["top left"]`, `offset={20}`, with a phone/tablet `yOffset` of `-32`.
- `client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsWidget.tsx`
  - Renders the trigger + inline popover (`setPopoverRenderMode("inline")`).
- Anchored by the fixed bottom action bar: `client/src/protoFleet/features/fleetManagement/components/MinerList/MinerListActionBar.tsx` (`fixed ... bottom-4 ...`).

### Expected

The popover should stay fully within the viewport on short screens — either flip/clamp its position so no items are clipped, cap its height with internal scrolling, or otherwise ensure every action remains reachable.

### Repro

1. Open the fleet / miner list.
2. Shrink the browser window to a short height (or use a short-height device).
3. Multi-select miners and open the bulk actions ("More"/"Actions") menu.
4. Observe the top of the menu is cut off.

---

## Related: consider adding "Sleep" to the action bar quick actions

While in this area, we should potentially promote a **Sleep** action into the action bar quick actions.

- Quick actions are defined by `quickActionOrder` in `client/src/protoFleet/features/fleetManagement/components/MinerList/MinerActionsMenu.tsx` and currently are: **Blink LEDs**, **Reboot**, **Manage power**.
- "Sleep" today is implemented as `deviceActions.shutdown` (user-facing copy already uses sleep language — "Put ... to sleep" / "Sleeping"), and it currently only lives in the "More" popover, not the quick-action row.
- Proposal: evaluate surfacing Sleep (and possibly its `wakeUp` counterpart) as a quick action so it's a one-tap operation like Reboot.

flesher:
🤖 Resolved by #741 (+#751) (bulk & row action menus kept within the viewport). Closing as completed.

PR #741, opened 2026-07-13
fix(ui): keep bulk & row action menus within the viewport (#727)

flesher:
Reviewable diff: +80/-9 across 5 files (excludes generated, test, and story files).

## Summary

The bulk-actions menu (opened from the fleet/miner list's bottom action bar) was clipped on short viewports — it opens upward, so its top ran off-screen, and it also rendered *underneath* the app's page header. The single-miner row menu ("⋯") could likewise overflow the viewport with no way to reach clipped items. This PR keeps both menus fully within the viewport and reachable, and reshuffles the bulk quick-action row.

Behavior after this change:
- **Bulk menu** renders above the page header and grows upward only until 8px below the top of the viewport, then scrolls internally.
- **Row menu** keeps its "open below the trigger, flip above when it doesn't fit" behavior, but when the list is taller than the screen it pins 8px off both the top and bottom edges and scrolls internally.
- **Quick actions** in the bulk bar are now Blink LEDs · Reboot · Sleep · Wake; Manage power moves into the "More" overflow menu.

## How it works

**Escaping the header (stacking context).** The action bar renders inside the app shell's scroll container, which is `position: fixed` with `z-20`. A fixed element always creates a stacking context, so nothing inside it — regardless of its own z-index — can paint above the shell header at `z-40`. The bulk menu, expanding upward into the header band, was therefore trapped beneath it. The fix renders `MinerListActionBar` through a React portal into `document.body`, lifting it out of that trapped layer; its `z-50` now sits above the header. Because no layout ancestor uses `transform`/`filter`, the fixed bar's on-screen position is identical before and after portaling — only its stacking layer changes. The menu remains a DOM descendant of its trigger, so click-outside handling and the height variable below still work.

**Capping the bulk menu height.** The bulk popover renders inline and opens upward from the trigger. Its final on-screen top edge is `trigger.top − popoverHeight − 20px` (a 12px positioning offset plus an 8px residual from the open animation). To make that land 8px below the viewport top, the component measures the trigger's position and publishes a max-height as a CSS variable on the trigger container; the inline popover inherits it and scrolls internally past that height. The value recomputes on viewport resize. Phones are unaffected — they use a full-height bottom sheet.

**Capping the row menu height.** The row menu already renders in "portal-fixed" mode, where the shared positioning hook flips the menu above the trigger when it won't fit below and clamps it inside the viewport. The missing piece was a height cap, so a taller-than-screen menu overflowed with no scroll. It now caps at `100vh − 16px` (8px top + 8px bottom) with internal scroll. Separately, the shared clamp was pinning the menu's *pre-animation* coordinate at 8px; the slide animation's residual `translateY(-8px)` then ate that margin, leaving it flush to the edge. The clamp now compensates for the animation's residual translate, so a pinned menu keeps a real 8px inset off both edges. This clamp fix applies to all `portal-fixed` popovers (row menus, selects), improving only the near-edge/overflow cases; menus that already fit are unchanged.

## Diagrams

```mermaid
flowchart TD
    Root["AppLayout root (no stacking context)"] --> Header["Shell header — fixed z-40"]
    Root --> Scroll["Scroll container — fixed z-20 (stacking context)"]
    Scroll --> Page["Miner list page"]
    Page -. "before: bar trapped under header" .-> BarOld["Action bar z-20 — clipped by header"]
    Root --> Body["document.body (portal target)"]
    Body --> BarNew["after: Action bar portal z-50 — above header"]
    BarNew --> Menu["Bulk menu — capped 8px below viewport top, scrolls"]
```

```mermaid
flowchart TD
    Open["Row menu opened"] --> Fit{"Fits below trigger?"}
    Fit -- "yes" --> Below["Open below trigger"]
    Fit -- "no" --> Flip["Flip above / pin to viewport"]
    Below --> Tall{"Taller than viewport?"}
    Flip --> Tall
    Tall -- "no" --> Done["Render as-is"]
    Tall -- "yes" --> Cap["Cap at 100vh-16px, scroll internally"]
    Cap --> Clamp["Clamp pins 8px off top & bottom (animation-compensated)"]
```

## Areas of the code involved

| --- | --- | --- |
| `MinerList/MinerListActionBar.tsx` | Portal the action bar to `document.body`; bar `z-20`→`z-50` | Confirms the bar escapes the shell's `z-20` stacking context and that portaling doesn't shift its fixed position |
| `BulkActions/BulkActionsPopover.tsx` | Measure trigger, publish a max-height CSS var on the trigger; cap + internal scroll on the menu | The offset math (12px + 8px animation) that lands the top edge 8px below the viewport top |
| `MinerActionsMenu/MinerActionsMenu.tsx` | Quick-action row: +Sleep, +Wake, −Manage power | Small, surface-level change to which actions are promoted |
| `RowActionsMenu/RowActionsMenu.tsx` | Cap menu at `100vh−16px` with internal scroll | Shared row menu — verify short menus are unaffected |
| `shared/components/Popover/usePopoverPosition.ts` | Compensate the portal-fixed clamp for the animation's residual translate | Shared positioning logic — the highest-blast-radius change; confirm down-opening popovers are untouched |

- **Portal the whole action bar vs. only the popover.** Portaling the bar keeps the menu inline (preserving the trigger-relative positioning and the max-height cascade) and makes its `z-50` effective, rather than re-plumbing the popover into portal-fixed mode. Safe because no layout ancestor establishes a containing block for fixed elements.
- **Dynamic CSS-variable max-height vs. a static rem buffer** for the bulk menu. The measured value lands the top edge at exactly 8px regardless of bar height (e.g. when the quick-action row wraps), which a fixed buffer can't do.
- **Fix the clamp at the source vs. per-caller.** The animation-translate compensation lives in the shared `usePopoverPosition` because that's the root cause; it's narrowly scoped to the `portal-fixed` branch and only alters near-edge/overflow positioning.
- **Reuse `minimalMargin` (8px)** for all insets so menu margins match the rest of the popover system.

## Testing & validation

- Unit tests updated/passing across Popover, RowActionsMenu, BulkActions, MinerListActionBar, and both action menus (57 passing, 1 pre-existing skip). The quick-actions test now asserts Sleep/Wake render and fire and that Manage power is no longer in the quick row.
- `tsc`, ESLint, and Prettier clean on all changed files.
- Verified in isolation that the Tailwind v4 arbitrary values used here (`max-h-[var(--…,80vh)]`, `max-h-[calc(100vh-theme(spacing.4))]`) compile to real CSS.
- **Not yet covered:** live browser verification at a short viewport. Layout-dependent behavior (the exact 8px landing, header layering, internal scroll) can't be exercised in jsdom and should be spot-checked manually.

### DIFF
--- a/client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
+++ b/client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
@@ -16,8 +16,6 @@ const ANIMATION_TRANSLATE = minimalMargin;
 // Total upward shift the popover applies on top of `-popoverHeight`: the layout
 // offset correction (`POPOVER_OFFSET` beyond `minimalMargin`) plus the animation.
 const UPWARD_SHIFT = POPOVER_OFFSET - minimalMargin + ANIMATION_TRANSLATE;
-// Floor so an extremely short viewport still shows a usable, scrollable menu.
-const MIN_MENU_HEIGHT = 88;
 const MENU_MAX_HEIGHT_VAR = "--bulk-actions-menu-max-h";
 
 // The menu is anchored above the bottom-fixed action bar and opens upward, so a
@@ -88,7 +86,11 @@ const BulkActionsPopover = <ActionType,>({
 
     const update = () => {
       const triggerTop = trigger.getBoundingClientRect().top;
-      const available = Math.max(triggerTop - UPWARD_SHIFT + yOffset - minimalMargin, MIN_MENU_HEIGHT);
+      // Cap at exactly the space above the trigger (never floor above it — a floor
+      // taller than the real gap pushes the top off-screen and hides the first
+      // actions). On a very short viewport this shrinks toward 0 and the whole
+      // list scrolls internally, keeping every row reachable (#727).
+      const available = Math.max(triggerTop - UPWARD_SHIFT + yOffset - minimalMargin, 0);
       trigger.style.setProperty(MENU_MAX_HEIGHT_VAR, `${available}px`);
     };
 
--- a/client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
+++ b/client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
@@ -1,9 +1,26 @@
+import { useLayoutEffect } from "react";
 import { BulkAction } from "./types";
 import Divider from "@/shared/components/Divider";
-import Popover, { popoverSizes } from "@/shared/components/Popover";
+import Popover, { popoverSizes, usePopover } from "@/shared/components/Popover";
 import Row from "@/shared/components/Row";
 import { type Position, positions } from "@/shared/constants";
 import { useWindowDimensions } from "@/shared/hooks/useWindowDimensions";
+
+// The menu is anchored above the bottom-fixed action bar and opens upward, so a
+// tall list would otherwise run off the top of the viewport (#727). We cap its
+// height so its top edge stops `minimalMargin` (8px) below the viewport top and
+// the overflow scrolls internally. Derived from the trigger's measured position
+// and the popover's own offset math so it's exact regardless of bar height:
+//   topInViewport = trigger.top - popoverHeight - UPWARD_SHIFT + yOffset
+// setting that to MIN_TOP_MARGIN gives the max height below.
+const MIN_TOP_MARGIN = 8;
+// Total upward shift the popover applies on top of `-popoverHeight`:
+//   • layout: `offset` (20) minus `minimalMargin` (8) = 12, plus
+//   • animation: `slideUpPopover` settles at `translateY(-8px)` (forwards).
+const UPWARD_SHIFT = 12 + 8;
+// Floor so an extremely short viewport still shows a usable, scrollable menu.
+const MIN_MENU_HEIGHT = 88;
+const MENU_MAX_HEIGHT_VAR = "--bulk-actions-menu-max-h";
 
 interface BulkActionsPopoverProps<ActionType> {
   actions: BulkAction<ActionType>[];
@@ -50,13 +67,48 @@ const BulkActionsPopover = <ActionType,>({
   closePopover,
 }: BulkActionsPopoverProps<ActionType>) => {
   const { isPhone, isTablet } = useWindowDimensions();
+  const { triggerRef } = usePopover();
+  const yOffset = isPhone || isTablet ? -32 : 0;
+
+  // Keep the max-height in sync with the trigger's viewport position (phones use a
+  // full-height bottom sheet instead, so skip them). The value is published as a CSS
+  // variable on the trigger container; the inline popover is a DOM descendant, so it
+  // inherits it via `max-h-[var(...)]` below.
+  useLayoutEffect(() => {
+    const trigger = triggerRef.current;
+    // Only the default (upward, bottom-bar-anchored) menu needs the cap; callers
+    // that pass their own className position the menu themselves.
+    if (isPhone || !trigger || className) return;
+
+    const update = () => {
+      const triggerTop = trigger.getBoundingClientRect().top;
+      const available = Math.max(triggerTop - UPWARD_SHIFT + yOffset - MIN_TOP_MARGIN, MIN_MENU_HEIGHT);
+      trigger.style.setProperty(MENU_MAX_HEIGHT_VAR, `${available}px`);
+    };
+
+    update();
+    window.addEventListener("resize", update);
+    window.visualViewport?.addEventListener("resize", update);
+    return () => {
+      window.removeEventListener("resize", update);
+      window.visualViewport?.removeEventListener("resize", update);
+      trigger.style.removeProperty(MENU_MAX_HEIGHT_VAR);
+    };
+  }, [triggerRef, isPhone, yOffset, className]);
+
   const onAction = (action: BulkAction<ActionType>) => {
     beforeEach(action.requiresConfirmation);
     action.actionHandler();
   };
   return (
     <Popover
-      className={className ?? "-mr-3 !space-y-0 !rounded-2xl px-0 pt-2 pb-1 phone:w-[calc(100vw-theme(spacing.4))]"}
+      className={
+        className ??
+        // Cap the menu height (see MENU_MAX_HEIGHT_VAR above) and scroll internally so
+        // it can't run off the top when opening upward from the bottom-anchored action
+        // bar on short viewports (#727).
+        "-mr-3 max-h-[var(--bulk-actions-menu-max-h,80vh)] !space-y-0 overflow-y-auto overscroll-contain !rounded-2xl px-0 pt-2 pb-1 phone:w-[calc(100vw-theme(spacing.4))]"
+      }
       position={position}
       size={popoverSizes.small}
       offset={20}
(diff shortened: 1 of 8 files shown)


client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
```diff
@@ -1,10 +1,27 @@
...
+// height so its top edge stops `minimalMargin` (8px) below the viewport top and
+// the overflow scrolls internally. Derived from the trigger's measured position
+// and the popover's own offset math so it's exact regardless of bar height:
+//   topInViewport = trigger.top - popoverHeight - UPWARD_SHIFT + yOffset
+// setting that to MIN_TOP_MARGIN gives the max height below.
+const MIN_TOP_MARGIN = 8;
+// Total upward shift the popover applies on top of `-popoverHeight`:
+//   • layout: `offset` (20) minus `minimalMargin` (8) = 12, plus
+//   • animation: `slideUpPopover` settles at `translateY(-8px)` (forwards).
+const UPWARD_SHIFT = 12 + 8;
```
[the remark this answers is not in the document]
flesher: 🤖 Good call — done in 165da17c. The cap math now derives from the shared `minimalMargin` plus a single `POPOVER_OFFSET = 20`, and that same constant is passed to the `offset` prop, so the two can no longer drift. The animation term is named `ANIMATION_TRANSLATE = minimalMargin` (the `slideUpPopover` residual translate).

client/src/protoFleet/features/fleetManagement/components/MinerList/MinerListActionBar.tsx
```diff
@@ -85,9 +86,13 @@ const MinerListActionBar = ({
...
 
-  return (
+  // Portal to the body so the bar escapes the app shell's fixed `z-20` scroll
+  // container (a stacking context). Otherwise the bulk-actions menu — which opens
+  // upward into the header band — is trapped below the `z-40` shell header no
+  // matter how high we set the bar's own z-index (#727).
+  return createPortal(
     <ActionBar
-      className="fixed right-0 bottom-4 left-0 z-20 laptop:left-16 desktop:left-50"
+      className="fixed right-0 bottom-4 left-0 z-50 laptop:left-16 desktop:left-50"
```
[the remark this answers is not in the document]
flesher: 🤖 Good catch — fixed in 165da17c. Lowered the portaled bar to `z-[45]`: still above the `z-40` shell header (so the upward-opening bulk menu clears it, the point of #727), but below the `z-50` phone/tablet nav drawer, so the drawer now paints over the bar when open. Verified the 41–49 band was otherwise unused.

client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
```diff
@@ -50,16 +73,51 @@ const BulkActionsPopover = <ActionType,>({
...
+  // inherits it via `max-h-[var(...)]` below.
+  useLayoutEffect(() => {
+    const trigger = triggerRef.current;
+    // Only the default (upward, bottom-bar-anchored) menu needs the cap; callers
+    // that pass their own className position the menu themselves.
+    if (isPhone || !trigger || className) return;
+
+    const update = () => {
+      const triggerTop = trigger.getBoundingClientRect().top;
+      const available = Math.max(triggerTop - UPWARD_SHIFT + yOffset - minimalMargin, MIN_MENU_HEIGHT);
```
rongxin-liu: Maybe we could avoid flooring this cap above the measured space? When the computed space is below 88px, the menu becomes taller than the available area, pushing its top off-screen and leaving the first actions unreachable.

Perhaps clamp the computed value to zero instead (or use a fallback placement) and add a very-short-viewport case.

client/src/protoFleet/features/fleetManagement/components/RowActionsMenu/RowActionsMenu.tsx
```diff
@@ -141,7 +141,10 @@ const RowActionsMenuInner = ({
       ) : null}
       {open && !isPhone ? (
         <Popover
-          className="!space-y-0 !rounded-2xl px-0 pt-2 pb-1"
+          // Cap at viewport height minus 8px top + 8px bottom and scroll internally
+          // so a menu taller than the screen stays fully reachable. Pairs with the
+          // portal-fixed flip/clamp, which pins a capped menu 8px off both edges (#727).
+          className="max-h-[calc(100vh-theme(spacing.4))] !space-y-0 overflow-y-auto overscroll-contain !rounded-2xl px-0 pt-2 pb-1"
```
rongxin-liu: Worth deriving this cap from the same visual viewport used by `usePopoverPosition`? `100vh` can remain larger than `window.visualViewport.height` during pinch zoom or browser-chrome changes, so clamping cannot keep the full scroll container inside the tappable area.

Could perhaps expose a viewport-constrained max height from `Popover` and cover a `visualViewport.resize` case.

rongxin-liu (approved):

client/src/protoFleet/features/fleetManagement/components/BulkActions/BulkActionsPopover.tsx
flesher: 🤖 Good catch — fixed in 862bae61. Dropped the 88px floor; it was flooring *above* the measured gap, so on a very short viewport the menu ended up taller than the space above the trigger and its top ran off-screen behind the header, hiding the first actions. Now it clamps to the exact available space (min 0), so the list always fits and the overflow scrolls internally — every row stays reachable.

client/src/protoFleet/features/fleetManagement/components/RowActionsMenu/RowActionsMenu.tsx
flesher: 🤖 Done in 862bae61. Added an opt-in `constrainHeightToViewport` prop to the shared `Popover` (threaded into `usePopoverPosition`) that caps a portal-fixed popover to the visible viewport, 8px off top and bottom. It reads `visualViewport.height` and recomputes on the `visualViewport.resize` listener the hook already registers, so the cap holds under pinch-zoom and mobile browser-chrome collapse where the old `100vh` className cap would exceed the tappable area. RowActionsMenu now opts in and I removed the `100vh` class. Made it opt-in rather than always-on because `useResponsivePopover` puts every mobile/tablet popover into portal-fixed — a blanket cap would silently clip popovers without internal scroll — so this stays scoped to callers that pair it with `overflow-y-auto`.

client/src/shared/components/Popover/usePopoverPosition.ts
```diff
@@ -292,16 +298,30 @@ const usePopoverPosition = (
...
 
         style = {
           top: `${top}px`,
           left: `${left}px`,
           visibility: "visible",
         };
+        if (constrainHeightToViewport) {
+          // Leave `minimalMargin` above and below; pairs with the top clamp so a
+          // capped menu stays fully on-screen and scrolls its own overflow.
+          style.maxHeight = `${Math.max(visibleViewport.height - 2 * minimalMargin, 0)}px`;
```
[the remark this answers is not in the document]
flesher: 🤖 Good catch, real bug — fixed in d5044529. The cap (`maxHeight`) lands on the positioned portal wrapper, but the `overflow-y-auto` was on the inner `PopoverContent`, which had no height constraint — so a tall menu painted past the capped wrapper and clipped off-screen with no scroll. Made the wrapper the scroller: `constrainHeightToViewport` now adds `overflow-y-auto overscroll-contain` to the wrapper (where the cap lives) and I dropped the misplaced overflow classes from RowActionsMenu. Added a regression test asserting the wrapper is the scroller.

Issue #723, opened 2026-07-09
Infrastructure device integration foundation — facility fans first


rongxin-liu:
## Summary

Foundational work for integrating external infrastructure devices into Proto Fleet: facility-side equipment (fans, and later PDUs, dampers, pumps, ...) that the server can inventory and eventually command, with all protocol knowledge isolated behind a driver adapter registry so the core stays protocol-agnostic. **Facility fans (single fan or fan group behind a PLC/drive) are the first device kind**, and Modbus TCP is the first driver adapter — motivated by curtailment sequencing (fans off after a delay once miners are curtailed; fans on first, then miners restored after a delay), but the foundation is deliberately kind- and protocol-agnostic.

No device behavior activates from this work — the write path lands with the later Modbus I/O and reconciler-sequencing phases. This is inert surface area until then.

Delivered by #724. Full design: `docs/plans/2026-07-09-facility-fan-curtailment-integration-plan.md` (see "Delivery: PR grouping").

## Scope

- [x] Migration pair: `infrastructure_device` table — `site_id` FK, `building_name`, `name`, `device_kind` (`single_fan` | `fan_group`), `fan_count`, `enabled`, `driver_type` (text), `driver_config` (opaque jsonb owned by the adapter), timestamps
- [x] `proto/infrastructure/v1/infrastructure.proto`: `InfrastructureDevice` (opaque `driver_config`, no per-protocol messages; `device_kind`/`driver_type` as strings so new kinds and protocols are never wire breaks) + `InfrastructureService` List/Get/Create/Update/Delete; `just gen`, generated code committed together
- [x] Driver seam: `server/internal/domain/infrastructure/driver/` — `Controller` interface (`SetState(ctx, device, DesiredState)`, `ValidateConfig`, `Capabilities()`; doc comment reserving `ReadStatus` for v2 read-back) and `driver_type` → adapter factory registry
- [x] `driver/modbustcp/` config parsing + `ValidateConfig`: unit ID 1–247, port 1–65535, register address 0–65535, endpoint restricted to private/loopback/link-local (mirrors the MQTT subscriber's `validateBrokerTransport` guard). Protocol I/O is a later phase, not here
- [x] sqlc queries (`server/sqlc/queries/infrastructure_device.sql`), org-scoped
- [x] Domain CRUD service: protocol-blind field validation in core, `driver_config` validation delegated to the registry; transactional writes with a site row-lock (TOCTOU guard vs concurrent `DeleteSite`)
[...]

PR #750, opened 2026-07-15
feat(curtailment): facility fan response profile settings (3/5)

negarn:
Reviewable diff: +1127/-56 across 28 files (excludes generated, test, and story files).

## Summary

Operators can now attach site-scoped facility fan devices and curtail/restore delays to curtailment response profiles using the API-backed infrastructure inventory. Profile saves persist that configuration end to end, while transactional reference guards prevent selected devices from being deleted, moved out of scope, or removed through a site cascade.

**Stack:** [#724](https://github.com/block/proto-fleet/pull/724) (infrastructure backend, merged) → [#743](https://github.com/block/proto-fleet/pull/743) (infrastructure client, merged) → **this PR (profile fan settings, 3/5)** → PR 4 (Modbus write path) → PR 5 (event stamping and reconciler sequencing). This diff is relative to `main`. The merged foundation supplies the site-authorized infrastructure CRUD API and `useInfrastructureDevices`; this PR consumes those contracts without exposing redacted driver configuration. Actual fan commands, concurrent-event fan claims, event state, restore gating, and physical-state monitoring remain out of scope and land in phases 4–5 from the [facility fan curtailment plan](https://github.com/block/proto-fleet/blob/main/docs/plans/2026-07-09-facility-fan-curtailment-integration-plan.md).

https://github.com/user-attachments/assets/268130e4-514e-4ebd-8c6b-2ccce18bc567

## How it works

The response-profile form adds Infrastructure alongside Sites and Miners. Opening it launches a 600px shared modal populated by `useInfrastructureDevices`; the list is filtered to the form's current site scope, grouped by site/building, and keeps disabled devices visible but non-selectable. Operators choose enabled fan devices and configure the delay after miner curtailment before fans turn off plus the delay after fans turn on before miners restore. The footer stays fixed while the device list scrolls, Select all includes enabled devices only, and changing site scope prunes selections that are no longer visible. The live Run curtailment entry point is intentionally gated until its request/event path can persist and execute fan settings in phase 5.

The client maps selected IDs and delay values into the profile create/update Connect requests. The handler translates them into the domain model, where `ResponseProfileService` normalizes duplicate IDs, verifies every device exists in the organization, and rejects any device outside the profile's site/miner scope. Disabled references remain valid for existing profiles and generate a structured warning; the UI prevents new selection while preserving an already-saved disabled selection until the operator clears it. Updates carry an explicit replacement marker, so clients generated before fan settings existed preserve the current values instead of silently clearing them.

The migration stores fan IDs as a `BIGINT[]` plus two non-negative delay columns and adds a GIN index for reference lookups. Profile saves, direct device deletes, device site moves, and site deletion all use the same site-then-device row-lock order. Saves compare locked device sites with the validation snapshot; deletes and moves return FailedPrecondition when a surviving profile still references the device. This closes the check/write race without activating fan commands, which remain in phases 4–5.

## Diagrams

```mermaid
flowchart LR
  Operator["Operator"] --> Form["Response profile form"]
  Form --> Picker["Facility fan selection modal"]
  Picker --> Hook["useInfrastructureDevices (from PR 2)"]
  Hook --> InfraAPI["InfrastructureService inventory"]
  Form --> ProfileAPI["Curtailment response profile Connect API"]
  ProfileAPI --> Validation["Org and profile-scope validation"]
  Validation --> Lock["Lock selected device rows"]
  Lock --> Store["Response profile SQL store"]
  Store --> DB[("facility fan IDs and delays")]
  Mutator["Delete device, move device, or delete site"] --> Lock
  Lock --> Guard["Count surviving profile references"]
  Guard --> Mutator
```

## Areas of the code involved

| --- | --- | --- |
| `proto/curtailment/v1/curtailment.proto` | Adds fan device IDs, two delay fields, and update replacement intent | Public contract, numeric/list limits, and legacy-client preservation |
| `server/migrations/000123_*`, `server/sqlc/queries/` | Persists fan IDs/delays and adds the indexed reference lookup | Schema defaults, constraints, and delete-guard performance |
| `server/internal/domain/curtailment/response_profile.go` | Deduplicates IDs, masks cross-org/missing devices, validates profile scope, warns on disabled references | Main authorization and safety boundary |
| `server/internal/domain/infrastructure/service.go`, `sites/service.go` | Blocks referenced device deletion/moves and referenced site cascades under row locks | Prevents dangling or out-of-scope profile configuration under concurrency |
| `server/internal/domain/stores/`, `server/internal/handlers/curtailment/` | Threads fields through transport/persistence and locks validated devices before profile writes | End-to-end persistence and reference-invariant boundary |
| `client/.../CurtailmentStartModal.tsx`, `FacilityFanSelectionModal.tsx` | Adds Infrastructure to profile Apply to, site-filtered selection, delays, disabled-device handling, fixed footer, and stale-scope pruning | Main operator workflow; live starts stay gated until execution support |
| `client/.../CurtailmentSettingsPage.tsx`, `CurtailmentManagementPanel.tsx` | Supplies live `useInfrastructureDevices` results and retry/loading/error state | Integration with the API-backed inventory from PR 2 |
| `client/src/shared/components/Modal/Modal.tsx` | Adds an outer surface boundary, width override, and fixed-footer slot while retaining the existing sticky header | Keeps footer actions visible and prevents footer clicks from dismissing the modal |
| `server/cmd/fleetcli/cmd_curtailment.go` | Generated CLI flags for the new profile fields | Generated from the updated proto-driven CLI surface |
| `client/src/protoFleet/api/generated/**`, `server/generated/**` | Regenerated protobuf and sqlc bindings | Generated — skip |
| `*.test.*`, `*_test.go`, `*.stories.tsx` | Covers persistence, scope isolation, compatibility, transactional guards, modal behavior, and parent integrations | Test/story files — skim |

- **Validate devices against the effective profile scope, not only the organization** — prevents a site-scoped profile from controlling a fan at a site whose miners remain active.
- **Mask missing and cross-organization devices as the same NotFound result** — avoids exposing OT inventory while still rejecting invalid profile references.
- **Allow persisted disabled references but disable new UI selection** — preserves existing configuration for operator review instead of silently rewriting it; runtime sequencing will skip disabled devices in phase 5.
- **Store IDs as an indexed array plus a shared row-lock protocol** — keeps the bounded profile-owned list while serializing profile saves against device/site mutation; chosen over introducing a join table in this phase.
- **Require explicit replacement intent on profile updates** — older clients preserve fan settings, while current clients can intentionally replace or clear all three fields.
- **Reuse the shared modal with an explicit fixed-footer slot** — keeps established header/dismiss behavior while supporting a long, scrollable selector.
- **Gate the live Run curtailment selector in this phase** — its current Start request cannot carry fan settings, so hiding the control avoids silently discarding operator choices until phase 5 wires event stamping and execution.

## Testing & validation

- `just gen` on the final `main` base; generated protobuf, sqlc, and Fleet CLI output committed.
- Client: scoped ESLint, `tsc --noEmit`, and 78 post-feedback Vitest tests covering profile API mapping, curtailment modal scope pruning, and shared Modal behavior; the broader 163-test focused suite passed before the feedback patch.
- Client: direct production Vite builds for both `protoOS` and `protoFleet` passed.
- Server: `golangci-lint` clean; affected infrastructure, sites, profile, handler, and sqlstore packages compile and their focused unit tests pass. Real-Postgres tests cover fan profile persistence, referenced device moves, and whole-org references rolling back site deletion.
- Not covered: no Modbus write, reconciler, or end-to-end hardware test because runtime fan control is intentionally deferred to phases 4–5.

Refs #723


server/internal/domain/curtailment/response_profile.go
```diff
@@ -128,41 +177,124 @@ func (s *ResponseProfileService) Delete(ctx context.Context, orgID, profileID in
...
+			return models.ResponseProfile{}, nil, fleeterror.NewNotFoundErrorf("site not found: %d", siteID)
 		}
 	}
+	var infrastructureDevices map[int64]models.ResponseProfileInfrastructureDevice
+	profile.FacilityFanDeviceIDs, infrastructureDevices, err = s.validateFacilityFanDevices(
+		ctx,
+		profile.OrgID,
+		profile.FacilityFanDeviceIDs,
+		req.AuthorizedFacilityFanDevices,
+	)
```
rongxin-liu, line 230: Should we validate these devices against the response profile's effective site/miner scope, not only the caller's RBAC scope? As written, a site-scoped profile can reference a fan at an unrelated site but I think we're ok with it for now at least.

client/src/protoFleet/api/useCurtailmentResponseProfiles.ts
```diff
@@ -199,6 +202,9 @@ function mapApiResponseProfile(profile: ApiCurtailmentResponseProfile, siteNameB
     curtailBatchIntervalSec: curtailBatchIntervalInputValue(profile),
     restoreBatchSize: numberToNonNegativeInputValue(profile.restoreBatchSize),
     restoreIntervalSec: numberToNonNegativeInputValue(profile.restoreBatchIntervalSec),
+    facilityFanDeviceIds: profile.facilityFanDeviceIds.map((id) => id.toString()),
+    fanOffDelaySec: numberToNonNegativeInputValue(profile.fanOffDelaySec),
+    fanRestoreDelaySec: numberToNonNegativeInputValue(profile.fanRestoreDelaySec),
```
rongxin-liu, line 207: Perhaps these API-backed fan fields should remain authoritative after the cache merge. Since `cachedFormValues` wins below, a newer server value for the IDs or delays can be masked and then overwritten by the next save.

Maybe restore these three fields from `formValues` after the spread and add a refetch regression test.

server/internal/handlers/curtailment/response_profiles.go
```diff
@@ -129,14 +151,24 @@ func (h *Handler) UpdateCurtailmentResponseProfile(ctx context.Context, req *con
 			profile.ScopeJSON = cloneBytes(existing.ScopeJSON)
 		}
 	}
+	if !req.Msg.GetReplaceFacilityFanSettings() {
+		profile.FacilityFanDeviceIDs = append([]int64(nil), existing.FacilityFanDeviceIDs...)
+		profile.FanOffDelaySec = existing.FanOffDelaySec
+		profile.FanRestoreDelaySec = existing.FanRestoreDelaySec
```
rongxin-liu, line 157: Should we make this preservation atomic? This copies fan settings from a pre-transaction read, so a legacy update can race with a newer replacement and write the stale values back; the SQL predicate currently checks only site/scope. Maybe preserve the columns in SQL when replacement is false, or include the observed fan fields in the optimistic `WHERE` clause.

rongxin-liu (approved):
Looks good overall!

PR #768, opened 2026-07-17
fix(db): require shared transaction helpers

ankitgoswami:
Reviewable diff: +61/-67 across 5 files (excludes generated, test, and story files).

## Summary

Server contributors now get a lint failure when production code opens a `database/sql` transaction directly instead of using the shared sqlc transaction helpers. Alertmanager notification batches and fleet metric rollups now use that boundary, while transaction paths with materially different retry semantics remain explicit, reviewable exceptions.

**Stack:** **[#768](https://github.com/block/proto-fleet/pull/768) this PR** -> [#759](https://github.com/block/proto-fleet/pull/759) `feat(db): support explicit Postgres DSNs`. PR #759 builds on this transaction boundary to add explicit DSN support and HA pool-reset behavior; those HA changes are intentionally out of scope here.

## How it works

The server lint configuration rejects calls whose selector is `Begin` or `BeginTx`. Test files remain exempt so integration fixtures can control transactions directly. The canonical helper and two higher-risk legacy paths use line-level, documented suppressions so every production exception stays visible in review.

Alertmanager notification chunks and each fleet-rollup delete/upsert/progress sequence pass their complete unit of work to `db.WithTransactionNoResult`. The helper binds sqlc queries to one transaction, rolls back on any failure, and replays the callback for retryable PostgreSQL errors. It preserves wrapped cancellation and PostgreSQL causes for `errors.Is`/`errors.As`, and terminal errors report the number of attempts actually made.

Telemetry metric batches keep their existing single-attempt transaction because their caller already falls back to one write per metric after a batch failure. Applying the retrying helper there could turn one transient 500-row failure into 1,503 transaction attempts. The raw paired-array device rename also remains on its legacy path until it can be moved to a separately validated sqlc query.

```mermaid
flowchart LR
    A["Alertmanager webhook"] --> C["Shared transaction retry boundary"]
    B["Fleet rollup routine"] --> C
    C --> D["sqlc queries bound to one transaction"]
    D --> E["Commit the complete unit of work"]
    F["Telemetry metric batch"] --> G["Documented single-attempt exception"]
    G --> H["Existing per-row fallback on batch failure"]
```

## Areas of the code involved

| --- | --- | --- |
| `server/.golangci.yaml` | Enabled `forbidigo` for production code and prohibited direct `Begin`/`BeginTx` calls | Defines the enforcement boundary and test-only exemption |
| `server/internal/domain/stores/sqlstores/notification_history.go` | Routed notification batch persistence through `WithTransactionNoResult` | Keeps all webhook rows atomic under the shared retry policy |
| `server/internal/infrastructure/timescaledb/telemetry_store.go` | Routed fleet rollups through the helper and documented why metric batches retain their transaction path | Applies retries only where they do not multiply an outer fallback |
| `server/internal/infrastructure/db/with_transaction.go` | Documented replay safety, preserved wrapped causes, and corrected attempt reporting | Makes the supported transaction boundary observable and retry-correct |
| `server/internal/domain/stores/sqlstores/device.go` | Retained one line-level exception for the raw paired-array rename query | Keeps the higher-risk sqlc migration explicit without weakening the repository-wide rule |

- Use the existing `forbidigo` linter instead of adding a custom analyzer; selector-name matching is lightweight but may require a justified suppression for an unrelated `Begin` method.
- Exempt test files because database-backed fixtures intentionally open transactions directly; production code remains covered.
- Migrate transaction paths whose complete callbacks are safe to replay; retain telemetry metric batches because their per-row fallback would amplify transient failures.
- Preserve non-retryable `FleetError` values unchanged, while wrapping retryable and terminal causes with `%w` so PostgreSQL and context classification still works.
- Retain the raw-array device rename until its paired `unnest` query can be moved to sqlc and validated separately.

## Testing & validation

- `source ./bin/activate-hermit && just _lint-server`
- `go test ./internal/infrastructure/db`
- `go test ./internal/domain/telemetry`
- `go test -short ./internal/infrastructure/timescaledb`
- `go test -run '^$' ./internal/domain/stores/sqlstores`
- No new tests or dependencies were added; existing unit coverage was used for this localized review fix.
- Database-backed sqlstores integration tests were not run locally; CI remains responsible for that coverage.
- No fault-injection test currently forces a retryable `BeginTx` or `Commit` error.

## Post-deploy monitoring & validation

- During the first 24 hours, watch `alertmanager webhook: failed to persist alert batch` and `fleet metric rollup routine: upsert failed` for a sustained increase.
- Healthy signals are stable webhook 2xx rates and fleet rollup progress continuing to advance.

---

### DIFF
--- a/server/internal/infrastructure/db/with_transaction.go
+++ b/server/internal/infrastructure/db/with_transaction.go
@@ -20,15 +20,17 @@ func WithTransaction[T any](ctx context.Context, db *sql.DB, action func(q *sqlc
 func withTransactionWithRetry[T any](ctx context.Context, db *sql.DB, action func(q *sqlc.Queries) (T, error), config RetryConfig, txOpts *sql.TxOptions) (T, error) {
 	var zero T
 	var lastErr error
+	attempts := 0
 	currentBackoff := config.InitialBackoff
 
 	for attempt := 1; attempt <= config.MaxAttempts; attempt++ {
 		select {
 		case <-ctx.Done():
-			return zero, fleeterror.NewInternalErrorf("context aborted: %v", ctx.Err())
+			return zero, fleeterror.NewInternalErrorf("context aborted: %w", ctx.Err())
 		default:
 		}
 
+		attempts = attempt
 		result, err := executeTransaction(ctx, db, action, txOpts)
 		if err == nil {
 			return result, nil
@@ -47,19 +49,20 @@ func withTransactionWithRetry[T any](ctx context.Context, db *sql.DB, action fun
 
 		select {
 		case <-ctx.Done():
-			return zero, fleeterror.NewInternalErrorf("context aborted: %v", ctx.Err())
+			return zero, fleeterror.NewInternalErrorf("context aborted: %w", ctx.Err())
 		case <-time.After(sleepDuration):
 		}
 
 		currentBackoff = time.Duration(float64(currentBackoff) * config.BackoffMultiplier)
 	}
 
-	// Preserve FleetError if the original error was already a business error
+	// Preserve non-retryable FleetError values so business error codes survive
+	// the transaction boundary unchanged.
 	var fleetErr fleeterror.FleetError
-	if errors.As(lastErr, &fleetErr) {
+	if !IsRetryablePostgresError(lastErr) && errors.As(lastErr, &fleetErr) {
 		return zero, fleetErr
 	}
-	return zero, fleeterror.NewInternalErrorf("transaction failed after %d attempts: %v", config.MaxAttempts, lastErr)
+	return zero, fleeterror.NewInternalErrorf("transaction failed after %d attempts: %w", attempts, lastErr)
 }
 
 func executeTransaction[T any](ctx context.Context, db *sql.DB, action func(q *sqlc.Queries) (T, error), txOpts *sql.TxOptions) (T, error) {
@@ -68,7 +71,7 @@ func executeTransaction[T any](ctx context.Context, db *sql.DB, action func(q *s
 	//nolint:forbidigo // This helper is the canonical boundary for opening SQL transactions.
 	tx, err := db.BeginTx(ctx, txOpts)
 	if err != nil {
-		return zero, fleeterror.NewInternalErrorf("error opening tx: %v", err)
+		return zero, fleeterror.NewInternalErrorf("error opening tx: %w", err)
 	}
 
 	//goland:noinspection GoUnhandledErrorResult
@@ -82,7 +85,7 @@ func executeTransaction[T any](ctx context.Context, db *sql.DB, action func(q *s
 
 	err = tx.Commit()
 	if err != nil {
-		return zero, fleeterror.NewInternalErrorf("error committing tx: %v", err)
+		return zero, fleeterror.NewInternalErrorf("error committing tx: %w", err)
 	}
 
 	return result, nil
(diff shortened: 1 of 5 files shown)


server/internal/infrastructure/timescaledb/telemetry_store.go
```diff
@@ -370,62 +371,48 @@ func (s *TimescaleTelemetryStore) StoreDeviceMetrics(ctx context.Context, data .
...
-		}
-	}()
-
-	qtx := s.queries.WithTx(tx)
-
-	if s.config.AsyncMetricCommit {
-		if err := qtx.DisableSyncCommit(ctx); err != nil {
-			return fmt.Errorf("failed to set async commit: %w", err)
+	return db.WithTransactionNoResult(ctx, s.db, func(q *sqlc.Queries) error {
+		if s.config.AsyncMetricCommit {
```
rongxin-liu: Perhaps we should distinguish an exhausted retryable batch from a row-specific failure before returning here. `metricsWriterRoutine` falls back to one call per metric; with 500 metrics and three helper attempts, a transient database failure can expand one flush to 1,503 transaction attempts. 

Maybe the store could preserve a retry-exhausted marker so the writer skips per-row fallback for retryable, canceled, or deadline errors.

server/internal/infrastructure/db/with_transaction.go
```diff
@@ -10,6 +10,9 @@ import (
 	"github.com/block/proto-fleet/server/internal/domain/fleeterror"
 )
 
+// WithTransaction runs action in a transaction and retries the entire action for
+// retryable PostgreSQL errors. The action must be safe to replay and should not
+// perform side effects outside the transaction.
```
rongxin-liu, line 15: Since this now documents the helper as retrying the whole action, should we preserve the original error with `%w` in the context/terminal wrappers and track the actual attempt count? Today `%v` prevents `errors.Is`/`errors.As`, and a non-retryable failure after one attempt is reported as `MaxAttempts`.

server/internal/infrastructure/db/with_transaction.go
```diff
@@ -62,6 +65,7 @@ func withTransactionWithRetry[T any](ctx context.Context, db *sql.DB, action fun
 func executeTransaction[T any](ctx context.Context, db *sql.DB, action func(q *sqlc.Queries) (T, error), txOpts *sql.TxOptions) (T, error) {
 	var zero T
 
+	//nolint:forbidigo // This helper is the canonical boundary for opening SQL transactions.
 	tx, err := db.BeginTx(ctx, txOpts)
 	if err != nil {
 		return zero, fleeterror.NewInternalErrorf("error opening tx: %v", err)
```
rongxin-liu: Worth wrapping the `BeginTx` and `Commit` errors with `%w` here? Otherwise `IsRetryablePostgresError` cannot see SQLSTATE `40001`/`40P01`, so a commit-time serialization failure probably exits after the first attempt instead of replaying the transaction.

rongxin-liu (approved):

server/internal/infrastructure/db/with_transaction.go
ankitgoswami: 🤖 > Worth wrapping the `BeginTx` and `Commit` errors with `%w` here?

Fixed: both wrappers now use `%w`, so `IsRetryablePostgresError` can reach an underlying `pgconn.PgError` and replay the transaction for retryable begin- or commit-time failures.

PR #770, opened 2026-07-17
feat(curtailment): sequence facility fans with miner restore (5/5)

negarn:
Reviewable diff: +2065/-223 across 36 files (excludes generated, test, and story files).

## Summary

Proto Fleet now coordinates commissioned facility fans with live and automated curtailment: fans turn off only after miners are confirmed curtailed, turn on before miner restoration, and are re-asserted every reconciliation cycle. Operators can run fan-enabled response profiles from the Energy page, inspect persisted fan phase/error state, and receive a critical alert when fan restore remains broken at the fail-open miner restore gate.

**Stack:** [#724](https://github.com/block/proto-fleet/pull/724) (infrastructure backend, 1/5) → [#743](https://github.com/block/proto-fleet/pull/743) (infrastructure client, 2/5) → [#750](https://github.com/block/proto-fleet/pull/750) (response-profile fan settings, 3/5) → [#762](https://github.com/block/proto-fleet/pull/762) (commissioned Modbus writes, 4/5) → **this PR (reconciler sequencing, 5/5)**. This diff is relative to `main`; all four dependencies are merged. Those PRs provide the site-authorized inventory, profile settings, and dual-allowlist Modbus command boundary consumed here. Fan status read-back, speed control, and protocol expansion remain out of scope and are tracked as future work in the [facility-fan integration plan](https://github.com/block/proto-fleet/blob/main/docs/plans/2026-07-09-facility-fan-curtailment-integration-plan.md).

## How it works

A manual or MQTT-automation Start copies the response profile's fan IDs and delays into the curtailment event. Manual requests authorize every fan against its site; the SQL transaction then takes deterministic per-fan claim locks, row-locks the live infrastructure devices, verifies their sites still match the authorization snapshot, rejects overlap with any non-terminal event, and persists aligned fan/site arrays. The persisted site snapshot remains the authorization source for event history even if inventory later changes. Device moves, device deletion, and site deletion use the same row-lock boundary and fail while an active event owns a selected fan.

During active curtailment, the reconciler waits until every potentially powered target has positively confirmed curtailment, then starts the fan-off delay from the latest confirmation before issuing fan OFF; unavailable or terminal bookkeeping alone is not treated as proof that a miner stopped hashing. The first OFF intent is committed before hardware I/O, so a crash or lost post-command write leaves a conservative durable marker that forces airflow recovery before later admission or termination. A targetless closed-loop watcher does not stop fans until it has admitted and confirmed at least one miner, even if the watcher itself has been active longer than the configured delay. If a closed-loop event discovers a new miner or detects drift after fans are already off, it requires a successful fan ON command and durable write before dispatching curtailment; failures retain the target for retry without sending a miner command. Fans remain on until every admitted miner again confirms curtailment, and the later of the newest confirmation or durable airflow-reopen timestamp reapplies the full cooling delay before the next OFF across reconciler restarts. Restore reversal preserves fan-off/error evidence, retries ON while the event is pending, and blocks miner dispatch until airflow reopening succeeds. The protocol-blind fan controller reloads each device and its current site commissioning, rejects any device whose current site differs from the persisted authorization snapshot, treats missing claimed devices as command failures, skips intentionally disabled references with a deduplicated audit event, invokes each enabled device’s registered driver, aggregates scrubbed failures, and re-asserts the requested state on every cycle. Failed command cycles rotate the first attempted device per event and power phase, so an early unreachable endpoint cannot consume every retry budget and starve later fans. New profile and UI selections are capped at eight commissioned fan devices; legacy profiles above that limit remain executable and may be saved unchanged or reduced, but cannot grow or swap devices until brought within the new limit. Each reconciliation pass divides its deadline fairly across events and caps fan work at half each event allocation so slow OT I/O cannot starve later events or the current event’s miner phases. While an event owns a fan, same-site commissioning, enabled-state, and driver-configuration changes are transactionally blocked alongside moves and deletion.

During restore, fan ON is attempted first and its initial timestamp is persisted. Miner batches remain gated until the configured fan restore delay elapses; both fan delays have a one-hour server-side safety ceiling, and restoration then proceeds fail-open even if fan commands still fail. Failed active or pending reopen retries preserve the first-attempt timestamp so Grafana’s critical-alert deadline cannot move forward forever; the first successful ON replaces it so the cooling delay begins from confirmed airflow. Grafana reads the durable per-event error and either restore/reopen gate timestamp, so a critical alert survives terminalization or restore reversal and clears only after a successful re-assertion. Every active/restoring fan command holds the event lifecycle row lock through the hardware call and state write. Administrative termination retries ON for every nonterminal fan-owning pending or restoring event before terminalization, including legacy or uncertain rows without a persisted OFF stamp. Force Release takes the same per-fan claim and device-row locks as Start, then terminalizes, sweeps ownership, disables automation, issues the authoritative ON command, and persists fan state in one transaction; a concurrent Start cannot claim physically-off fans until recovery finishes. If the reconciler wins the terminal transition after Force Release preloads the event, the locked path rejects a newer owner and otherwise recovers the returned terminal row instead of silently skipping ON. This runs for every nonterminal fan-owning event—even if a prior successful physical OFF lost its timestamp write—while lifecycle locking makes an in-flight OFF finish first and rejects a stale OFF before hardware access. Reissuing Force Release on a terminal event is the explicit operator recovery path whenever a failed ON remains, including when the earlier OFF timestamp write was lost; it refuses newer owners or site drift, retries ON, and clears the durable alert after a successful command.

The Energy page now loads infrastructure inventory for live starts, carries fan IDs/delays through request construction, and removes the interim block on fan-enabled response profiles and automations. Event responses expose the stamped fan IDs, delays, timestamps, and last error for operator-facing state.

## Diagrams

```mermaid
flowchart LR
  Operator["Operator or MQTT automation"] --> Start["Start curtailment with fan IDs and delays"]
  Start --> Auth["Site authorization snapshot"]
  Auth --> Claims["Advisory claim locks and device row locks"]
  Claims --> EventDB[("Curtailment event and aligned fan-site snapshot")]
  EventDB --> Reconciler["Curtailment reconciler"]
  Reconciler --> FanController["Protocol-blind facility fan controller"]
  FanController --> Commissioning["Current site commissioning"]
  FanController --> Driver["Registered infrastructure driver"]
  Driver --> Fans["Facility fans"]
  Reconciler --> Miners["Miner curtail and restore batches"]
  EventDB --> Alert["Durable restore failure state and critical alert"]
```

```mermaid
sequenceDiagram
  participant M as Miners
  participant R as Reconciler
  participant F as Facility fans
  participant D as Event store
  participant A as Alerting

  R->>M: Curtail batches
  M-->>R: Confirm curtailed or terminal
  R->>D: Persist first OFF intent
  R->>F: Re-assert OFF after fan-off delay
  R->>D: Persist latest OFF result and error
  R->>F: Re-assert ON before restore
  R->>D: Persist first ON attempt and latest error
  R->>R: Wait fan-restore delay
  alt Fan ON succeeds
    R->>A: Clear durable restore-failure state
  else Fan ON still fails at gate
    R->>A: Keep durable critical alert active
  end
  R->>M: Restore batches fail-open after delay
```

## Areas of the code involved

| --- | --- | --- |
| `proto/curtailment/v1/curtailment.proto` | Adds fan settings to Start and fan phase/error fields to events | Public API contract and additive compatibility |
| `server/migrations/000125_*`, `server/sqlc/queries/curtailment.sql` | Persists fan IDs, aligned site snapshots, delays, timestamps, errors, active-claim index, and locking queries | Claim exclusivity, authorization stability, and migration constraints |
| `server/internal/domain/curtailment/facility_fans.go` | Adds the protocol-blind controller, current commissioning lookup, claimed-device failure handling, and audits | Physical-command boundary and topology-safe error handling |
| `server/internal/domain/curtailment/reconciler/` | Adds OFF/ON ordering, delays, per-cycle re-assertion, targetless watcher gate, and restore alert emission | Core lifecycle and fail-open safety behavior |
| `server/internal/domain/curtailment/service.go`, `automation.go` | Stamps manual/automation events and serializes fan recovery with forced terminal paths | Ensures all entry and recovery paths use the same sequence without stale hardware commands |
| `server/internal/domain/stores/`, `infrastructure/service.go`, `sites/service.go` | Adds per-event fan claims, authorization snapshots, and active-event mutation guards, including site commissioning | Closes Start-vs-move/delete/site-delete/commissioning races |
| `server/internal/handlers/curtailment/` | Authorizes Start fan sites, uses persisted fan-site snapshots for event access, and preserves whole-org requirements when fan contexts are merged | Site-scoped security boundary and historical authorization |
| `server/internal/infrastructure/metrics/`, Grafana alert rules | Records fan restore outcomes and alerts from durable event failure state at the miner gate | Operator-visible unsafe restore signal that survives terminalization |
| `client/src/protoFleet/features/energy/`, curtailment request builders | Enables fan selection for live starts and updates force-release messaging | Main operator workflow |
| `server/generated/**`, `client/**/generated/**` | Regenerated protobuf, sqlc, retrying-query, and TypeScript bindings | Generated - skip |
| `*.test.*`, `*_test.go` | Covers sequencing, alerts, authorization snapshots, claims, mutation guards, UI requests, and recovery paths | Test files - skim |

- **Persist aligned fan and site arrays at Start** instead of resolving historical authorization from mutable inventory; this adds a schema invariant but prevents permission drift after moves or deletion.
- **Serialize claims with deterministic advisory locks plus device row locks** instead of relying on an array-overlap check alone; Start, terminal recovery, and Force Release share this boundary so ownership cannot move while fan hardware state is unsettled.
- **Re-assert fan state every reconciliation cycle** instead of treating the first write as proof of physical state; this recovers from controller/device restarts but intentionally does not provide status read-back.
- **Commit the first OFF intent before hardware I/O** instead of relying on a post-command stamp; uncertain command outcomes conservatively require ON recovery before admission or administrative termination.
- **Restore miners fail-open after the configured fan delay** instead of blocking indefinitely on an unverified fan command; the critical alert and persisted error make the operational risk explicit.
- **Serialize nonterminal fan commands with lifecycle row locks and terminalize before Force Release ON** instead of allowing hardware writes outside the event transition boundary; stale commands cannot override the authoritative recovery, which also shares Start's claim locks and refuses newer owners.
- **Recover the returned terminal event when Force Release loses the terminal-transition race** instead of treating the idempotent release as completed; held fan locks and a newer-claim check keep that recovery authoritative.
- **Cap new selections at eight while grandfathering unchanged or shrinking legacy profiles up to the historical 1,024-device API ceiling** instead of invalidating persisted configuration; fan groups remain the preferred aggregation boundary and legacy sets cannot grow above the new limit.
- **Apply the half-budget deadline only to fan hardware I/O while retaining the parent event context for the transactional state write** instead of letting a controller timeout roll back its own failure evidence; alerting, restore confirmation, and fail-open miner dispatch retain durable progress every tick.
- **Rotate the first fan attempted after every failed command cycle** instead of restarting at the first configured device; every claimed fan eventually receives the front of the bounded retry budget without parallelizing OT writes.
- **Treat missing claimed fans as failures, audit disabled fans as intentional skips, and block command-affecting or commissioning edits while claimed** instead of allowing mutable configuration to strand recovery; miner recovery remains fail-open after the bounded delay, while retries and durable alerts reflect unsafe fan state.
- **Reopen airflow around dynamic closed-loop admission, persist its renewed cooling window, and revalidate fan sites on every command** instead of assuming the initial target set and inventory topology remain stable; this adds one internal timestamp and extra ON writes to prevent hashing without airflow, premature OFF, or cross-site terminal recovery.
- **Claim stable inventory IDs** rather than deduplicating opaque driver configurations; commissioning must avoid duplicate records for the same physical target until a protocol-aware physical identity contract exists.

## Testing & validation

- `just gen` passed after rebasing onto the latest `origin/main`; the worktree remained clean.
- Server: `just lint` (`0 issues`), `go build ./...`, changed-domain/handler short-mode suites, and SQL-store package/integration-test compilation passed.
- Client: full ESLint and Prettier checks passed; 137 focused curtailment Vitest tests passed; TypeScript and the ProtoFleet production Vite build passed.
- Added coverage for lost post-OFF state writes forcing later airflow recovery, missing-stamp administrative termination recovery, failed and successful dynamic-admission/drift fan reopening before dispatch, pending-phase ON retry gating, bounded fan time with same-tick fail-open restore, holding airflow through confirmation and a renewed cooling delay, pending-event administrative recovery, restore-reversal recovery-state preservation, terminal recovery site-drift/device-mutation serialization, claimed-fan commissioning guards, legacy profile preservation, and real-PostgreSQL fan claims, concurrent Starts, Force Release races, and durable alerts. Local PostgreSQL execution was not available because the existing `fleet` user on `127.0.0.1:5432` rejected the configured password; these tests compile and are expected to run in CI.
- Grafana alert YAML parses successfully, and `git diff --check` passed.

Related: #723


server/internal/handlers/curtailment/handler.go
```diff
@@ -123,6 +123,10 @@ func (h *Handler) StartCurtailment(ctx context.Context, req *connect.Request[pb.
 	if err != nil {
 		return nil, err
 	}
+	authorizedFans, err := h.authorizeFacilityFanDevices(ctx, info.OrganizationID, req.Msg.GetFacilityFanDeviceIds())
+	if err != nil {
+		return nil, err
+	}
```
rongxin-liu, line 129: Should we authorize the persisted `plan.ReplayEvent` before returning it below? The request scope and fan IDs are checked here, but an org-scoped idempotency-key hit can return targets and fan fields from a site where the caller is narrowed.

Maybe reuse the event resource-context checks on the replayed row and add a site-A request -> site-B replay test.

server/internal/domain/curtailment/automation.go
```diff
@@ -545,6 +535,9 @@ func startRequestFromAutomationProfile(rule *models.AutomationRule, profile *mod
 		CurtailBatchSize:          cloneInt32Ptr(profile.CurtailBatchSize),
 		CurtailBatchIntervalSec:   profile.CurtailBatchIntervalSec,
 		UseProfileCurtailSettings: true,
+		FacilityFanDeviceIDs:      append([]int64(nil), profile.FacilityFanDeviceIDs...),
+		FanOffDelaySec:            profile.FanOffDelaySec,
+		FanRestoreDelaySec:        profile.FanRestoreDelaySec,
```
rongxin-liu, line 594: Now that fan-backed automation is enabled here, could we also reserve and verify the predictable `curtailment_automation_rule:<id>` idempotency namespace? A public Start can preseed that key, and the automation path may replay and bind the unrelated event without checking attribution/profile/fans. Perhaps reject reserved keys for non-automation actors and validate the replayed event atomically before binding it.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -758,11 +838,163 @@ func (r *Reconciler) observeActive(ctx context.Context, ev *models.Event) {
...
+		// later ticks keep them ON until every admitted target confirms curtailment.
+		if !r.reconcileActiveFans(ctx, ev, append(targets, claimed...)) {
+			return
+		}
+		airflowReopened = true
+	}
 	r.dispatchClaimedCurtailTargets(cmdCtx, ev, claimed)
+	if !airflowReopened {
+		_ = r.reconcileActiveFans(ctx, ev, append(targets, claimed...))
+	}
```
rongxin-liu, line 882: Worth treating a candidate-query failure or stale/nil telemetry as insufficient fan-safety evidence before this reassertion? Targets remain `CONFIRMED` in those cases, so this can send OFF again while a rebooted miner may be hashing. Perhaps reopen ON, retain the airflow marker, and require fresh confirmation plus the full delay; a query-failure and stale-telemetry test would be worth adding.

server/sqlc/queries/infrastructure_device.sql
```diff
@@ -128,3 +128,13 @@ SELECT COUNT(*)
...
+
+-- name: CountActiveCurtailmentEventsByInfrastructureDevices :one
+SELECT COUNT(*)
+FROM curtailment_event
+WHERE org_id = sqlc.arg('org_id')
+  AND (
+    state IN ('pending', 'active', 'restoring')
+    OR fan_last_error IS NOT NULL
+  )
+  AND facility_fan_device_ids && sqlc.arg('infrastructure_device_ids')::BIGINT[];
```
rongxin-liu, line 140: Do we want to distinguish nonterminal ownership from terminal recovery errors here? Including every `fan_last_error` row blocks same-site enabled/configuration and commissioning repairs, while Force Release retries the same broken settings, so recovery can deadlock. Maybe keep moves/deletes blocked but allow serialized repair-only updates for terminal-error-only owners, then test repair -> Force Release.

server/internal/domain/curtailment/service.go
```diff
@@ -482,6 +505,17 @@ func (s *Service) AdminTerminate(ctx context.Context, req AdminTerminateRequest)
...
+		event, err := s.store.GetEventByUUID(ctx, req.OrgID, req.EventUUID)
+		if err != nil {
+			return nil, err
+		}
+		if event.State == models.EventStatePending || event.State == models.EventStateRestoring {
+			if err := s.restoreFansForOperatorRecovery(ctx, event); err != nil {
+				return nil, err
+			}
+		}
+	}
```
rongxin-liu, line 531: Worth serializing this state decision, fan-ON recovery, and terminal transition under the same event lock. AdminTerminate could read `ACTIVE` and skip this branch, then race with Stop moving the row to `RESTORING`, after which the terminal update succeeds without an ON attempt. Perhaps a locked store operation similar to Force Release plus a concurrent Stop/AdminTerminate test would close that gap.

rongxin-liu (approved):

PR #786, opened 2026-07-21
refactor(telemetry): isolate activation results (7/9)

ankitgoswami:
Reviewable diff: +369/-637 across 5 files (excludes generated, test, and story files).

## Summary

Telemetry background collection can now stop, fully drain, and restart without carrying queued work or results into a later Fleet activation. Shutdown closes admission to new telemetry producers, lets already-admitted polling and request refreshes finish, flushes their status and metrics under bounded contexts, and requeues unsent scheduler work. The unused legacy broadcaster path is removed so telemetry exposes one unambiguous `Start`/`Stop` lifecycle.

**Stack:**

1. Lifecycle foundation [#780](https://github.com/block/proto-fleet/pull/780) — merged
2. Diagnostics [#781](https://github.com/block/proto-fleet/pull/781) — merged
3. IP scanning [#782](https://github.com/block/proto-fleet/pull/782) — merged
4. Scheduling [#783](https://github.com/block/proto-fleet/pull/783) — merged
5. Curtailment [#785](https://github.com/block/proto-fleet/pull/785) — merged
6. Command execution [#787](https://github.com/block/proto-fleet/pull/787) — merged
7. **Telemetry [#786](https://github.com/block/proto-fleet/pull/786) — this PR**
8. Runtime orchestration [#784](https://github.com/block/proto-fleet/pull/784) — open
9. Catalog and `fleetd` cutover [#788](https://github.com/block/proto-fleet/pull/788) — open

This PR targets `main` and builds on the `runtimejobs.Lifecycle` contract merged in #780. #784 is an independent sibling. #788 is temporarily based on the integration branch until #784 and #786 merge, after which it will be rebased and retargeted to `main`. Passive-mode coordination, epoch fencing, and active request gating remain later HA work.

## How it works

Each `Start` creates a fresh telemetry activation with its own task queues, result queues, flush requests, cancellation, and goroutine tracking. Pollers and request-driven refreshes register as producers before publishing work. Workers collect device data, and activation-owned writers batch status and metrics into their existing stores.

When the activation context is canceled or `Stop` is called, the service rejects new producer registration and cancels background admission. Work already admitted is allowed to finish; writers stay alive until those producers exit, then drain the final queued results and perform bounded store flushes. Pending scheduler tasks are requeued one device at a time so a concurrently removed device cannot prevent later valid devices from being restored.

The activation remains published while cleanup is still running. If the caller's `Stop` deadline expires, cleanup continues and a replacement `Start` is rejected until the old activation exits, preventing overlapping owners. `fleetd` uses the same bounded lifecycle stop helper as the other runtime jobs.

The former per-organization broadcaster was unreachable from production callers: the public metrics stream uses its own request-scoped polling loop, and the raw subscription helper had no callers. Removing that dead path also removes the overlapping `Close` operation.

```mermaid
flowchart LR
    S["Start activation"] --> Q["Fresh task and result queues"]
    Q --> P["Pollers and request refreshes"]
    P --> W["Telemetry workers"]
    W --> R["Status and metrics results"]
    R --> B["Batch writers"]
    B --> D["Device and telemetry stores"]
    X["Stop or activation cancellation"] --> A["Close producer admission"]
    A --> G["Drain admitted producers"]
    G --> F["Final bounded flush"]
    F --> U["Requeue unsent scheduler work"]
    U --> E["Activation stopped and restartable"]
```

```mermaid
stateDiagram-v2
    [*] --> Stopped
    Stopped --> Running: Start
    Running --> Stopping: Stop or activation canceled
    Stopping --> Stopped: producers and writers drained
    Stopping --> Stopping: caller deadline expires; cleanup continues
    Stopping --> Stopping: Start rejected
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/domain/telemetry/activation.go` (new) | Owns per-run queues, producer admission, cancellation, drain, requeue, and lifecycle state | Defines the no-overlap and cross-activation isolation guarantees |
| `server/internal/domain/telemetry/service.go` | Routes polling, workers, request refreshes, flushes, and writers through the active run | Review producer/writer ordering, bounded flushes, and inactive request behavior |
| `server/internal/domain/telemetry/broadcaster.go` (deleted) | Removes an unused raw-update broadcaster and subscription API | Eliminates a second, otherwise ambiguous process lifetime |
| `server/internal/domain/telemetry/broadcaster_metrics.go` | Clarifies that the active metrics observer consumes writer results, not the deleted broadcaster | Prevents the similarly named but unrelated alert-metrics path from being confused with dead code |
| `server/cmd/fleetd/main.go` | Stops telemetry through the shared runtime-job helper | Keeps shutdown on the common lifecycle contract |
| Telemetry service tests | Cover cancellation, restart, drain, queue isolation, request refreshes, timeouts, and scheduler requeue | Exercise lifecycle edges under the race detector |

- Allocate queues per activation instead of attempting to sanitize shared buffered queues after shutdown.
- Close admission immediately but drain already-registered refreshes and workers instead of canceling work after it may have reached a device.
- Let writers outlive activation cancellation long enough to drain producer results, while bounding each final store interaction to five seconds.
- Requeue pending scheduler work per device rather than as one batch, trading a few shutdown-only calls for progress after a removed-device error.
- Keep a timed-out activation in the stopping state until cleanup actually finishes rather than allowing a replacement owner to overlap it.
- Delete the unused broadcaster and overlapping `Close` API rather than introducing another ownership abstraction for dead code.

## Related

Related: #740

## Testing & validation

- `go test -short -race -count=1 ./internal/domain/telemetry ./cmd/fleetd`
- `just _lint-server` in the Hermit environment reported zero issues.
- `git diff --check`

Watch telemetry poll/write errors, sample freshness, status regressions, scheduler requeue warnings, and shutdown drain logs across at least two polling intervals. Roll back if samples stop arriving, status rewinds to an older observation, pending devices stop being rescheduled, or lifecycle drains repeatedly fail.

---


server/internal/domain/telemetry/activation.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("stop telemetry service: %w", ctx.Err())
+	}
+}
+
+// Close stops polling and closes per-organization broadcasters during process
+// teardown. Runtime demotion should call Stop so read-only streams can remain
+// independent of the active polling lifecycle.
+func (s *TelemetryService) Close(ctx context.Context) error {
```
mcharles-square: Exposing `Close` and `Stop` feels like it adds ambiguity even with the comment/doc. How can we make this harder to misuse?

mcharles-square (approved):

server/internal/domain/telemetry/activation.go
ankitgoswami: Good point. Close is a superset of Stop, so callers need to figure out two lifetimes on one type. broadcasters are essentially legacy code that's not currently used now (ever since miner-list streaming was replaced with polling). Instead of adding more complexity I'm going to just get rid of the dead code now, it is time :)

ankitgoswami: 🤖

> Exposing `Close` and `Stop` feels like it adds ambiguity even with the comment/doc. How can we make this harder to misuse?

Addressed differently in `129ecf62e`: removed the unused legacy broadcaster path instead of adding another lifecycle abstraction. `TelemetryService` now exposes only its restartable `Start`/`Stop` lifecycle, and fleetd uses the shared lifecycle shutdown helper. The broadcaster implementation, raw subscription API, status-publish loop, and broadcaster-only tests are gone; the similarly named metrics observer remains because it is active and unrelated.

PR #787, opened 2026-07-21
refactor(command): make execution restartable (6/9)

ankitgoswami:
Reviewable diff: +277/-100 across 7 files (excludes generated, test, and story files).

## Summary

Command execution now supports repeatable activation and shutdown without overlapping processors or canceling commands that have already reached a miner. Shutdown closes admission immediately, drains admitted queue writes and command workers within the caller's budget, and only force-cancels that work when the budget expires.

**Stack:**

- Foundation: [#780](https://github.com/block/proto-fleet/pull/780) (merged)
- Parallel domain refactors: diagnostics [#781](https://github.com/block/proto-fleet/pull/781), IP scanning [#782](https://github.com/block/proto-fleet/pull/782), scheduling [#783](https://github.com/block/proto-fleet/pull/783), curtailment [#785](https://github.com/block/proto-fleet/pull/785), **command execution [#787](https://github.com/block/proto-fleet/pull/787)**, and telemetry [#786](https://github.com/block/proto-fleet/pull/786)
- Orchestration: [#784](https://github.com/block/proto-fleet/pull/784)
- Catalog and `fleetd` cutover: [#788](https://github.com/block/proto-fleet/pull/788)

This PR targets `main` and builds on the `runtimejobs.Lifecycle` contract from #780. The sibling domain refactors are independent; #788 contains the later catalog and `fleetd` cutover. Passive-mode coordination, epoch fencing, and request gating remain later HA work.

## How it works

`Start` creates one activation with two contexts. The admission context owns the queue processor and stuck-message reaper; canceling it stops new dequeues and closes the enqueue gate. The work context belongs to operations already admitted by that activation, including queue writes and command workers.

Command requests hold the activation gate through `Enqueue`, `EnqueueMany`, or worker-name reapply, making enqueue ordering atomic with shutdown. `Stop` closes that gate and cancels admission, then waits for the activation's wait group. A clean drain preserves an in-flight command’s normal result bookkeeping; if the caller’s stop deadline expires, `Stop` force-cancels the work context and returns the deadline error while the activation remains unavailable until its goroutines actually exit. If that forced cancellation interrupts a queue transaction, the admission wrapper classifies it as stopped before enqueue so the saved batch is marked finished instead of remaining `PENDING`; request cancellation remains distinguishable and follows the normal error path.

Callers that cancel the activation context still pair that cancellation with bounded `Stop` so admitted work cannot outlive its dependencies. The shared test service provider owns this teardown for both direct service tests and the HTTP infrastructure harness.

The queue processor reserves currently available worker slots before dequeueing and passes that limit to the database queue. Operational dequeue failures exhaust the configured retries, back off for one polling interval, and retry while the activation remains open.

```mermaid
flowchart LR
    S["Start activation"] --> A["Admission context"]
    S --> W["Work context"]
    A --> P["Queue processor"]
    A --> R["Stuck-message reaper"]
    P --> C["Reserve worker capacity"]
    C --> Q["Dequeue up to capacity"]
    Q --> X["Admitted command workers"]
    G["Command request"] --> L["Activation admission gate"]
    L --> E["Enqueue or worker-name reapply"]
    E --> W
    X --> W
    T["Stop or activation cancellation"] --> K["Close admission"]
    K --> D["Bounded drain of admitted work"]
    D --> N["Safe later activation"]
```

```mermaid
sequenceDiagram
    participant Request
    participant Activation
    participant Queue
    participant Worker
    participant Stop
    Request->>Activation: Acquire admission
    Activation->>Queue: Persist pending command
    Queue-->>Activation: Commit
    Activation-->>Request: Release admission
    Activation->>Worker: Run admitted command
    Stop->>Activation: Close admission and stop dequeue
    Activation-->>Stop: Wait for admitted work
    Worker-->>Activation: Finish command and bookkeeping
    Activation-->>Stop: Drain complete
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/domain/command/execution_service.go` | Adds activation-owned admission/work contexts, cause-aware forced cancellation, bounded drain, capacity reservation, and persistent dequeue retry | Defines the no-overlap and no-premature-cancellation lifecycle guarantees |
| `server/internal/domain/command/service.go` | Holds activation admission through all queue-write paths and finishes batches whose enqueue is force-canceled | Removes the enqueue-versus-Stop TOCTOU window and prevents orphaned `PENDING` batches |
| `server/internal/infrastructure/queue/` | Adds and enforces an explicit dequeue limit | Prevents claiming more work than the activation can run |
| `server/cmd/fleetd/main.go` | Uses context-free construction and bounded standalone stop | Preserves standalone startup while making shutdown explicit |
| `server/internal/testutil/` | Replaces cancel-only cleanup with a bounded execution-service stop | Keeps admitted workers from outliving test DB/plugin dependencies |
| Command tests | Cover cancellation, restart, drain, timeout, admission ordering, and retry behavior | Exercises the lifecycle boundaries under the race detector |
| `server/internal/infrastructure/queue/mocks/` | Generated mock signature update | Generated — skip |

- Separate admission cancellation from admitted-work cancellation instead of sharing one context that can turn a successfully delivered miner command into a retry.
- Require lifecycle owners to pair activation cancellation with bounded `Stop` rather than reattaching admitted work to an unbounded cancellation source.
- Use one private activation gate across all enqueue paths instead of exposing a reusable lease abstraction before another consumer needs it.
- Keep an activation published until every admitted goroutine exits, even after a stop timeout, rather than allowing a replacement processor to overlap it.
- Claim only currently reserved worker capacity rather than caching unclaimed queue candidates whose ownership could become stale across multiple consumers.
- Retry operational dequeue failures inside the activation rather than adding a second asynchronous health interface to every lifecycle.

## Related

Related: #740

## Testing & validation

- `go test -short -race -count=1 ./internal/domain/command ./internal/infrastructure/queue ./cmd/fleetd`
- Targeted lifecycle race tests cover activation cancellation followed by bounded drain, forced cancellation at the stop deadline, cleanup of shutdown-canceled enqueue attempts, rejection of new admission, and deterministic enqueue-versus-Stop ordering.
- Targeted `golangci-lint` reports zero issues; `git diff --check` passes.

Watch dequeue failure/retry logs, queue depth and processing age, active worker counts, stuck-message reaping, and shutdown drain errors. Roll back if commands stop leaving the queue, worker concurrency exceeds configuration, duplicate dispatch appears, or drain deadlines recur.

---

### DIFF
--- a/server/internal/domain/command/execution_service.go
+++ b/server/internal/domain/command/execution_service.go
@@ -73,9 +73,9 @@ type ExecutionService struct {
 	lifecycleMu           sync.Mutex
 	queueProcessorRunning bool
 	runCancel             context.CancelFunc
+	runCtxDone            <-chan struct{}
 	runWG                 sync.WaitGroup
 	runDone               <-chan struct{}
-	stopping              bool
 }
 
 var _ runtimejobs.Lifecycle = (*ExecutionService)(nil)
@@ -121,11 +121,13 @@ func (es *ExecutionService) WithMetricsEmitter(emitter MetricsEmitter) *Executio
 // active.
 func (es *ExecutionService) Start(ctx context.Context) error {
 	es.lifecycleMu.Lock()
-	if es.stopping {
-		es.lifecycleMu.Unlock()
-		return fmt.Errorf("command execution service is still stopping")
-	}
 	if es.runCancel != nil {
+		select {
+		case <-es.runCtxDone:
+			es.lifecycleMu.Unlock()
+			return fmt.Errorf("command execution service activation is still draining")
+		default:
+		}
 		es.lifecycleMu.Unlock()
 		return nil
 	}
@@ -133,6 +135,7 @@ func (es *ExecutionService) Start(ctx context.Context) error {
 	runCtx, runCancel := context.WithCancel(ctx)
 	runDone := make(chan struct{})
 	es.runCancel = runCancel
+	es.runCtxDone = runCtx.Done()
 	es.runDone = runDone
 	es.queueProcessorRunning = true
 
@@ -164,8 +167,8 @@ func (es *ExecutionService) finishRun(done chan struct{}) {
 	es.lifecycleMu.Lock()
 	if es.runDone == (<-chan struct{})(done) {
 		es.runCancel = nil
+		es.runCtxDone = nil
 		es.runDone = nil
-		es.stopping = false
 		es.queueProcessorRunning = false
 	}
 	es.lifecycleMu.Unlock()
@@ -180,10 +183,7 @@ func (es *ExecutionService) Stop(ctx context.Context) error {
 		es.lifecycleMu.Unlock()
 		return nil
 	}
-	if !es.stopping {
-		es.stopping = true
-		es.runCancel()
-	}
+	es.runCancel()
 	done := es.runDone
 	es.lifecycleMu.Unlock()
 
--- a/server/internal/domain/command/execution_service.go
+++ b/server/internal/domain/command/execution_service.go
@@ -117,7 +117,8 @@ func (es *ExecutionService) WithMetricsEmitter(emitter MetricsEmitter) *Executio
 	return es
 }
 
-// Start starts the queue processor thread if it is not already running.
+// Start runs command execution for the lifetime of ctx if it is not already
+// active.
 func (es *ExecutionService) Start(ctx context.Context) error {
 	es.lifecycleMu.Lock()
 	if es.stopping {
--- a/server/internal/domain/command/execution_service.go
+++ b/server/internal/domain/command/execution_service.go
@@ -70,15 +70,36 @@ type ExecutionService struct {
 
 	workerSemaphore chan struct{}
 
-	lifecycleMu           sync.Mutex
-	queueProcessorRunning bool
-	runCancel             context.CancelFunc
-	runCtxDone            <-chan struct{}
-	runWG                 sync.WaitGroup
-	runDone               <-chan struct{}
+	lifecycleMu sync.Mutex
+	run         *executionRun
+}
+
+type executionRun struct {
+	admissionCtx    context.Context //nolint:containedctx // scoped to this activation's admission phase
+	cancelAdmission context.CancelFunc
+	workCtx         context.Context //nolint:containedctx // scoped to this activation's drain phase
+	cancelWork      context.CancelFunc
+	accepting       bool
+	wg              sync.WaitGroup
+	done            chan struct{}
 }
 
 var _ runtimejobs.Lifecycle = (*ExecutionService)(nil)
+
+var errExecutionNotAccepting = errors.New("command execution service is not accepting work")
+
+func newExecutionRun(ctx context.Context) *executionRun {
+	admissionCtx, cancelAdmission := context.WithCancel(ctx)
+	workCtx, cancelWork := context.WithCancel(context.WithoutCancel(ctx))
+	return &executionRun{
+		admissionCtx:    admissionCtx,
+		cancelAdmission: cancelAdmission,
+		workCtx:         workCtx,
+		cancelWork:      cancelWork,
+		accepting:       true,
+		done:            make(chan struct{}),
+	}
+}
 
 func NewExecutionService(config *Config, conn *sql.DB, messageQueue queue.MessageQueue, encryptService *encrypt.Service, tokenService *tokenDomain.Service, minerService CachedMinerGetter, deviceStore stores.DeviceStore, telemetryListener TelemetryListener, filesService *files.Service) *ExecutionService {
 	if config.StuckMessageTimeout <= 0 {
@@ -94,18 +115,17 @@ func NewExecutionService(config *Config, conn *sql.DB, messageQueue queue.Messag
 		config.FirmwareUpdateStuckTimeout = 20 * time.Minute
 	}
 	return &ExecutionService{
-		config:                config,
-		conn:                  conn,
-		messageQueue:          messageQueue,
-		encryptService:        encryptService,
-		tokenService:          tokenService,
-		minerService:          minerService,
-		deviceStore:           deviceStore,
-		telemetryListener:     telemetryListener,
-		filesService:          filesService,
-		metricsEmitter:        NoCommandMetrics(),
-		workerSemaphore:       make(chan struct{}, config.MaxWorkers),
-		queueProcessorRunning: false,
+		config:            config,
+		conn:              conn,
+		messageQueue:      messageQueue,
+		encryptService:    encryptService,
+		tokenService:      tokenService,
+		minerService:      minerService,
+		deviceStore:       deviceStore,
+		telemetryListener: telemetryListener,
+		filesService:      filesService,
+		metricsEmitter:    NoCommandMetrics(),
+		workerSemaphore:   make(chan struct{}, config.MaxWorkers),
 	}
 }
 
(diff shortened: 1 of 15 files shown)


server/internal/domain/command/execution_service.go
```diff
@@ -111,44 +117,92 @@ func (es *ExecutionService) WithMetricsEmitter(emitter MetricsEmitter) *Executio
...
+// Stop cancels the active queue processor, reaper, and workers and waits for
+// all execution-owned goroutines to drain.
+func (es *ExecutionService) Stop(ctx context.Context) error {
+	es.lifecycleMu.Lock()
+	if es.runCancel == nil {
+		es.lifecycleMu.Unlock()
+		return nil
+	}
+	es.queueProcessorRunning = false
+	es.runCancel()
```
rongxin-liu: Worth separating stopping new dequeues from canceling admitted workers here? Because each worker inherits the run context and post-command bookkeeping survives cancellation, shutdown can turn a command that already reached the miner into PENDING or FAILED and later repeat it.

Perhaps we could first close admission and stop dequeues, then drain admitted workers under the shutdown budget (or explicitly classify activation cancellation without consuming retry budget), with a database-backed in-flight command test?

server/internal/domain/command/service.go
```diff
@@ -906,6 +915,9 @@ func (s *Service) processCommand(ctx context.Context, command *Command) (*Comman
 		return nil, fleeterror.NewInternalErrorf("error saving command batch log to db: %v", err)
 	}
 
+	if !s.executionService.IsRunning() {
+		return nil, s.finishUnenqueuedCommandBatch(ctx, batchLogIdentifier)
+	}
```
rongxin-liu: Should we make this check linearizable with enqueue: Stop can run after IsRunning returns true but before Enqueue commits, leaving a successful PENDING batch with no processor. Maybe we could use an activation-scoped admission lease shared with Stop and hold it through Enqueue or EnqueueMany (including worker-name reapply), then add a deterministic test for that interleaving?

server/internal/domain/command/execution_service.go
```diff
@@ -297,48 +351,90 @@ func (es *ExecutionService) dequeueWithRetry(ctx context.Context) ([]queue.Messa
...
-			if err != nil {
-				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-					return err
-				}
-				return fleeterror.NewInternalErrorf("error dequeueing messages: %v", err)
+		reservedSlots, err := es.reserveWorkerSlots(ctx)
+		if err != nil {
+			return err
+		}
+		messages, err := es.dequeueWithRetry(ctx, reservedSlots)
```
rongxin-liu, line 418: We could probably dequeue batching under saturation. With staggered worker completions, reservedSlots will often be 1, so this can run a PostgreSQL transaction and ordered anti-join per completed command instead of amortizing discovery across a batch. For testing could also benchmark the fleet-wide command path and, if confirmed, cache unclaimed candidates or otherwise batch discovery while still claiming no more than reserved capacity?

rongxin-liu (approved):

server/internal/domain/command/execution_service.go
ankitgoswami: 🤖

> Worth separating stopping new dequeues from canceling admitted workers here?
>
> Should we make this check linearizable with enqueue?

Addressed differently in `3bb34cca0`: one private activation gate now covers `Enqueue`, `EnqueueMany`, and worker-name reapply. Stop closes admission and cancels dequeue/reaper work immediately, while admitted queue writes and command workers drain on a separate work context; only an expired Stop deadline force-cancels them. Deterministic race tests cover clean worker draining, forced cancellation, and enqueue-versus-Stop ordering.

server/internal/domain/command/service.go
ankitgoswami: 🤖

> Worth separating stopping new dequeues from canceling admitted workers here?
>
> Should we make this check linearizable with enqueue?

Addressed differently in `3bb34cca0`: one private activation gate now covers `Enqueue`, `EnqueueMany`, and worker-name reapply. Stop closes admission and cancels dequeue/reaper work immediately, while admitted queue writes and command workers drain on a separate work context; only an expired Stop deadline force-cancels them. Deterministic race tests cover clean worker draining, forced cancellation, and enqueue-versus-Stop ordering.

server/internal/domain/command/execution_service.go
ankitgoswami: 🤖

> With staggered worker completions, reservedSlots will often be 1, so this can run a PostgreSQL transaction and ordered anti-join per completed command instead of amortizing discovery across a batch.

Keeping this unchanged in this PR. The reserved-slot count is deliberately the claim limit so the queue never creates `PROCESSING` rows without worker capacity. Caching unclaimed candidates would add stale ownership and multi-consumer coordination to a path that has no measured bottleneck yet. We should benchmark or add query metrics first, then batch discovery only if saturation data shows this transaction is material.

server/internal/domain/command/execution_service.go
```diff
@@ -69,12 +70,38 @@ type ExecutionService struct {
...
+	done            chan struct{}
+}
+
+var _ runtimejobs.Lifecycle = (*ExecutionService)(nil)
+
+var errExecutionNotAccepting = errors.New("command execution service is not accepting work")
+
+func newExecutionRun(ctx context.Context) *executionRun {
+	admissionCtx, cancelAdmission := context.WithCancel(ctx)
+	workCtx, cancelWork := context.WithCancel(context.WithoutCancel(ctx))
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> keep the worker context cancelable from the activation or update those callers to perform a bounded `Stop`.

Addressed with the second option in `9c11cde25`. Admitted workers remain detached from activation cancellation so a command that reached a miner cannot be converted into a retry. `NewServiceProvider` now owns teardown for every direct and infrastructure caller: cleanup cancels admission, then calls `ExecutionService.Stop` with a bounded 10-second context while DB and plugin dependencies remain alive. The cancel-only hook was removed, and the lifecycle regression now covers activation cancellation followed by bounded Stop without prematurely canceling the admitted worker.

server/internal/domain/command/service.go
```diff
@@ -906,16 +915,20 @@ func (s *Service) processCommand(ctx context.Context, command *Command) (*Comman
...
-	} else {
-		err = s.messageQueue.EnqueueMany(ctx, batchLogIdentifier, command.commandType, queuePayloads)
-		if err != nil {
-			return nil, fleeterror.NewInternalErrorf("error enqueuing per-device command payloads: %v", err)
+		return s.messageQueue.EnqueueMany(workCtx, batchLogIdentifier, command.commandType, queuePayloads)
+	})
+	if errors.Is(err, errExecutionNotAccepting) {
+		return nil, s.finishUnenqueuedCommandBatch(ctx, batchLogIdentifier)
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 Fixed in `957c15139`.

> Treat stop-canceled enqueue failures like the `finishUnenqueuedCommandBatch` path before returning.

`withAdmission` now attaches a distinct cause when `Stop` force-cancels admitted work. Failed `Enqueue`/`EnqueueMany` and worker-name reapply operations therefore take their existing unenqueued-batch cleanup path, while ordinary request cancellation remains `context.Canceled`. Added deterministic regression coverage for both cases.

PR #788, opened 2026-07-21
refactor(fleetd): start jobs from runtime catalog (9/9)

ankitgoswami:
Reviewable diff: +254/-112 across 2 files (excludes generated, test, and story files).

## Summary

`fleetd` now activates its 12 background jobs from one explicit ordered catalog and drains them as one runtime group. Standalone deployments still run the same work, while the HA supervisor gains a single boundary it can later leave stopped on passive instances and toggle on ownership changes.

**Stack:**

1. Lifecycle foundation [#780](https://github.com/block/proto-fleet/pull/780) — merged
2. Diagnostics [#781](https://github.com/block/proto-fleet/pull/781) — merged
3. IP scanning [#782](https://github.com/block/proto-fleet/pull/782) — merged
4. Scheduling [#783](https://github.com/block/proto-fleet/pull/783) — merged
5. Curtailment [#785](https://github.com/block/proto-fleet/pull/785) — merged
6. Command execution [#787](https://github.com/block/proto-fleet/pull/787) — merged
7. Telemetry [#786](https://github.com/block/proto-fleet/pull/786) — merged
8. Runtime orchestration [#784](https://github.com/block/proto-fleet/pull/784) — merged
9. **Catalog and `fleetd` cutover [#788](https://github.com/block/proto-fleet/pull/788) — this PR**

This final PR now targets the latest `main`, which includes the runtime `Job`/`Group` implementation from #784. Its reviewable diff contains only the catalog and `fleetd` cutover.

Passive-mode coordinator wiring, epoch fencing, and active request gating remain HA follow-up work. This PR supplies the activation boundary; it does not choose when Fleet is active.

## How it works

Construction remains in dependency order, but no active loop starts during construction. `fleetd` builds the canonical catalog and starts one `runtimejobs.Group`; the group passes one activation context to every catalog entry. Four context-driven loops—identity cleanup, command-artifact cleanup, chunked-upload cleanup, and system monitoring—use a small `backgroundLoop` lifecycle adapter. The other eight entries are the restartable services introduced by the earlier stack.

Command execution is a drain dependency for later producers. A small stop-order wrapper keeps it alive when the group broadcasts activation cancellation, allowing schedule and telemetry producers to finish first. Group shutdown gets one bounded 10-second attempt; incomplete cleanup is terminal rather than retried. Command execution then receives its own fresh bounded stop budget so it is still stopped if an earlier producer consumed the group budget.

System monitoring waits on a listener-ready signal before it emits its first heartbeat. Other jobs complete synchronous startup before `fleetd` binds the HTTP listener; once binding succeeds, the signal releases system monitoring and the server begins accepting connections. Telemetry now has only the common `Start`/`Stop` lifecycle, so group shutdown is its sole cleanup owner.

```mermaid
flowchart TB
    C["Construct services without starting jobs"] --> K["Build ordered 12-job catalog"]
    K --> G["Start runtime Group"]
    G --> F["Startup failure"]
    F --> R["Cancel and rollback started prefix"]
    G --> L["Bind HTTP listener"]
    L --> H["Release system-monitoring heartbeat"]
    H --> A["Standalone active runtime"]
    A --> S["Shutdown or future demotion"]
    S --> D["Cancel and reverse-stop under one bounded budget"]
    D --> X["Final command-execution stop"]
    X --> P["Process teardown"]
```

```mermaid
sequenceDiagram
    participant O as Runtime owner
    participant G as Job group
    participant P as Schedule and telemetry producers
    participant C as Command execution
    O->>G: Stop with 10-second budget
    G->>P: Cancel activation and drain in reverse order
    Note over C: Cancellation is detached to preserve the drain dependency
    G->>C: Stop if group budget remains
    O->>C: Final independent bounded Stop
```

## Areas of the code involved

| --- | --- | --- |
| `server/cmd/fleetd/runtime_jobs.go` (new) | Defines the context-loop adapter, command drain ordering, bounded group shutdown, and canonical ordered catalog | Review exact job membership, order, optional omission, and shutdown dependency handling |
| `server/cmd/fleetd/main.go` | Separates construction from activation and cuts startup/shutdown over to `Group` | Review startup strictness, listener/heartbeat ordering, and single-owner shutdown |
| Focused tests | Pin catalog order and optionals, adapter restart/drain, command drain ordering, bounded final cleanup, and heartbeat cancellation behavior | Protect the HA-sensitive lifecycle edges |

- Keep catalog policy in `fleetd`; domain services know lifecycle, not deployment names or enablement.
- Treat the group `Start` context as the shared activation lifetime and `Stop` as the explicit bounded drain.
- Use one local adapter for four existing context loops instead of introducing new service types or a general registry.
- Omit optional jobs with true nil interfaces; disabled alert metrics and system monitoring are not registered.
- Keep command execution alive until its producers drain, then stop it under its own final bounded budget.
- Treat incomplete group cleanup as terminal rather than retrying a potentially unsafe activation.
- Start the group before listener binding while gating the first system heartbeat on a successful bind.
- Treat any grouped startup failure as fatal and roll back instead of continuing with a partially active Fleet.

## Related

Related: #740

## Testing & validation

- `go test -short -race -count=1 ./cmd/fleetd ./internal/infrastructure/sysmon ./internal/runtimejobs ./internal/domain/telemetry`
- Repo-pinned `golangci-lint` 2.9.0 (`0 issues`)
- `git diff --check` against current `main`

For one normal job interval after rollout, watch startup errors containing `start runtime jobs`, shutdown errors containing `failed to stop runtime jobs`, command queue depth/age, telemetry freshness, schedule dispatch, scanner completion, MQTT subscription state, curtailment actions, diagnostics closing, cleanup logs, and system heartbeat freshness. Roll back on missing job activity, duplicate side effects, stale telemetry/control inputs, or any group drain failure.

### DIFF
--- a/server/cmd/fleetd/runtime_jobs.go
+++ b/server/cmd/fleetd/runtime_jobs.go
@@ -103,20 +103,27 @@ type runtimeJobGroupStopper interface {
 }
 
 // stopRuntimeJobGroup gives the group one graceful-shutdown budget, then one
-// fresh bounded budget to retry only the jobs whose first stop did not finish.
-func stopRuntimeJobGroup(group runtimeJobGroupStopper, timeout time.Duration) {
+// fresh bounded retry. Command execution receives a final independent budget
+// after producer retries because its activation is detached from group
+// cancellation to preserve shutdown ordering.
+func stopRuntimeJobGroup(group runtimeJobGroupStopper, commandExecution runtimejobs.Lifecycle, timeout time.Duration) {
 	shutdownCtx, cancel := context.WithTimeout(context.Background(), timeout)
 	err := group.Stop(shutdownCtx)
 	cancel()
-	if err == nil {
-		return
+	if err != nil {
+		slog.Error("failed to stop runtime jobs", "error", err)
+		drainCtx, drainCancel := context.WithTimeout(context.Background(), timeout)
+		if err := group.Stop(drainCtx); err != nil {
+			slog.Error("failed to drain runtime jobs", "error", err)
+		}
+		drainCancel()
 	}
 
-	slog.Error("failed to stop runtime jobs", "error", err)
-	drainCtx, drainCancel := context.WithTimeout(context.Background(), timeout)
-	defer drainCancel()
-	if err := group.Stop(drainCtx); err != nil {
-		slog.Error("failed to drain runtime jobs", "error", err)
+	commandCtx, commandCancel := context.WithTimeout(context.Background(), timeout)
+	err = commandExecution.Stop(commandCtx)
+	commandCancel()
+	if err != nil {
+		slog.Error("failed to stop command execution after runtime jobs", "error", err)
 	}
 }
 
--- a/server/cmd/fleetd/runtime_jobs.go
+++ b/server/cmd/fleetd/runtime_jobs.go
@@ -3,7 +3,9 @@ package main
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"sync"
+	"time"
 
 	"github.com/block/proto-fleet/server/internal/runtimejobs"
 )
@@ -73,6 +75,28 @@ func (l *backgroundLoop) Stop(ctx context.Context) error {
 		return nil
 	case <-ctx.Done():
 		return fmt.Errorf("stop background loop: %w", ctx.Err())
+	}
+}
+
+type runtimeJobGroupStopper interface {
+	Stop(ctx context.Context) error
+}
+
+// stopRuntimeJobGroup gives the group one graceful-shutdown budget, then one
+// fresh bounded budget to retry only the jobs whose first stop did not finish.
+func stopRuntimeJobGroup(group runtimeJobGroupStopper, timeout time.Duration) {
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	err := group.Stop(shutdownCtx)
+	cancel()
+	if err == nil {
+		return
+	}
+
+	slog.Error("failed to stop runtime jobs", "error", err)
+	drainCtx, drainCancel := context.WithTimeout(context.Background(), timeout)
+	defer drainCancel()
+	if err := group.Stop(drainCtx); err != nil {
+		slog.Error("failed to drain runtime jobs", "error", err)
 	}
 }
 
--- a/server/cmd/fleetd/runtime_jobs.go
+++ b/server/cmd/fleetd/runtime_jobs.go
@@ -35,6 +35,9 @@ func (l stopOrderedLifecycle) Start(ctx context.Context) error {
 }
 
 func (l stopOrderedLifecycle) Stop(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


ankitgoswami:
> [MEDIUM] Shutdown timeout can invert command execution dependency ordering
Category: Reliability
Location: [server/cmd/fleetd/runtime_jobs.go:108](https://github.com/block/proto-fleet/blob/fd963c06a1b71f91252945cf17ade8e0d365af0a/server/cmd/fleetd/runtime_jobs.go#L108)
Description: stopRuntimeJobGroup first stops the whole runtime job group, then gives command execution a separate final stop budget. Because command execution is also registered inside the group via stopOrderedLifecycle, a slow or hung producer earlier in reverse stop order can exhaust the group's shared stop context before the group reaches command execution. In that case command execution is skipped inside the group and only stopped after the group has continued on to cancel later dependencies such as telemetry.
Impact: In-flight command workers can keep running after dependencies they may need for cleanup have already been cancelled. This can turn a bad shutdown into failed or retried command cleanup, especially for flows such as unpair post-processing that rely on telemetry removal, and can leave command state inconsistent until later retry/reaper paths run.
Recommendation: Do not put command execution in the same all-jobs stop pass when it needs a preserved dependency order. Split shutdown into phases with independent budgets: stop command producers first, then stop executionService, then stop telemetry and lower-level dependencies. Add a timeout-path test where a producer stop consumes the group budget and assert command execution is still stopped before telemetry is cancelled.

Command execution does not depend on telemetry’s background activation remaining active. Its only telemetry call is RemoveDevices, which remains usable after telemetry activation cancellation, and the underlying DB etc are torn down. Activation cancellation is broadcast before all drains even on the normal path, so requiring command execution to stop before telemetry cancellation would introduce a new lifecycle requirement rather than fix a timeout-specific inversion.

server/cmd/fleetd/runtime_jobs.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	if err := add("chunked-upload-cleanup", lifecycles.chunkedUploadCleanup); err != nil {
+		return nil, err
+	}
+	if lifecycles.systemMonitoring != nil {
+		if err := add("system-monitoring", lifecycles.systemMonitoring); err != nil {
+			return nil, err
+		}
+	}
```
mcharles-square, line 185: could this ordering be made more generic by adding a _priority_ field to the lifecycle?

mcharles-square (approved):

server/cmd/fleetd/runtime_jobs.go
ankitgoswami: I’d still keep this as an explicit ordered catalog. Since all of these lifecycles are assembled in one place, a priority field would just be a second representation of the same ordering: assign numbers, sort them (and mentally reconstruct the start/stop order when reading the code). The list makes the dependency order visible directly, and it also makes the reverse shutdown behavior easier to reason about.

PR #793, opened 2026-07-21
feat(fleet): show miner site/building/rack in embedded single-miner view

flesher:
Reviewable diff: +80/-1 across 9 files (excludes generated, test, and story files).

## Summary

When you open a Proto Rig's single-miner view embedded inside the Fleet shell, the ProtoOS navigation sidebar now shows a **Location** block — the miner's Fleet placement as site / building / rack, each on its own line. Previously the embedded view exposed only device-local identity (name, IP, MAC, firmware) fetched from the miner itself; it had no idea where the miner physically sits in the fleet hierarchy. This closes that gap using data the Fleet shell already holds, with no server, proto, or database changes.

## How it works

The embedded single-miner experience is the real ProtoOS React app mounted inside the Fleet shell (not an iframe); its data hooks call the miner through the Fleet proxy. Placement (site/building/rack) is **Fleet-side** data the miner does not know about, so it can't come through that proxy. Instead it travels through the app-neutral `MinerMetadata` bridge that already carries name/IP/MAC/firmware from the Fleet shell into the ProtoOS hosting context:

1. **Source** — when a miner is opened from the list, the opener already holds its `MinerStateSnapshot`, which includes a `placement` object with site/building/rack labels. Existing `minerPlacement` helpers (`getMinerSiteLabel`, etc.) read those labels.
2. **Carry** — `buildSingleMinerMetadata` copies the three labels onto the `MinerMetadata` object (three new optional string fields). This object rides in via navigation state and the device-keyed metadata cache, exactly like the existing identity fields, and is handed to the `MinerHostingProvider`.
3. **Render** — `AppLayout` reads the labels off the hosting metadata, filters out empty levels, and — only when fleet-hosted — passes the surviving list to a new `LocationInfo` row in the navigation sidebar's info panel. The row stacks each level on its own line under a single "Location" label.

Gating: the row renders only in `mode="fleet"` and only when at least one level is present. A miner placed at a site but not a rack shows just the site; a fully unplaced miner shows no row. Standalone ProtoOS (direct mode) never shows it, since placement is a Fleet-only concept.

```mermaid
flowchart LR
    A["Miner list (MinerStateSnapshot.placement)"] --> B["buildSingleMinerMetadata()"]
    B --> C["MinerMetadata bridge (+site/building/rack)"]
    C --> D["MinerHostingProvider (mode=fleet)"]
    D --> E["AppLayout: filter empties, gate on fleet-hosted"]
    E --> F["LocationInfo row in nav sidebar"]
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/shared/types/minerMetadata.ts` | Added optional `site`/`building`/`rack` to the app-neutral `MinerMetadata` type | The contract bridging the two apps; kept as plain strings so no proto types leak into ProtoOS |
| `client/.../SingleMinerWrapper/routeState.ts` | `buildSingleMinerMetadata` populates the three labels via existing placement helpers | Where Fleet placement enters the metadata bridge |
| `client/.../SingleMinerWrapper/SingleMinerWrapper.tsx` | Passes the new fields through its metadata memo | Keeps memo identity stable for embedded consumers |
| `client/.../NavigationMenu/InfoItem/LocationInfo/` (new) | New info-panel row; takes a pre-filtered `values` array, stacks each on its own line | Primary new UI; mirrors sibling `InfoItem` styling |
| `client/.../NavigationMenu/Navigation.tsx`, `NavigationMenu.tsx`, `FloatingNavigation.tsx` | Thread a new `locationInfo` prop through desktop + mobile nav; render `LocationInfo` when present | Prop plumbing; render is conditional on the prop existing |
| `client/.../AppLayout/AppLayout.tsx` | Builds the filtered level list and gates on `isFleetHosted` | The decision point for what shows and when |
| `client/.../AppLayout/AppLayout.test.tsx` | Tests for breadcrumb render, partial placement, and direct-mode hiding | Coverage for the gating logic |

- **Reuse the existing `MinerMetadata` bridge** rather than fetching placement inside ProtoOS. No new RPC/proxy surface, and placement stays out of ProtoOS's device-local model. Trade-off below.
- **Plain strings, not proto refs, on the bridge type** — keeps `shared/` app-neutral and prevents ProtoFleet proto types from leaking into ProtoOS.
- **Render gated on `isFleetHosted` + non-empty** — placement is meaningless in standalone ProtoOS, so it never appears there.
- **Known limitation (deferred):** metadata is delivered via navigation state + an in-memory cache, both of which reset on a full page reload. So after a hard refresh or deep-link, the Location row (and the snapshot-sourced name) is absent until re-navigated. IP/MAC/firmware survive because ProtoOS re-fetches them from the miner; placement has no such fallback. A durable fix would fetch the miner's snapshot by identifier (`lookupMinerByIdentifier`) when the cache is cold — not included here.

## Testing & validation

- `tsc --noEmit`, ESLint, and Prettier clean on all changed files.
- Unit tests (`AppLayout.test.tsx`): full breadcrumb renders each level; partial placement (site only) renders just that level; direct (non-fleet) mode hides the row entirely. NavigationMenu suite still green.
- **Not covered:** the refresh/deep-link cold-cache case described above (behaves as documented — row absent until re-navigation).

### DIFF
--- a/client/src/protoOS/components/AppLayout/AppLayout.tsx
+++ b/client/src/protoOS/components/AppLayout/AppLayout.tsx
@@ -45,6 +45,12 @@ const AppLayout = ({
   const ipAddress = useIpAddress();
   const pendingNetworkInfo = useNetworkInfoPending();
 
+  // Placement is a Fleet-only concept and rides in on the hosting metadata.
+  // Show a single "Location" row that stacks site / building / rack on their
+  // own lines, omitting any levels the miner isn't placed at.
+  const locationLines = [metadata.site, metadata.building, metadata.rack].filter(Boolean) as string[];
+  const locationInfo = isFleetHosted && locationLines.length ? { values: locationLines } : undefined;
+
   return (
     <div className="flex min-h-screen bg-surface-base">
       <div className="fixed top-0 left-0 z-40 h-screen grow overflow-hidden">
@@ -57,6 +63,7 @@ const AppLayout = ({
           }}
           isVisible={isMenuOpen}
           closeMenu={() => setIsMenuOpen(false)}
+          locationInfo={locationInfo}
           versionInfo={{
             value: isFleetHosted ? metadata.firmwareVersion || osVersion : osVersion,
             loading: isFleetHosted ? !metadata.firmwareVersion && pendingSystemInfo : pendingSystemInfo,
--- a/client/src/protoFleet/components/SingleMinerWrapper/SingleMinerWrapper.tsx
+++ b/client/src/protoFleet/components/SingleMinerWrapper/SingleMinerWrapper.tsx
@@ -56,6 +56,9 @@ const SingleMinerWrapper = ({ children }: { children: ReactNode }) => {
       ipAddress: cachedMetadata?.ipAddress,
       macAddress: cachedMetadata?.macAddress,
       firmwareVersion: cachedMetadata?.firmwareVersion,
+      site: cachedMetadata?.site,
+      building: cachedMetadata?.building,
+      rack: cachedMetadata?.rack,
     }),
     [cachedMetadata, displayId],
   );
--- a/client/src/protoFleet/components/SingleMinerWrapper/routeState.ts
+++ b/client/src/protoFleet/components/SingleMinerWrapper/routeState.ts
@@ -1,4 +1,9 @@
 import type { MinerStateSnapshot } from "@/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb";
+import {
+  getMinerBuildingLabel,
+  getMinerRackLabel,
+  getMinerSiteLabel,
+} from "@/protoFleet/features/fleetManagement/utils/minerPlacement";
 import type { MinerMetadata } from "@/shared/types/minerMetadata";
 
 export type SingleMinerMetadata = MinerMetadata;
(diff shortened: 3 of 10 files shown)


client/src/protoOS/components/AppLayout/AppLayout.tsx
```diff
@@ -45,6 +45,12 @@ const AppLayout = ({
   const ipAddress = useIpAddress();
   const pendingNetworkInfo = useNetworkInfoPending();
 
+  // Placement is a Fleet-only concept and rides in on the hosting metadata.
+  // Show a single "Location" row that stacks site / building / rack on their
+  // own lines, omitting any levels the miner isn't placed at.
+  const locationLines = [metadata.site, metadata.building, metadata.rack].filter(Boolean) as string[];
```
rongxin-liu, line 51: Worth preserving each value's hierarchy level here? A rack can sit directly under a site with no building, so filtering `[site, undefined, rack]` into two unlabeled lines makes the second line ambiguous. Maybe pass `{ level, value }` entries (or otherwise label the rows) and add a site + rack/no-building case.

client/src/protoOS/components/AppLayout/AppLayout.tsx
```diff
@@ -45,6 +45,12 @@ const AppLayout = ({
...
+  const locationInfo = isFleetHosted && locationLines.length ? { values: locationLines } : undefined;
```
rongxin-liu, line 52: Perhaps add a Fleet-hosted test with no site/building/rack and asserting the Location row is absent. The current tests cover populated Fleet metadata and direct mode, but not this `locationLines.length === 0` branch.

rongxin-liu (approved):

client/src/protoOS/components/AppLayout/AppLayout.tsx
flesher: designs call was to just print the levels w/o label and only where they exist. Easy to revisit later if we decide this gets confusing

client/src/protoOS/components/AppLayout/AppLayout.tsx
flesher: 🤖 Good call — added `hides the location row when fleet-hosted with no placement`, which renders fleet mode with no site/building/rack and asserts the Location row is absent. This covers the `locationLines.length === 0` branch. Pushed in the latest commit.

PR #796, opened 2026-07-22
Curtailment: Curtail ETA can overstate progress while drifted-target churn starves pending dispatch waves

### LINKED ISSUE #679: Curtail ETA can overstate progress while drifted-target churn starves pending dispatch waves
## Problem

The active curtailment card's "Estimated time to curtail" (#660 / #662) can materially overstate how close completion is while a drifted target churns. The estimate models remaining work as `ceil(pending / curtail_batch_size)` waves paced by `curtail_batch_interval_sec` — but the reconciler's pacing gate can be held closed indefinitely by a single non-compliant miner, so the displayed "3 minutes" never elapses while pending targets starve.

Mechanism (verified against the reconciler during the #662 review):

- `observeActive` redispatches `drifted` targets unconditionally every tick, outside the pending batch gate (`reconciler.go` — `dispatchOneCurtail(..., models.TargetStateDrifted)`).
- Every successful redispatch restamps `CurtailPhase.DispatchedAt`, without incrementing `RetryCount` for a device that keeps succeeding-then-drifting.
- `curtailBatchIntervalElapsed` keys on the **newest** `DispatchedAt` across all desired-curtailed targets — drifted included.

With the default 30s tick and any `curtail_batch_interval_sec > 30`, one miner that accepts sleep commands but never powers down (exactly the condition the drift machinery exists for) keeps the gate closed every tick: pending waves are never dispatched, while the card shows a confident finite ETA. This violates the surface's hard constraint of never overstating SLA progress.

Two layers:

1. **Server (root cause):** drifted per-tick redispatches restamping the batch-interval clock starve pending dispatch waves. This is pre-existing reconciler behavior that #662's progress view now makes visible; consider excluding drift-path redispatches from the `curtailBatchIntervalElapsed` reference, or tracking the batch clock separately from drift retries.
2. **Client (mitigation until the server changes):** `getCurtailRemainingSeconds` (`ActiveCurtailmentStatus.tsx`) already treats `driftedCount > 0` as prior-dispatch evidence for wave charging, but still renders an uncaveated finite estimate. Options: hide the ETA while drifted targets exist, or annotate it ("may take longer while miners retry"). Drift is often transient, so a caveat may fit better than hiding — see the "start small and monitor" design direction in the #670 thread.

## Acceptance sketch

- A curtailing event with pending + persistently-drifting targets never shows an ETA that stays constant/expired while pending waves make no progress — either the pending waves actually dispatch (server fix), or the estimate hides/caveats itself (client mitigation).
- Existing behavior preserved when no drifted targets exist (first-wave-free rule, wave charging).
- Derivation/component tests cover the drifted+pending coexistence case beyond the current wave-charging rule.

## References

- Deferred from the PR #662 multi-agent review (finding #5, adversarial reviewer, independently validated against `reconciler.go`).
- Related: #661 (confirmation pulse-check) reduces confirmation latency but does not address dispatch pacing; #660 (closed by #662) shipped the ETA.
- Code anchors: `server/internal/domain/curtailment/reconciler/reconciler.go` (`observeActive` drift redispatch, `curtailBatchIntervalElapsed`), `client/src/protoFleet/features/energy/ActiveCurtailmentStatus.tsx` (`getCurtailRemainingSeconds`).

negarn:
Reviewable diff: +247/-142 across 8 files (excludes generated, test, and story files).

## Summary
This PR prevents curtailment retry recovery from starving fresh pending dispatch waves when a batch interval is configured. Operators get more accurate curtail and restore duration previews when facility fan delays are selected, and the reconciler now paces new pending curtail waves from a durable event-level dispatch clock instead of retry target timestamps.

## How it works
The database adds `curtailment_event.last_curtail_pending_dispatch_at` as the durable pacing clock for fresh pending curtail dispatches. The migration backfills active/pending events from the newest existing curtail dispatch timestamp so in-flight events keep their operator-configured pacing after deploy.

The curtailment store exposes `RecordCurtailPendingDispatch`, guarded by the loaded event state, so stale reconcilers lose the race instead of writing over an advanced event. During reconciliation, retry and orphan-recovery dispatches still run first, but they intentionally skip the pending-wave clock update. Fresh pending dispatches record the event clock only when at least one target is actually sent; if that durable write fails, the reconciler does not advance the in-memory clock. Recurtail clears the clock when a restoring event is resumed to pending so a renewed curtailment can start immediately. The ProtoFleet plan preview adds selected facility fan off/restore delays to miner batch timing so ETA copy reflects the full operator-visible sequence.

## Diagrams

```mermaid
flowchart TD
  A["Reconciler tick"] --> B["Retry or orphan recovery dispatch"]
  B --> C["Do not update pending-wave clock"]
  C --> D["Check event pending-wave clock"]
  D --> E["Dispatch eligible pending curtail batch"]
  E --> F["Persist last_curtail_pending_dispatch_at"]
  F --> G["Next tick paces fresh pending waves from event clock"]
```

```mermaid
flowchart TD
  A["Operator selects facility fans"] --> B["Plan preview"]
  B --> C["Miner batch duration"]
  B --> D["Fan off or restore delay"]
  C --> E["Displayed curtail or restore ETA"]
  D --> E
```

## Areas of the code involved

| --- | --- | --- |
| `server/migrations/000126_*` | Adds/removes `last_curtail_pending_dispatch_at` and backfills active/pending events from prior curtail dispatch timestamps. | Persists the pacing clock used by active reconcilers without bypassing pacing for in-flight events after deploy. |
| `server/sqlc/queries/curtailment.sql` | Reads the new column, adds `RecordCurtailPendingDispatch`, and clears the clock during recurtail resume. | Defines the state-guarded write boundary and renewed-curtailment lifecycle behavior for the clock. |
| `server/generated/sqlc/` | Regenerated sqlc models, query methods, and retry wrapper. | Generated - skip except to confirm it matches SQL inputs. |
| `server/internal/domain/stores/` | Adds the store interface method and SQL implementation. | Converts zero-row state races into the existing curtailment race-loss path. |
| `server/internal/domain/curtailment/models/` | Carries the new event clock through domain models. | Lets reconciliation use event-level timing instead of target-level dispatch timestamps. |
| `server/internal/domain/curtailment/reconciler/` | Separates retry dispatches from fresh pending dispatch clock updates and treats clock persistence errors as failed clock writes. | Core starvation fix; verify retry recovery no longer blocks eligible pending waves or advances only memory. |
| `client/src/protoFleet/features/energy/useCurtailmentPlanPreview.ts` | Adds selected facility fan delays to curtail and restore estimates. | Keeps operator-facing ETA copy aligned with the actual fan-plus-miner sequence. |
| `*.test.*` / `*_test.go` | Adds focused tests for retry churn, blocked recovery waves, and facility fan ETA math. | Covers the behavioral regression and preview calculation changes. |

- Use an event-level pending dispatch clock instead of scanning target `DispatchedAt` timestamps, so retries cannot reset pacing for unrelated pending work.
- Backfill active/pending events from existing target dispatch history instead of accepting `NULL` for in-flight rows, so deploy does not allow an immediate extra wave.
- Record the clock only after a successful fresh pending dispatch instead of before dispatch, so empty or fully skipped claims do not throttle later work.
- Keep retry/orphan recovery dispatches eligible while the pending-wave gate is closed, instead of serializing all curtail dispatch work behind the same interval.
- Clear the clock on recurtail resume instead of preserving the old wave timestamp, so renewed demand starts a fresh curtail sequence.
- Add facility fan delay to preview estimates in the client instead of changing server defaults, so the existing persisted scheduling semantics stay unchanged.

## Testing & validation

- Ran `just gen`.
- Ran `go test ./internal/domain/curtailment/...`.
- Ran `go test ./internal/domain/stores/sqlstores -run '^$'`.
- Ran `go test ./internal/handlers/curtailment -run 'Test.*AdminTerminate|Test.*Start|Test.*Stop|Test.*Update|Test.*List'`.
- Ran `golangci-lint run -c .golangci.yaml`.
- Ran `golangci-lint run -c .golangci.yaml ./internal/domain/curtailment/... ./internal/domain/stores/interfaces ./internal/domain/stores/sqlstores`.
- Ran `npx vitest run src/protoFleet/features/energy/useCurtailmentPlanPreview.test.ts`.
- Ran `npx tsc --noEmit`.
- Ran `npx eslint src/protoFleet/features/energy/useCurtailmentPlanPreview.ts src/protoFleet/features/energy/useCurtailmentPlanPreview.test.ts --max-warnings 0`.
- Ran `npx prettier --check src/protoFleet/features/energy/useCurtailmentPlanPreview.ts src/protoFleet/features/energy/useCurtailmentPlanPreview.test.ts`.
- Ran `npm run build:protoFleet` (Vite emitted an existing generated-CSS optimization warning for `.max-h-[var(...)]`, build succeeded).
- Attempted broad `go test ./internal/domain/curtailment/... ./internal/domain/stores/sqlstores`; the `sqlstores` integration tests could not connect to local Postgres because `127.0.0.1:5432` rejected the expected `fleet` credentials, so local validation used the compile-only sqlstores check above.

Closes #679


server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -658,6 +660,10 @@ func (r *Reconciler) dispatchCurtailBatch(ctx context.Context, ev *models.Event,
 		t.CurtailPhase.State = models.TargetStateDispatched
 		t.CurtailPhase.DispatchedAt = &now
 		t.CurtailPhase.BatchUUID = &batchID
+		wroteDispatch = true
+	}
+	if recordPendingDispatch && wroteDispatch {
+		return r.recordCurtailPendingDispatch(ctx, ev, now)
```
rongxin-liu: Worth making this fail closed before sending the command? If the target update succeeds but `RecordCurtailPendingDispatch` fails, the stale clock can let the next pending wave run on the same or next tick. Perhaps we could reserve the wave durably before dispatch, or persist the clock atomically with the successful enqueue/state update.

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -468,21 +473,17 @@ func (r *Reconciler) dispatchPendingCurtailBatches(ctx context.Context, ev *mode
...
-	if intervalActive && hadDispatching {
+	if !dispatchByState(models.TargetStateDispatching, intervalActive, skipPendingDispatchClock) {
 		return
 	}
-	if intervalActive && !r.curtailBatchIntervalElapsed(ev, targets) {
+	if intervalActive && !r.curtailBatchIntervalElapsed(ev) {
 		return
 	}
-	_ = dispatchByState(models.TargetStatePending, intervalActive)
+	_ = dispatchByState(models.TargetStatePending, intervalActive, recordPendingDispatchClock)
```
rongxin-liu: Should we distinguish a pre-command `DISPATCHING` orphan from a post-command retry here? After a crash between the pre-write and `Curtail`, recovery is the first actual send; because it skips the clock and then immediately checks the elapsed clock, a fresh `PENDING` batch can follow in the same tick. Maybe we should restore the paced early return unless durable state shows the command was already enqueued.

server/migrations/000126_add_curtailment_pending_dispatch_clock.up.sql
```diff
@@ -0,0 +1,18 @@
...
+UPDATE curtailment_event AS ce
+SET last_curtail_pending_dispatch_at = latest.dispatched_at
+FROM (
+    SELECT
+        curtailment_event_id,
+        MAX(COALESCE(curtail_dispatched_at, last_dispatched_at)) AS dispatched_at
+    FROM curtailment_target
+    WHERE COALESCE(curtail_dispatched_at, last_dispatched_at) IS NOT NULL
+      AND desired_state = 'curtailed'
+    GROUP BY curtailment_event_id
```
rongxin-liu: I think we could split the column addition and backfill into separate migrations. Since this file runs in one implicit transaction, the `ACCESS EXCLUSIVE` lock from `ADD COLUMN` is held while the update aggregates historical target rows. Perhaps `000126` could add the nullable column, with a follow-up migration that scopes the backfill to eligible live events.

rongxin-liu (approved):

PR #812, opened 2026-07-24
Curtailment: Enable end-to-end curtailment confirmation with dev-stack fake miners

### LINKED ISSUE #709: Dev-stack fake miners cannot exercise curtailment confirmation (no power reporting / no sleep power state)
## Problem

The dev stack (`just dev`) cannot exercise the curtailment reconciler's confirmation path end-to-end. Targets dispatch and go to sleep, but the fakes can never produce the telemetry evidence `isCurtailed` requires (observed power ≤ `DriftThresholdFactor` × baseline, positive evidence required on the confirm path), so miners sit in "Curtailing" for the lifetime of an event and never reach "Curtailed".

Observed on a 14-miner dev fleet during #663 / #705 testing: all 10 commandable miners dispatched correctly but none ever confirmed. On restore they resolve immediately (power never actually dropped), so the restore path masks the gap.

## Root causes (two independent fake limitations)

1. **`server/fake-antminer/` never reports power.** The antminer plugin parses `chain_power` (or numeric `power`/`Power`) from the RPC `stats` response (`plugin/antminer/pkg/antminer/client.go`, `parseWattageFromRPCStats`); the fake implements sleep mode (hashrate → 0, status → `sleeping`) but exposes no power field anywhere, so `LatestPowerW` is nil for every antminer-family container. The confirm path deliberately refuses to confirm on missing power, so these targets can never leave "Curtailing" even though the sleep command landed. Real Antminers on power-reporting firmware confirm normally.

2. **`server/fake-proto-rig/` has no sleep power state distinct from idle.** The fake models exactly two power levels (`server/fake-proto-rig/models.go`): ~3400 W mining and ~200 W (±5%) for any non-mining state. A pool-less proto rig idles at ~200 W, gets a ~200 W baseline persisted at admission (#663 targets pool-less miners and persists idle-draw baselines), and "sleeping" it does not reduce the fake's reported power — so observed power can never reach the ≤ 50%-of-baseline confirmation threshold. Real hardware sleep draw (PSU standby, tens of watts) sits well below idle draw.

## Proposed fix

- `fake-proto-rig`: model a distinct sleep/curtailed state at low power (e.g. ~30 W ± variation) separate from stopped/idle (~200 W), entered on the sleep/curtail command and exited on wake.
- `fake-antminer`: report `chain_power` in the RPC `stats` payload (e.g. ~3250 W mining, ~30 W in sleep mode) so the plugin's power parsing and the confirm path light up in dev.

Per `.claude/skills/fake-rig-fixtures`: these fakes feed `tests/plugin-contract/`, the E2E suites, and `just dev`, so behavior changes need a pass over all three surfaces (contract fixtures and E2E assertions that may pin current telemetry shapes).

## Acceptance criteria

- An all-paired curtailment on the dev stack drives commandable fakes through pending → dispatched → **confirmed**, and restore through dispatched → **resolved** via genuine power transitions (not vacuous instant-resolves).
- A pool-less fake proto rig confirms at an idle-draw baseline (~200 W baseline, ~30 W sleep draw), exercising the low-baseline confirmation semantics from #663/#705 that are currently untestable outside unit tests.
- Plugin contract tests and E2E suites updated/passing with the new telemetry shapes.

## Why it matters

Beyond dev UX, this is the only way to integration-test two documented risks from #705 (drift-flap when sleep draw exceeds half of an idle baseline, and restore confirmation at noisy idle baselines) before they are observed on real hardware.

## Reference

- Confirmation predicates: `isCurtailed` / `isRestored` in `server/internal/domain/curtailment/reconciler/reconciler.go`
- Fake sleep handling: `server/fake-antminer/models.go` (`effectiveHashRateLocked`, `summaryStatusLocked`); `server/fake-proto-rig/models.go` (idle power constant)
- Related: #663 (commandability admission), #705 (implementation; see Known Residuals)

---
🤖 Filed by rongxin-liu's AI agent.

negarn:
Reviewable diff: +165/-56 across 7 files (excludes generated, test, and story files).

## Summary

This PR lets the local dev-stack fake Proto and Antminer devices produce the power transitions required for end-to-end curtailment confirmation. Full curtailment now reports minimal draw, while restore returns each device to mining or idle power according to its pre-curtail state, including pool-less and intentionally stopped Proto devices.

https://github.com/user-attachments/assets/e6fafd46-f0a3-42ce-9288-20a5ba8c1c7c

## How it works

For Proto devices, the plugin snapshots whether the miner was active before full curtailment and marks the stop request as curtailment intent. The fake rig uses that marker to enter a distinct `Curtailed` state at approximately 30 W; ordinary stop remains an idle state at approximately 200 W. During restore, the plugin starts devices that were mining and sends an ordinary stop to devices that were not, moving them out of curtailment without incorrectly starting hash production.

For Antminers, the existing sleep/wake workflow remains the control path. The fake cgminer server now implements `stats` with the firmware metadata and mining-stat records expected by the Antminer plugin, including dynamic `chain_power` at approximately 3,250 W while mining and 30 W while sleeping. HTTP stats report the same power value for consistency.

## Diagrams

```mermaid
flowchart TD
  A["Curtailment reconciler dispatches FULL"] --> B["Miner plugin command"]
  B --> C["Proto plugin snapshots mining state"]
  B --> D["Antminer plugin selects sleep mode"]
  C --> E["POST mining/stop with curtailment marker"]
  E --> F["Fake Proto rig reports Curtailed and about 30 W"]
  D --> G["Fake Antminer enters sleep mode"]
  G --> H["cgminer stats reports chain_power about 30 W"]
  F --> I["Persisted telemetry confirms target"]
  H --> I
  I --> J["Restore dispatch"]
  J --> K["Previously mining devices start or wake"]
  J --> L["Previously idle Proto devices receive ordinary stop"]
  K --> M["Mining power returns"]
  L --> N["Idle power returns without hashing"]
```

```mermaid
stateDiagram-v2
  [*] --> Mining
  [*] --> Idle
  Mining --> Curtailed: FULL curtail
  Idle --> Curtailed: FULL curtail
  Curtailed --> Mining: restore previous mining state
  Curtailed --> Idle: restore previous non-mining state
```

## Areas of the code involved

| --- | --- | --- |
| `plugin/proto/internal/device/` | Marks full-curtail stop requests and restores both active and inactive pre-curtail states explicitly. | Defines the production command semantics used to return low-power curtailed devices to their prior state. |
| `plugin/proto/pkg/proto/` | Adds optional request headers, maps `curtailed` as healthy-inactive, and exposes the curtail-specific stop call. | Carries curtailment intent across the HTTP boundary while preserving normal stop behavior. |
| `server/fake-proto-rig/` | Separates mining, idle, and curtailed power; allows curtailment to remain visible on pool-less rigs. | Supplies truthful telemetry for both confirmation and restore without starting pool-less devices. |
| `server/fake-antminer/` | Implements cgminer `stats` and reports mode-dependent power through RPC and HTTP. | Matches the actual Antminer parser path used to populate `PowerWatts`. |
| Focused tests | Covers Proto restore intent, request marking, fake Proto power states, and Antminer RPC/HTTP power. | Pins the state transitions and parser-facing response shapes used by confirmation. |

- Reuse the existing Proto mining-stop endpoint with a backward-compatible request marker instead of adding a simulator-only endpoint; normal manual stop remains distinguishable from full curtailment.
- Issue an ordinary stop when restoring a previously non-mining Proto device instead of doing nothing; the redundant command is idempotent and moves the fake from curtailed draw back to idle draw.
- Give `Curtailed` precedence over the pool-derived `NoPools` presentation so telemetry can expose the temporary low-power state, then return to `NoPools` after restore.
- Match the Antminer plugin's existing second-record `chain_power` contract instead of introducing a fake-only power field or changing the production parser.
- Keep virtual-miner handle persistence and load-test behavior out of scope; this change targets the process-backed dev-stack fake miners.

## Testing & validation

- Passed Proto plugin formatting, `golangci-lint`, package/unit tests, and `go build ./...`.
- Passed server `golangci-lint`.
- Passed `GOWORK=off go test --count=1 .` in both `server/fake-antminer` and `server/fake-proto-rig`.
- Passed `go test --count=1 ./pkg/antminer/...` in `plugin/antminer`.
- The full Proto plugin test command reached the Docker integration package after all package/unit tests passed, but the local Docker connection closed while creating its container.
- `just test-contract` could not start because Cargo, required by its asicrs prerequisite, is not installed locally.
- The complete `just dev` curtailment lifecycle and Docker E2E suites were not run locally.

Closes #709

### DIFF
--- a/plugin/proto/internal/device/device.go
+++ b/plugin/proto/internal/device/device.go
@@ -105,19 +105,6 @@ func miningStateBeforeFullCurtail(wasMining bool) fullCurtailMiningState {
 	return fullCurtailMiningStateWasNotMining
 }
 
-func (s fullCurtailMiningState) restoreMiningDecision() (bool, bool) {
-	switch s {
-	case fullCurtailMiningStateUnknown:
-		return false, false
-	case fullCurtailMiningStateWasMining:
-		return true, true
-	case fullCurtailMiningStateWasNotMining:
-		return false, true
-	default:
-		return false, false
-	}
-}
-
 type DeviceOption func(*Device)
 
 func SetStatusTTL(ttl time.Duration) func(*Device) {
@@ -637,11 +624,23 @@ func (d *Device) curtailFull(ctx context.Context) error {
 	}
 	wasMining := isMiningHealth(metrics.Health)
 
-	if err := d.stopMining(ctx); err != nil {
+	if err := d.curtailMining(ctx); err != nil {
 		return wrapCurtailDispatchError(d.id, err)
 	}
 	d.recordFullCurtailment(wasMining)
 	d.invalidateStatusCache()
+	return nil
+}
+
+func (d *Device) curtailMining(ctx context.Context) error {
+	slog.Info("Plugin device entering full curtailment",
+		"device_id", d.id,
+		"host", d.deviceInfo.Host)
+
+	if err := d.client.CurtailMining(ctx); err != nil {
+		return fmt.Errorf("failed to stop mining for curtailment: %w", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


plugin/proto/pkg/proto/client.go
```diff
@@ -842,15 +855,19 @@ func (c *Client) GetStatus(ctx context.Context) (*Status, error) {
...
 
-	// The actual pool list is the source of truth, not MiningState (which can be stale).
+	// The actual pool list is normally the source of truth because MiningState
+	// can be stale. Explicit curtailment takes precedence so a pool-less device
+	// remains visibly inactive while it is held at minimal power.
 	needsPool, err := c.checkNeedsMiningPool(ctx)
 	if err != nil {
 		slog.Warn("failed to check pool configuration", "error", err)
-	} else if needsPool {
+	} else if needsPool && !isCurtailed {
```
rongxin-liu: Perhaps we should keep the pool list authoritative unless Proto Fleet has trusted local evidence that full curtailment is active. As written, a stale or unsolicited `Curtailed` status can hide `HealthNeedsMiningPool`. Worth passing the raw curtailed signal up and apply the inactive override only from `Device.Status`, with trusted and unsolicited cases covered?

plugin/proto/internal/device/device.go
```diff
@@ -666,40 +665,45 @@ func (d *Device) curtailEfficiency(ctx context.Context) error {
...
+	var miningRestoreErr error
+	switch miningStateToRestore {
+	case fullCurtailMiningStateUnknown:
+	case fullCurtailMiningStateWasMining:
+		miningRestoreErr = d.startMining(ctx)
+	case fullCurtailMiningStateWasNotMining:
+		miningRestoreErr = d.stopMining(ctx)
+	}
+	if miningRestoreErr != nil {
+		return wrapCurtailDispatchError(d.id, miningRestoreErr)
```
rongxin-liu, line 697: Do we want to cover a failed-then-successful restore for `WasNotMining`? The first stop may return a transient error; the retry should issue stop again and never start, which would pin that the snapshot is retained until restore succeeds.

rongxin-liu (approved):

PR #815, opened 2026-07-27
feat(firmware): add target metadata and compatible update workflows

krisztiankurucz:
Reviewable diff: +2169/-346 across 36 files (excludes generated, test, and story files).

## Summary

Firmware payloads now carry a manufacturer, model, and firmware version from upload through deployment, and updates are rejected unless every selected miner matches that target. Operators can upload, inspect, repair, reuse, and deploy target-scoped firmware from ProtoFleet or the Fleet CLI, while server-side authorization, durable metadata indexing, and post-reboot version refresh keep those workflows safe and observable.

## How it works

1. ProtoFleet derives manufacturer/model choices from fleet model groups and requires a firmware version; the Fleet CLI requires equivalent target flags.
2. Every raw HTTP mutation first checks the canonical firmware-update permission, then direct and chunked uploads validate the complete target and compute SHA-256 while staging.
3. The file service publishes each payload beside an atomic `metadata.json` sidecar. Checksum reuse becomes eligible only after the directory entry is durable; metadata edits temporarily withdraw the artifact from reuse until the replacement sidecar is synced.
4. Preflight checks and saves reuse only an exact checksum-plus-metadata match. Keyed upload locks make concurrent identical uploads converge on one artifact, while `force` creates a new copy.
5. Inventory APIs return target metadata and support metadata repair without changing the original upload timestamp. Missing artifacts return not-found errors, while malformed storage layouts surface as internal errors.
6. ProtoFleet filters saved firmware to the selected miner target, and the server independently validates every resolved device before dispatch. After a Proto device reboots, its cached status and firmware-check timestamp are cleared so the next status reports the activated version immediately.

```mermaid
flowchart LR
    U["ProtoFleet or Fleet CLI"] --> A["Authorize firmware mutation"]
    A --> H["Validate target and stage upload"]
    H --> S["Firmware file service"]
    S --> D["Durable payload plus metadata.json"]
    D --> I["Checksum plus normalized target index"]
    I --> R["Exact-match reuse or forced copy"]
    D --> L["Inventory and metadata repair"]
    L --> P["Target-filtered firmware picker"]
    P --> V["Server-side device compatibility check"]
    V --> C["Firmware command dispatch"]
    C --> B["Miner reboot and activation"]
    B --> F["Fresh firmware version on next status"]
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/infrastructure/files/` | Persists target metadata and upload time, maintains metadata-aware checksum indexes, atomically edits sidecars, classifies storage failures, and serializes identical uploads | Core durability, legacy compatibility, error semantics, and deduplication behavior |
| `server/internal/handlers/firmware/` | Validates upload/check metadata, exposes metadata editing, applies firmware-update authorization to every mutation, and records new-upload activity | HTTP validation, authorization, audit, and error-response boundary |
| `server/internal/domain/command/` | Validates resolved device manufacturer/model before firmware dispatch | Final safety boundary against incompatible deployment |
| `server/cmd/fleetcli/` | Requires target flags and carries them through direct and chunked requests | CLI contract and upload-mode parity |
| `client/src/protoFleet/api/` and `components/FirmwareUpload/` | Makes target metadata mandatory at the typed upload boundary and shares upload/version handling | Prevents incomplete first-party upload calls |
| `client/src/protoFleet/features/settings/` | Adds fleet-derived target fields, version inference, metadata columns, and metadata editing | Operator inventory and repair workflow |
| `client/src/protoFleet/features/fleetManagement/` | Restricts firmware actions to one known target and filters reusable payloads | Prevents mixed-target selections in the UI |
| `plugin/proto/internal/device/` | Invalidates both status and firmware-check caches after reboot | Ensures activation is visible without waiting for the normal firmware refresh interval |
| `server/fake-antminer/` and `server/fake-proto-rig/` | Models staged firmware activation and filename-derived version changes | End-to-end simulator fidelity |
| Test and E2E files | Cover storage durability, authorization responses, compatibility, concurrent deduplication, upload workflows, and activated-version reporting | Regression coverage; review with the corresponding production area |

- Store metadata beside filesystem-owned payloads instead of introducing a database or protobuf migration; sidecar integrity remains part of artifact validity.
- Require complete metadata for new uploads but keep legacy files readable and editable; deployment stays blocked until target metadata is repaired.
- Use exact normalized metadata for deduplication and case-insensitive manufacturer/model matching for deployment; storage identity remains precise while device compatibility tolerates casing differences.
- Remove an artifact from checksum reuse before replacing metadata and restore it only after directory sync; a failed edit may be temporarily unavailable for reuse rather than advertise state that is not durable.
- Preserve `uploaded_at` in the sidecar and fall back to the legacy directory timestamp; metadata edits do not reorder inventory.
- Serialize only identical checksum-and-target uploads with reference-counted keyed locks; unrelated uploads remain concurrent.
- Clear the firmware refresh throttle only after a successful reboot; routine status polling keeps its existing cache behavior.

## Testing & validation

- `just lint`
- `go test -race ./internal/infrastructure/files ./internal/handlers/firmware -count=1`
- `(cd plugin/proto && go test ./internal/device -count=1)`
- Targeted firmware handler, command-domain, Fleet CLI, FleetNode artifact, and simulator tests
- Targeted Vitest coverage, including `useFirmwareApi.test.ts` (35 tests passed after rebase)
- `npx tsc --noEmit`
- Playwright firmware spec discovery passed and now asserts the activated version; the full Docker-backed browser flow was not rerun because the frontend stack was unavailable
- `git diff --check`

There are no database migrations, protobuf changes, or generated files in this PR.

## Screenshots
Add firmware payload
Add firmware payload
Add firmware payload
hamware
Firmware update requires miners from the
image

### DIFF
--- a/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
+++ b/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
@@ -130,6 +130,33 @@ interface FileProcessingStatusProps {
   uploadProgress: number;
 }
 
+interface FileSelectedStatusProps {
+  fileName: string;
+  fileSize: number;
+  onRemove: () => void;
+}
+
+export function FileSelectedStatus({ fileName, fileSize, onRemove }: FileSelectedStatusProps) {
+  return (
+    <div className="flex items-center justify-between gap-4 rounded-lg border border-border-5 p-4">
+      <div className="flex min-w-0 flex-col">
+        <div className="truncate text-300 text-text-primary" title={fileName}>
+          {fileName}
+        </div>
+        <div className="text-200 text-text-primary-70">{formatFileSize(fileSize)}</div>
+      </div>
+      <Button
+        variant={variants.textOnly}
+        textColor="text-core-accent-fill"
+        textOnlyUnderlineOnHover={false}
+        className="shrink-0"
+        onClick={onRemove}
+        text="Remove"
+      />
+    </div>
+  );
+}
+
 export function FileProcessingStatus({ state, fileName, fileSize, uploadProgress }: FileProcessingStatusProps) {
   return (
     <div className="flex items-center gap-4 rounded-lg border border-border-5 p-4">
--- a/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
+++ b/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
@@ -60,16 +60,18 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
   );
 
   const handleClick = useCallback(() => {
+    if (disabled) return;
     fileInputRef.current?.click();
-  }, []);
+  }, [disabled]);
 
   const handleFileInputChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
+      if (disabled) return;
       const selected = e.target.files?.[0];
       if (selected) onFileSelect(selected);
       if (fileInputRef.current) fileInputRef.current.value = "";
     },
-    [onFileSelect],
+    [disabled, onFileSelect],
   );
 
   const formattedExtensions =
@@ -92,7 +94,8 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
         onDrop={handleDrop}
         data-testid="firmware-drop-zone"
         role="button"
-        tabIndex={0}
+        aria-disabled={disabled}
+        tabIndex={disabled ? -1 : 0}
       >
         <div className="text-300 text-text-primary">Drag update files here</div>
         <div className="text-200 text-text-primary-70">or</div>
@@ -112,6 +115,7 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
         type="file"
         accept={buildAcceptString(extensions)}
         onChange={handleFileInputChange}
+        disabled={disabled}
         className="hidden"
         data-testid="firmware-file-input"
       />
--- a/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
+++ b/client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
@@ -1,4 +1,4 @@
-import type { ChangeEvent, DragEvent } from "react";
+import type { ChangeEvent, DragEvent, KeyboardEvent } from "react";
 import { useCallback, useRef, useState } from "react";
 import clsx from "clsx";
 import { Checkmark } from "@/shared/assets/icons";
@@ -64,6 +64,15 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
     fileInputRef.current?.click();
   }, [disabled]);
 
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<HTMLDivElement>) => {
+      if (e.target !== e.currentTarget || (e.key !== "Enter" && e.key !== " ")) return;
+      if (e.key === " ") e.preventDefault();
+      handleClick();
+    },
+    [handleClick],
+  );
+
   const handleFileInputChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
       if (disabled) return;
@@ -88,6 +97,7 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
           isDragActive && "ring-2 ring-border-primary",
         )}
         onClick={handleClick}
+        onKeyDown={handleKeyDown}
         onDragEnter={handleDragEnter}
         onDragOver={handleDragOver}
         onDragLeave={handleDragLeave}
--- a/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
+++ b/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
@@ -1,0 +1,79 @@
+import { useCallback, useEffect, useMemo, useState } from "react";
+import type { MinerModelGroup } from "@/protoFleet/api/generated/fleetmanagement/v1/fleetmanagement_pb";
+import useMinerModelGroups from "@/protoFleet/api/useMinerModelGroups";
+
+const MINER_MODELS_ERROR_MESSAGE = "Couldn't load fleet miner models.";
+
+interface UseMinerTargetOptionsArgs {
+  /** Fetch model groups while true (typically the dialog's open state). */
+  active: boolean;
+  /** Currently selected manufacturer; model options are limited to it. */
+  selectedManufacturer: string;
+  /** Existing values kept selectable even when absent from the fleet. */
+  seedManufacturer?: string;
+  seedModel?: string;
+}
+
+/**
+ * Loads the fleet's miner model groups and builds the Manufacturer/Model
+ * select options shared by the firmware metadata dialogs. `reset` clears the
+ * fetched groups so the next activation refetches.
+ */
+export function useMinerTargetOptions({
+  active,
+  selectedManufacturer,
+  seedManufacturer = "",
+  seedModel = "",
+}: UseMinerTargetOptionsArgs) {
+  const { getMinerModelGroups } = useMinerModelGroups();
+  const [modelGroups, setModelGroups] = useState<MinerModelGroup[] | null>(null);
+  const [modelsError, setModelsError] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (!active || modelGroups !== null || modelsError !== null) return;
+    let cancelled = false;
+    void getMinerModelGroups(null)
+      .then((groups) => {
+        if (!cancelled) setModelGroups(groups);
+      })
+      .catch(() => {
+        if (!cancelled) {
+          setModelGroups([]);
+          setModelsError(MINER_MODELS_ERROR_MESSAGE);
+        }
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [active, getMinerModelGroups, modelGroups, modelsError]);
+
+  const reset = useCallback(() => {
+    setModelGroups(null);
+    setModelsError(null);
+  }, []);
+
+  const manufacturerOptions = useMemo(() => {
+    const manufacturers = new Set((modelGroups ?? []).map((group) => group.manufacturer.trim()).filter(Boolean));
+    if (seedManufacturer.trim()) manufacturers.add(seedManufacturer.trim());
+    return [
+      { value: "", label: "Select manufacturer" },
+      ...[...manufacturers].sort().map((manufacturer) => ({ value: manufacturer, label: manufacturer })),
+    ];
+  }, [modelGroups, seedManufacturer]);
+
+  const modelOptions = useMemo(() => {
+    const models = new Set(
+      (modelGroups ?? [])
+        .filter((group) => group.manufacturer.trim() === selectedManufacturer.trim())
+        .map((group) => group.model.trim())
+        .filter(Boolean),
+    );
+    if (seedModel.trim()) models.add(seedModel.trim());
+    return [
+      { value: "", label: "Select model" },
+      ...[...models].sort().map((model) => ({ value: model, label: model })),
+    ];
+  }, [modelGroups, selectedManufacturer, seedModel]);
+
+  return { modelGroups, modelsError, manufacturerOptions, modelOptions, reset };
+}
--- a/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
+++ b/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
@@ -9,6 +9,8 @@ interface UseMinerTargetOptionsArgs {
   active: boolean;
   /** Currently selected manufacturer; model options are limited to it. */
   selectedManufacturer: string;
+  /** Currently selected model, reconciled against the options as `resolvedModel`. */
+  selectedModel: string;
   /** Existing values kept selectable even when absent from the fleet. */
   seedManufacturer?: string;
   seedModel?: string;
@@ -18,10 +20,16 @@ interface UseMinerTargetOptionsArgs {
  * Loads the fleet's miner model groups and builds the Manufacturer/Model
  * select options shared by the firmware metadata dialogs. `reset` clears the
  * fetched groups so the next activation refetches.
+ *
+ * Because `modelOptions` narrows to the selected manufacturer and arrives
+ * asynchronously, `selectedModel` can fall out of the list. This hook owns that
+ * invariant: use `resolvedModel` rather than the raw selection so a stale model
+ * is never submitted or rendered as a phantom option.
  */
 export function useMinerTargetOptions({
   active,
   selectedManufacturer,
+  selectedModel,
   seedManufacturer = "",
   seedModel = "",
 }: UseMinerTargetOptionsArgs) {
@@ -75,5 +83,10 @@ export function useMinerTargetOptions({
     ];
   }, [modelGroups, selectedManufacturer, seedModel]);
 
-  return { modelGroups, modelsError, manufacturerOptions, modelOptions, reset };
+  const resolvedModel = useMemo(
+    () => (modelOptions.some((option) => option.value === selectedModel) ? selectedModel : ""),
+    [modelOptions, selectedModel],
+  );
+
+  return { modelGroups, modelsError, manufacturerOptions, modelOptions, resolvedModel, reset };
 }
--- a/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
+++ b/client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
@@ -76,12 +76,14 @@ export function useMinerTargetOptions({
         .map((group) => group.model.trim())
         .filter(Boolean),
     );
-    if (seedModel.trim()) models.add(seedModel.trim());
+    const selectedManufacturerKey = selectedManufacturer.trim().toLowerCase();
+    const seedManufacturerKey = seedManufacturer.trim().toLowerCase();
+    if (seedModel.trim() && selectedManufacturerKey === seedManufacturerKey) models.add(seedModel.trim());
     return [
       { value: "", label: "Select model" },
       ...[...models].sort().map((model) => ({ value: model, label: model })),
     ];
-  }, [modelGroups, selectedManufacturer, seedModel]);
+  }, [modelGroups, seedManufacturer, seedModel, selectedManufacturer]);
 
   const resolvedModel = useMemo(
     () => (modelOptions.some((option) => option.value === selectedModel) ? selectedModel : ""),
--- a/plugin/proto/internal/device/device.go
+++ b/plugin/proto/internal/device/device.go
@@ -305,13 +305,13 @@ func (d *Device) refreshFirmwareVersion(ctx context.Context, metrics *sdk.Device
 	if time.Since(d.lastFirmwareCheckAt) < firmwareRefreshInterval {
 		return
 	}
-	d.lastFirmwareCheckAt = time.Now()
 	fwVersion, err := d.client.GetFirmwareVersion(ctx)
 	if err != nil {
 		slog.Debug("failed to get firmware version during Status", "error", err)
 		return
 	}
 	if fwVersion != "" {
+		d.lastFirmwareCheckAt = time.Now()
 		d.deviceInfo.FirmwareVersion = fwVersion
 		metrics.FirmwareVersion = fwVersion
 	}
--- a/plugin/proto/internal/device/device.go
+++ b/plugin/proto/internal/device/device.go
@@ -306,12 +306,12 @@ func (d *Device) refreshFirmwareVersion(ctx context.Context, metrics *sdk.Device
 		return
 	}
 	fwVersion, err := d.client.GetFirmwareVersion(ctx)
+	d.lastFirmwareCheckAt = time.Now()
 	if err != nil {
 		slog.Debug("failed to get firmware version during Status", "error", err)
 		return
 	}
 	if fwVersion != "" {
-		d.lastFirmwareCheckAt = time.Now()
 		d.deviceInfo.FirmwareVersion = fwVersion
 		metrics.FirmwareVersion = fwVersion
 	}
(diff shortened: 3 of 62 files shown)


client/src/protoFleet/components/FirmwareUpload/FirmwareUploadComponents.tsx
```diff
@@ -92,7 +94,8 @@ export function FileDropZone({ extensions, onFileSelect, disabled }: FileDropZon
         onDrop={handleDrop}
         data-testid="firmware-drop-zone"
         role="button"
-        tabIndex={0}
+        aria-disabled={disabled}
+        tabIndex={disabled ? -1 : 0}
       >
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Add an `onKeyDown` handler to trigger `handleClick()` for Enter/Space (and prevent default for Space).

Addressed in `e7e1b62bd`: the drop zone now opens the file picker with Enter or Space, prevents Space scrolling, and ignores bubbled key events from the nested button to avoid double activation. Regression tests cover keyboard activation, nested-button bubbling, and the disabled state.

server/internal/domain/command/service.go
```diff
@@ -1551,6 +1560,52 @@ func (s *Service) FirmwareUpdate(ctx context.Context, deviceSelector *pb.DeviceS
...
+	devices []resolvedDevice,
+	firmwareFileID string,
+) error {
+	metadata, err := s.filesService.GetFirmwareMetadata(firmwareFileID)
+	if err != nil {
+		return fleeterror.NewFailedPreconditionError("firmware target metadata is unavailable; re-upload the firmware before deploying it")
+	}
+	if err := files.ValidateFirmwareMetadata(metadata); err != nil {
+		return fleeterror.NewFailedPreconditionError("firmware target metadata is unknown; re-upload the firmware before deploying it")
+	}
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> `GetFirmwareMetadata` errors ... are all mapped to a generic FailedPrecondition ... and the follow-up message is misleading now that metadata can be repaired.

Addressed in `e7e1b62bd`: metadata lookup errors now retain their existing invalid-argument, not-found, or internal codes, while blank legacy metadata remains a failed precondition with guidance to repair its metadata. Regression tests cover all four cases.

server/internal/infrastructure/files/firmware.go
```diff
@@ -136,96 +249,224 @@ func (s *Service) ValidateFirmwareFile(filename string, size int64) error {
...
+	}
+	if err := s.syncFirmwareDir(stagingDir); err != nil {
+		return "", fleeterror.NewInternalErrorf("failed to sync firmware staging directory: %v", err)
+	}
+	if err := os.Rename(stagingDir, finalDir); err != nil {
+		return "", fleeterror.NewInternalErrorf("failed to publish firmware directory: %v", err)
+	}
+	removeStaging = false
+	if err := s.syncFirmwareDir(firmwareDir); err != nil {
+		return "", fleeterror.NewInternalErrorf("failed to sync firmware directory: %v", err)
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Roll back the published directory when its sync fails.

Addressed in `e99e1d578`: a failed parent-directory sync now removes the published firmware directory and syncs the parent again before returning. The original sync error is preserved, rollback failures are reported, and a regression test verifies no visible firmware or checksum-index entry leaks.

krisztiankurucz:
🤖

> Firmware metadata and reuse are not scoped to the authenticated organization.

Acknowledged. This is a real broader boundary in Fleet's existing global firmware store, and organization ownership needs to cover the full firmware lifecycle consistently rather than be patched into only the new metadata path. This PR intentionally preserves the existing storage ownership semantics while adding target metadata and compatibility enforcement; organization-owned firmware is deferred to a dedicated follow-up.

plugin/proto/internal/device/device.go
```diff
@@ -760,6 +760,13 @@ func (d *Device) invalidateStatusCache() {
 	d.clearStatusCacheLocked()
 }
 
+func (d *Device) invalidateStatusAndFirmwareCaches() {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.clearStatusCacheLocked()
+	d.lastFirmwareCheckAt = time.Time{}
+}
```
rongxin-liu, line 768: Since a firmware-update reboot lands here to force a prompt version re-read, the next `Status()` poll effectively gets one shot: `refreshFirmwareVersion` stamps `lastFirmwareCheckAt` *before* the probe, so if the device is still rebooting (or answers with the pre-update version), the 5-minute throttle re-arms and telemetry keeps showing the old version.

Maybe worth moving the timestamp update to after a successful non-empty read, so a failed probe retries on the next poll? This would matter for the upcoming rollout feature too, where post-reboot version verification is on the critical path. A small test that fails the first post-reboot probe would pin the behavior.

server/cmd/fleetcli/firmware.go
```diff
@@ -283,74 +292,119 @@ func buildFirmwareDeployRequest(ctx context.Context, cmd *cli.Command, client *C
...
+	Manufacturer string
+	Model        string
+	Version      string
+}
+
+func firmwareTargetFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{Name: "target-manufacturer", Usage: "Firmware target manufacturer/product"},
+		&cli.StringFlag{Name: "target-model", Usage: "Firmware target miner model"},
+		&cli.StringFlag{Name: "firmware-version", Usage: "Firmware version reported after a successful install"},
```
rongxin-liu, line 354: The PR description mentions repair from ProtoFleet or the Fleet CLI, but there doesn't seem to be a CLI counterpart for the new PATCH metadata endpoint yet — and repair is the only way to make a legacy (pre-metadata) file deployable again.

Should we add a `firmware edit-metadata --firmware-file-id ...` subcommand reusing these flags, plus a small `Client.FirmwareUpdateMetadata` PATCH wrapper in `firmware_client.go`? Alternatively, softening the PR description would also work if CLI parity is intentionally deferred.

server/internal/handlers/firmware/handler.go
```diff
@@ -193,21 +222,27 @@ func (h *uploadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	const multipartOverhead int64 = 1 * 1024 * 1024 // 1 MB
-	r.Body = http.MaxBytesReader(w, r.Body, h.maxUploadBytes+multipartOverhead)
+	r.Body = http.MaxBytesReader(w, r.Body, h.filesService.MaxFirmwareFileSize()+multipartOverhead)
 
-	filename, fileReader, err := extractMultipartFile(r)
+	upload, err := extractMultipartFile(r, h.filesService)
 	if err != nil {
 		writeError(w, http.StatusBadRequest, err.Error())
 		return
 	}
```
rongxin-liu, line 236: This branch maps every `extractMultipartFile` error to a 400 with the raw error text — but staging now happens inside extract, and `StageFirmwareUpload` returns internal errors for disk failures (create/copy/close). A disk-full incident during upload would be reported as a client error and never reach the logs.

Perhaps route this through the same `isClientError` + `slog.Error` pattern used for the save call just below? Extract's own client-input errors (bad content type, duplicate file field, etc.) would need wrapping as invalid-argument for the classification to hold.

server/internal/infrastructure/files/firmware.go
```diff
@@ -242,13 +497,92 @@ func (s *Service) GetFirmwareFilePath(fileID string) (string, error) {
...
+	// its replacement metadata is not yet durable.
+	s.removeFirmwareChecksumEligibility(checksum, canonical)
+	if err := writeFirmwareMetadata(dir, metadata, uploadedAt); err != nil {
+		return err
+	}
+	if err := s.syncFirmwareDir(dir); err != nil {
+		return fleeterror.NewInternalErrorf("failed to sync firmware directory: %v", err)
+	}
+	s.rememberFirmwareChecksum(checksum, canonical)
+	return nil
```
rongxin-liu: Both error returns here exit after `removeFirmwareChecksumEligibility` but before the restoring `rememberFirmwareChecksum`, so a failed edit leaves the file out of the reuse index until the next restart — repeated failures (e.g. during disk pressure) would silently disable dedup for those files.

Maybe a `restored` flag with a deferred re-advertise guard would keep the withdraw/restore pair balanced on all paths? Re-advertising looks safe either way: on a failed write the old sidecar still stands, and on a failed sync the new one is already visible to readers.

server/internal/handlers/firmware/handler.go
```diff
@@ -352,25 +443,108 @@ func (h *listFilesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// NewUpdateMetadataHandler returns an http.Handler that updates a stored
+// firmware file's deployment metadata.
+func NewUpdateMetadataHandler(
+	filesService *files.Service,
+	sessionService *session.Service,
+	userStore interfaces.UserStore,
+	permissionResolver effectivePermissionResolver,
+) http.Handler {
```
rongxin-liu, line 470: Uploads now log an activity event, but metadata edits don't — and after this PR the target label is what decides which devices a payload can be dispatched to, so a retarget feels like something operators would want in the activity trail (especially once rollout history builds on it).

Worth passing `activitySvc` here and emitting a `firmware_metadata_updated` event with the before/after target, mirroring `logFirmwareUploadActivity`?

client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
```diff
@@ -0,0 +1,92 @@
...
+  }, []);
+
+  const manufacturerOptions = useMemo(() => {
+    const manufacturers = new Set((modelGroups ?? []).map((group) => group.manufacturer.trim()).filter(Boolean));
+    if (seedManufacturer.trim()) manufacturers.add(seedManufacturer.trim());
+    return [
+      { value: "", label: "Select manufacturer" },
+      ...[...manufacturers].sort().map((manufacturer) => ({ value: manufacturer, label: manufacturer })),
+    ];
+  }, [modelGroups, seedManufacturer]);
```
rongxin-liu, line 70: Options here come exclusively from fetched model groups, so on a fleet with no paired miners the selects stay empty, Upload stays disabled, and the dialog gives no hint why — while the settings page's empty-state copy invites uploading firmware before deploying.

Since the server and CLI accept arbitrary strings for these fields, could we perhaps allow free-text entry as a fallback, or at least render an explanatory empty state ("No paired miner models found...")? Even a small message would avoid the silent dead-end on first-run fleets.

rongxin-liu (approved):
Thanks for the PR. Looks good to me overall!

plugin/proto/internal/device/device.go
krisztiankurucz: 🤖

> `refreshFirmwareVersion` stamps `lastFirmwareCheckAt` before the probe, so if the device is still rebooting, the 5-minute throttle re-arms.

Addressed in `a6b956eec`: the throttle now advances only after a successful, non-empty firmware-version response. A regression test verifies that a failed first post-reboot probe is retried by the next `Status()` call.

server/internal/handlers/firmware/handler.go
krisztiankurucz: 🤖

> A disk-full incident during upload would be reported as a client error and never reach the logs.

Addressed in `a6b956eec`: malformed multipart input remains a client-facing 400, while internal staging/filesystem failures are logged and return a generic 500. Focused tests cover both malformed input and staging failure.

server/internal/infrastructure/files/firmware.go
krisztiankurucz: 🤖

> Both error returns here exit after `removeFirmwareChecksumEligibility` but before the restoring `rememberFirmwareChecksum`.

Addressed differently in `a6b956eec`: checksum eligibility is restored when the atomic metadata write fails, because the unchanged sidecar remains authoritative. A directory-sync failure still leaves the artifact ineligible so nondurable metadata is not advertised for reuse.

server/internal/handlers/firmware/handler.go
krisztiankurucz: 🤖

> Worth passing `activitySvc` here and emitting a `firmware_metadata_updated` event with the before/after target?

Addressed in `a6b956eec`: successful metadata edits now emit a best-effort `firmware_metadata_updated` event with the authenticated actor, firmware ID, and previous/current target metadata. Failed edits emit no event, and corrupt legacy metadata remains repairable.

server/cmd/fleetcli/firmware.go
krisztiankurucz: 🤖

> Should we add a `firmware edit-metadata --firmware-file-id ...` subcommand reusing these flags, plus a small `Client.FirmwareUpdateMetadata` PATCH wrapper?

Addressed in `41e455184`: added the handwritten `firmware edit-metadata` command with required file ID and target validation, an escaped JSON PATCH client wrapper, useful JSON confirmation output, focused unit coverage, and metadata-edit/list verification in the Fleet CLI E2E workflow.

client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
krisztiankurucz: 🤖

> With no paired miners the selects stay empty, Upload stays disabled, and the dialog gives no hint why.

Addressed in `0845c9ce8`: after model groups load successfully with no usable targets, the upload dialog now explains that a paired miner is required. Loading and load-error states remain authoritative.

client/src/protoFleet/features/settings/components/useMinerTargetOptions.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const modelOptions = useMemo(() => {
+    const models = new Set(
+      (modelGroups ?? [])
+        .filter((group) => group.manufacturer.trim() === selectedManufacturer.trim())
+        .map((group) => group.model.trim())
+        .filter(Boolean),
+    );
+    if (seedModel.trim()) models.add(seedModel.trim());
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Only retain the seeded model while the selected manufacturer still matches the seeded manufacturer.

Addressed in `0845c9ce8`: the original file metadata is now the immutable seed, and its model remains available only while the normalized selected manufacturer matches the original manufacturer. Changing manufacturers clears the resolved model and prevents selecting or saving the previous model.

server/internal/infrastructure/files/firmware.go
```diff
@@ -242,13 +497,97 @@ func (s *Service) GetFirmwareFilePath(fileID string) (string, error) {
...
+	if !ok {
+		checksum, err = computeFileChecksum(filePath)
+		if err != nil {
+			return fleeterror.NewInternalErrorf("failed to compute firmware checksum: %v", err)
+		}
+	}
+
+	// Keep the cached checksum, but stop advertising this file for reuse while
+	// its replacement metadata is not yet durable.
+	wasEligible := s.removeFirmwareChecksumEligibility(checksum, canonical)
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Removing eligibility does not invalidate the in-flight snapshot, so metadata replacement must be synchronized with checksum lookups/uploads or those lookups must revalidate before returning.

Addressed in `4f6ced423`: checksum lookups now hold a read lock across the index snapshot and sidecar validation, while metadata updates hold the write lock through eligibility withdrawal, atomic replacement, directory sync, and re-indexing or restoration. A deterministic concurrency test verifies that a lookup blocks during replacement, rejects the old target, and accepts the new target.

server/internal/infrastructure/files/firmware.go
```diff
@@ -242,13 +497,97 @@ func (s *Service) GetFirmwareFilePath(fileID string) (string, error) {
...
+	if err := writeFirmwareMetadata(dir, metadata, uploadedAt); err != nil {
+		// The atomic metadata publish did not complete, so the previous sidecar
+		// remains authoritative and can safely be advertised again.
+		if wasEligible {
+			s.rememberFirmwareChecksum(checksum, canonical)
+		}
+		return err
+	}
+	if err := s.syncFirmwareDir(dir); err != nil {
+		return fleeterror.NewInternalErrorf("failed to sync firmware directory: %v", err)
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Restore the previous sidecar/index entry or otherwise reconcile the published metadata before returning.

Declined for this PR: a failed directory `fsync` means the renamed sidecar is visible but not known durable. Re-indexing the new metadata could advertise state that disappears after a crash, while restoring the old index could disagree with the currently visible sidecar. The current behavior intentionally fails closed, and `TestUpdateFirmwareMetadata_DoesNotIndexMetadataBeforeDirectorySync` pins that durability boundary. A retry or service restart reconciles the entry; a fully durable rollback would require a second atomic publish and separate failure handling beyond this change.

plugin/proto/internal/device/device.go
```diff
@@ -305,13 +305,13 @@ func (d *Device) refreshFirmwareVersion(ctx context.Context, metrics *sdk.Device
...
 		return
 	}
-	d.lastFirmwareCheckAt = time.Now()
 	fwVersion, err := d.client.GetFirmwareVersion(ctx)
 	if err != nil {
 		slog.Debug("failed to get firmware version during Status", "error", err)
 		return
 	}
 	if fwVersion != "" {
+		d.lastFirmwareCheckAt = time.Now()
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Keep failed firmware probes throttled

Addressed in `681449d98`: every firmware-version probe now advances the refresh timestamp, including errors and empty responses. Reboot still clears the timestamp for one immediate refresh, and a regression test verifies a failed post-reboot probe is retried only after the normal interval.

server/internal/infrastructure/files/firmware.go
```diff
@@ -329,17 +698,60 @@ func (s *Service) rememberFirmwareChecksum(checksum, canonicalID string) {
...
+				continue
+			}
+			slog.Warn("skipping firmware with invalid metadata during checksum lookup", "file_id", fileID, "error", err)
+			s.mu.Lock()
+			s.removeFirmwareChecksumLocked(sha256Hex, fileID)
+			s.mu.Unlock()
+			continue
+		}
+		if storedMetadata.matches(metadata) {
+			return fileID, true
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Serialize checksum reuse with firmware deletion

Addressed in `8e0cfa1e0`: firmware deletion now holds the same lifecycle lock as checksum reuse, so a lookup either returns an artifact that still exists at the method boundary or observes its completed deletion. A concurrency regression test covers the lock boundary.

krisztiankurucz:
🤖

Addressed both findings from the latest security review:

- `681449d98` records every firmware-version probe attempt, preserving the refresh throttle on errors and empty responses.
- `8e0cfa1e0` serializes firmware deletion with checksum reuse so lookups cannot return an ID removed mid-lookup.

Both fixes include focused regression coverage.

server/internal/handlers/firmware/handler.go
```diff
@@ -352,25 +452,127 @@ func (h *listFilesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		case fleeterror.IsInvalidArgumentError(err):
+			writeError(w, http.StatusBadRequest, err.Error())
+		default:
+			slog.Error("failed to update firmware metadata", "file_id", fileID, "error", err)
+			writeError(w, http.StatusInternalServerError, "failed to update firmware metadata")
+		}
+		return
+	}
+
+	currentMetadata, err := h.filesService.GetFirmwareMetadata(fileID)
```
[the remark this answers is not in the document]
krisztiankurucz: 🤖

> Capture audit metadata atomically with the edit

Addressed in `2ac1ca101`: metadata updates now return the previous and normalized current values captured inside the serialized update, and the handler logs that exact pair.

krisztiankurucz:
🤖

> Multipart uploads write large files before rejecting invalid metadata or filename

Addressed differently in `2ac1ca101`: the one-pass multipart flow still accepts metadata after the file, but filenames are now validated immediately when the file part is encountered, and a complete metadata set supplied before the file is validated before any payload bytes are staged. Final service validation remains in place for file-first and mixed-order requests.

server/internal/domain/command/service.go
```diff
@@ -1551,6 +1560,52 @@ func (s *Service) FirmwareUpdate(ctx context.Context, deviceSelector *pb.DeviceS
 	return result, nil
 }
 
+func (s *Service) validateFirmwareUpdateTargets(
+	ctx context.Context,
+	organizationID int64,
+	devices []resolvedDevice,
+	firmwareFileID string,
+) error {
+	metadata, err := s.filesService.GetFirmwareMetadata(firmwareFileID)
```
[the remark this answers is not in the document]
krisztiankurucz: > hold the metadata stable through dispatch or revalidate it under the same lifecycle synchronization immediately before enqueueing.

🤖 Addressed in `1c1338030`: firmware dispatch now leases the validated metadata under the lifecycle read lock through batch persistence and queue enqueue. A deterministic concurrency regression verifies metadata edits remain blocked throughout validation and dispatch.

server/internal/infrastructure/files/firmware.go
```diff
@@ -136,96 +256,238 @@ func (s *Service) ValidateFirmwareFile(filename string, size int64) error {
...
+		return "", fleeterror.NewInternalErrorf("failed to close staged firmware file: %v", err)
+	}
+
+	if err := writeFirmwareMetadata(stagingDir, metadata, time.Now().UTC()); err != nil {
+		return "", err
+	}
+	if err := s.syncFirmwareDir(stagingDir); err != nil {
+		return "", fleeterror.NewInternalErrorf("failed to sync firmware staging directory: %v", err)
+	}
+	if err := os.Rename(stagingDir, finalDir); err != nil {
```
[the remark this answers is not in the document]
krisztiankurucz: > publication/indexing/return must be serialized with deletion or revalidated while deletion is excluded.

🤖 Addressed in `1c1338030`: new firmware publication now holds the lifecycle read lock after deduplication through durable directory publication, checksum indexing, and return. A deterministic concurrency regression verifies deletion cannot complete while a newly published directory is still being finalized.

PR #836, opened 2026-07-30
feat(updates): add host updater engine

mcharles-square:
Reviewable diff: +3633/-0 across 5 files (excludes generated, test, and story files).

## Summary

Adds the dormant host updater engine that will power one-click upgrades: a single-flight daemon, durable operation state, and a narrow Unix-socket HTTP interface. Its trigger contract accepts optional caller-generated operation IDs, making same-ID retries idempotent while retaining target-only compatibility for older callers. It stages and preflights a release while Fleet remains online, then crosses a marker-protected activation boundary with bounded commands, crash recovery, and coordinated updater self-restart. The binary is not packaged, installed, or connected to Fleet API in this PR, so merging it exposes no operator-facing behavior.

*Stack:* The update-notification foundation [#841](https://github.com/block/proto-fleet/pull/841) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) is merged. The one-click phase is [#835](https://github.com/block/proto-fleet/pull/835) (merged) → **#836** → [#837](https://github.com/block/proto-fleet/pull/837) → [#838](https://github.com/block/proto-fleet/pull/838) → [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840). This is **2/6** and its diff is relative to `main`; #835 supplies release metadata plus the preflight/activation contract. Fleet API bridging, packaging, installation, and client exposure remain in #837–#840.

## How it works

At startup, the daemon validates and canonicalizes every component of its state, install, executable, and socket trust boundaries before loading or binding anything. The install chain may contain root plus one consistent deployment-admin UID—the owner #839 has already verified controls the same rootful Docker daemon—while the updater executable remains root/daemon-controlled. It then takes a lifetime advisory lock, reconciles only activation swaps proven by a durable marker, and removes crash-abandoned artifacts whose exact names are derived from the operation ID. Operation logs are confined and retained within fixed count/byte bounds; the local API refuses live or ambiguous socket owners and explicitly sets the bound socket identity and mode. A trigger accepts only a stable or RC tag, binds an optional canonical operation ID to that target, deduplicates concurrent same-ID admission, and constructs artifact URLs from the fixed Proto Fleet GitHub Releases origin.

The manager downloads the archive and SHA-256 sidecar, verifies transfer integrity, safely extracts bounded regular-file content, retains the updater payload through a private file descriptor, and preserves deployment configuration including the operator-owned HA node identity. Commands have phase deadlines, whole-process-group cancellation, and bounded output. After preflight, the staged tree and its parent are synchronized under a cancellable activation-length preparation deadline while the operation is still in `Preflight`; only then can the marker-backed, non-cancelable swap begin. Shutdown cancels pre-activation work and waits for terminal state before releasing the daemon lock.

After #835's preflight succeeds with Fleet still online and the staged tree is durable, the manager atomically persists an activation marker and rotates `deployment` to `deployment.previous` before installing the new tree. Pre-command failures are reconciled before terminal state is written; a restart restores the previous tree only when the marker proves the two-rename gap, never merely because an old backup exists. Success and failure become terminal only after that exact transition is durable; if the state write fails, memory retains the last nonterminal recovery context and blocks retriggers until startup reconciliation. Once Fleet is healthy, the protected updater candidate must execute `--version` successfully on the host and report the exact target tag. Before replacement, the manager hard-links the working executable into a retained rollback slot and durably writes a protected sibling handoff marker. The matching argv identifies the first candidate attempt; returned startup errors restore `.previous`, while a signal, kill, or power loss leaves the marker for the next supervisor start to reconcile before manager initialization. Secured-socket readiness durably removes the marker. Failed `exec`, missing destination entries, and restore/marker-clear crash windows are idempotently recoverable without consuming the backup; ordinary starts and post-ready failures cannot downgrade the updater.

```mermaid
flowchart LR
  C["Local Unix-socket caller"] --> A["Bounded updater API"]
  A --> M["Single-flight manager"]
  M --> G["Pinned GitHub Releases origin"]
  M --> S["Durable state, marker, lock, and retained logs"]
  M --> D["Staged, active, and previous deployments"]
  M --> R["run-fleet.sh preflight and activation"]
  M --> X["Verified replacement and durable startup handoff"]
```

```mermaid
sequenceDiagram
  participant C as Local caller
  participant U as Host updater
  participant G as GitHub Releases
  participant R as run-fleet.sh

  U->>U: Acquire daemon lock
  U->>U: Reconcile marker-proven swap, then clean stale artifacts
  C->>U: POST /v1/upgrade with target tag
  U->>G: Download archive and SHA-256 sidecar
  U->>U: Verify, safely extract, and preserve configuration
  U->>R: --preflight-only with deadline

  alt Preflight fails or shutdown begins
    U->>U: Bounded cleanup and persisted failure
  else Preflight passes
    U->>U: Sync staged tree under cancellable deadline
    U->>U: Persist swap marker and activating state
    U->>U: Fsync and rotate deployment directories
    alt Pre-command swap fails or process stops
      U->>U: Reconcile only the marker-proven layout
    else Swap is durable
      U->>U: Persist recovery command and consume marker
      U->>R: --skip-build with activation deadline
      alt Fleet fails to become healthy
        U->>U: Keep forward deployment and recovery details
      else Fleet is healthy
        U->>U: Smoke-test updater and retain rollback binary
        U->>U: Persist handoff marker, replace updater, and persist success
        U->>U: Drain API, release lock, and exec with one-shot handoff
        alt Exec fails, startup returns, or restart finds pending marker
          U->>U: Restore previous updater and exit nonzero
        else Manager initializes and secured socket binds
          U->>U: Consume rollback eligibility
        end
      end
    end
  end

  C->>U: GET /v1/status
  U-->>C: Persisted operation status
```

## Areas of the code involved

|---|---|---|
| `server/internal/updaterapi` | Transport-neutral operation and error types | Defines the complete local privilege-boundary contract |
| `server/internal/updater/manager.go` | Validation, download, staging, marker-gated activation recovery, locking, deadlines, confined log retention, exact stale-artifact cleanup, and self-refresh | Primary security, durability, and lifecycle review surface |
| `server/internal/updater/self_update_handoff.go` | Protected sibling marker plus idempotent executable restoration/commit state machine | Review crash ordering across replacement, supervisor restart, and secured-socket readiness |
| `server/internal/updater/server.go` | Bounded HTTP API plus validated socket-directory ownership and explicit socket identity/mode | Review request limits, local filesystem authentication, and duplicate-daemon behavior |
| `server/cmd/fleet-updater` | Signal handling, graceful draining, production configuration, and canonical replacement-process exec | Pins production downloads and coordinates listener/operation/lock handoff without re-resolving configured symlinks |
| `server/internal/updater/*_test.go` | Adversarial archive, activation, locking, socket, deadline, cleanup, process-group, and recovery coverage | Tests — review alongside the manager invariants |

- **Publisher trust:** Production is pinned to the official GitHub Releases origin. The sidecar detects corruption, not an independently compromised publisher; independent signing needs a separate release-key custody, rotation, revocation, and bootstrap design rather than a same-origin key added only here.
- **Activation provenance:** A file-fsynced, atomically installed marker exists only around the two-rename/pre-command window. Automatic restoration requires that marker; terminal state or the mere presence of `deployment.previous` is insufficient because forward migrations may already exist.
- **Crash cleanup:** Operation artifacts use exact SHA-256-derived direct-child names. Startup and pre-trigger sweeps run under the daemon lock, after layout reconciliation, unlink non-directories without following them, and recursively remove only exact reserved staging directories.
- **Forward recovery:** Once the activation command begins, shutdown waits instead of canceling because `run-fleet.sh` may have applied forward-only migrations. A 45-minute deadline is the liveness bound, and failed starts keep the new deployment plus an actionable recovery command.
- **Daemon serialization:** A host-state advisory lock is acquired before state load and held through operation drain; socket cleanup occurs only after proving `ECONNREFUSED` on an existing socket.
- **Filesystem trust:** State, install, executable, and socket chains reject unrelated owners and unsafe write modes, validate configured plus resolved paths, and continue through canonical paths. Install content may be owned by root plus one consistent rootful-Docker admin; state/socket/executable paths remain root/daemon-controlled. Protected final nodes cannot be symlinks, and the bound socket receives the daemon's explicit UID/GID and mode `0660`.
- **Durable completion:** Success or failure is published as terminal—and can make the manager reusable—only after that exact state is durable. A write failure restores the last nonterminal state and recovery command, blocking another upgrade or self-restart until startup reconciliation.
- **Release asset contract:** #838 builds the versioned updater into `deployment/updater/proto-fleet-updater` and publishes a `.sha256` sidecar for every architecture archive; #839 installs that payload and service. The production-bundle contract test belongs with that packaging workflow rather than this dormant engine.
- **Resource bounds:** Preflight, staged-tree durability, activation, and cleanup have bounded phases; cancellation kills the command process group and is checked between file/directory syncs before the non-cancelable swap. An individual kernel `fsync` remains inherently non-preemptible. Command output is capped at 64 MiB, while confined retention keeps the current plus newest prior logs within eight files and 256 MiB.
- **Self-refresh ordering:** The protected updater payload replaces the installed binary only after Fleet is healthy and the host kernel executes its exact-version smoke test. A hard-linked previous binary and fsynced sibling marker are installed before replacement; a canonical argv value distinguishes the first candidate attempt, and the marker survives signals, kills, power loss, missing destination entries, and supervisor restarts. Restoration retains the backup and is idempotent across crashes. Only real manager initialization plus secured production-socket readiness durably removes rollback eligibility—without adding a second supervisor or a duplicate startup probe.
- **Execution privilege:** Supported one-click installs already grant the deployment owner rootful Docker administration, which is host-administrator-equivalent. Any future relaxation of #839's eligibility gate must revisit this boundary.
- **Downstream service lifecycle:** #839 must set systemd `TimeoutStopSec` above the 45-minute activation bound (with margin or infinity) so service shutdown preserves this PR's graceful-wait contract.

## Testing & validation

- `DB_PASSWORD=fleet GOWORK=off go test ./...`
- `GOWORK=off go test ./cmd/fleet-updater ./internal/updater`
- `GOWORK=off go test -race ./cmd/fleet-updater ./internal/updater`
- `GOWORK=off go vet ./cmd/fleet-updater ./internal/updater`
- `golangci-lint run -c .golangci.yaml`
- Linux `amd64` static build and `git diff --check`
- Coverage includes traversal/size rejection, checksum failure, preserved secrets/TLS/HA node configuration and restrictive mode, failed-preflight cleanup deadlines, cancellable/deadline-bounded staged-tree durability before marker creation, marker write atomicity, pre-command rollback provenance, terminal-state reconciliation, crash-abandoned artifact cleanup/confinement, state/install/deployment/executable/socket ownership and mode policies, symlink rejection/canonical retarget resistance, canonical self-exec signaling, explicit socket identity, symlink-safe log confinement and count/byte retention, queued/terminal-state persistence failures, daemon-lock retention, live/stale socket handling, shutdown races, activation deadlines, real descendant-process termination, candidate version/format rejection, protected handoff/argv mismatch rejection, pre-rename and post-restore crash reconciliation, missing executable restoration, readiness commit, degraded self-refresh, and supervisor restart handoff.
- Fleet API integration, systemd packaging, installer eligibility, and browser behavior are intentionally deferred to #837–#840.

### DIFF
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -54,9 +54,13 @@ func run() error {
 		fmt.Println(version)
 		return nil
 	}
+	selfUpdateStartup, err := updater.PrepareSelfUpdateStartup(*selfUpdatePath, *selfUpdateHandoff)
+	if err != nil {
+		return fmt.Errorf("prepare updater startup: %w", err)
+	}
 	absoluteInstallRoot, err := filepath.Abs(*installRoot)
 	if err != nil {
-		return fmt.Errorf("resolve install root: %w", err)
+		return handleSelfUpdateStartupFailure(selfUpdateStartup, fmt.Errorf("resolve install root: %w", err))
 	}
 	manager, err := updater.NewManager(updater.Config{
 		InstallRoot:    absoluteInstallRoot,
@@ -64,7 +68,7 @@ func run() error {
 		SelfUpdatePath: *selfUpdatePath,
 	})
 	if err != nil {
-		return handleSelfUpdateStartupFailure(*selfUpdateHandoff, fmt.Errorf("initialize updater: %w", err))
+		return handleSelfUpdateStartupFailure(selfUpdateStartup, fmt.Errorf("initialize updater: %w", err))
 	}
 	defer func() {
 		if err := manager.Close(); err != nil {
@@ -86,8 +90,14 @@ func run() error {
 	// Signals are intentional shutdowns, not evidence that the candidate is bad.
 	select {
 	case <-server.Ready():
+		if err := selfUpdateStartup.Commit(); err != nil {
+			startupErr := fmt.Errorf("commit refreshed updater startup: %w", err)
+			if shutdownErr := shutdownUpdater(server, manager); shutdownErr != nil {
+				startupErr = errors.Join(startupErr, fmt.Errorf("shutdown after startup commit failure: %w", shutdownErr))
+			}
+			return handleSelfUpdateStartupFailure(selfUpdateStartup, startupErr)
+		}
 		log.Printf("proto-fleet-updater %s listening on %s", version, *socketPath)
-		*selfUpdateHandoff = ""
 	case sig := <-signals:
 		log.Printf("received %s during startup, shutting down", sig)
 		if err := shutdownUpdater(server, manager); err != nil {
@@ -98,7 +108,7 @@ func run() error {
 		if errors.Is(err, http.ErrServerClosed) {
 			return nil
 		}
-		return handleSelfUpdateStartupFailure(*selfUpdateHandoff, fmt.Errorf("serve updater API: %w", err))
+		return handleSelfUpdateStartupFailure(selfUpdateStartup, fmt.Errorf("serve updater API: %w", err))
 	}
 
 	select {
@@ -136,11 +146,11 @@ func run() error {
 	}
 }
 
-func handleSelfUpdateStartupFailure(handoffPath string, startupErr error) error {
-	if handoffPath == "" {
+func handleSelfUpdateStartupFailure(selfUpdateStartup *updater.SelfUpdateStartup, startupErr error) error {
+	if selfUpdateStartup == nil {
 		return startupErr
 	}
-	if rollbackErr := updater.RollbackSelfUpdateExecutable(handoffPath); rollbackErr != nil {
+	if rollbackErr := selfUpdateStartup.Rollback(); rollbackErr != nil {
 		return errors.Join(
 			startupErr,
 			fmt.Errorf("restore previous updater after replacement startup failure: %w", rollbackErr),
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -29,16 +29,11 @@ func main() {
 	if defaultSocketPath == "" {
 		defaultSocketPath = "/run/proto-fleet-updater/updater.sock"
 	}
-	defaultDownloadBase := os.Getenv("PROTO_FLEET_DOWNLOAD_BASE_URL")
-	if defaultDownloadBase == "" {
-		defaultDownloadBase = "https://github.com/block/proto-fleet/releases/download"
-	}
 	defaultSelfUpdatePath := os.Getenv("PROTO_FLEET_UPDATER_BINARY_PATH")
 
 	installRoot := flag.String("install-root", defaultInstallRoot, "Proto Fleet installation root")
 	stateDir := flag.String("state-dir", defaultStateDir, "Durable updater state directory")
 	socketPath := flag.String("socket-path", defaultSocketPath, "Unix socket path")
-	downloadBase := flag.String("download-base-url", defaultDownloadBase, "Official release download base URL")
 	selfUpdatePath := flag.String("self-update-path", defaultSelfUpdatePath, "Installed updater binary path to atomically refresh")
 	showVersion := flag.Bool("version", false, "Print version and exit")
 	flag.Parse()
@@ -52,10 +47,9 @@ func main() {
 		log.Fatal(err)
 	}
 	manager, err := updater.NewManager(updater.Config{
-		InstallRoot:     absoluteInstallRoot,
-		StateDir:        *stateDir,
-		DownloadBaseURL: *downloadBase,
-		SelfUpdatePath:  *selfUpdatePath,
+		InstallRoot:    absoluteInstallRoot,
+		StateDir:       *stateDir,
+		SelfUpdatePath: *selfUpdatePath,
 	})
 	if err != nil {
 		log.Fatalf("initialize updater: %v", err)
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -54,6 +54,11 @@ func main() {
 	if err != nil {
 		log.Fatalf("initialize updater: %v", err)
 	}
+	defer func() {
+		if err := manager.Close(); err != nil {
+			log.Printf("close updater manager: %v", err)
+		}
+	}()
 	server := updater.NewServer(manager)
 	errs := make(chan error, 1)
 	go func() {
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -22,9 +22,17 @@ var version = "dev"
 const selfUpdateHandoffFlag = "self-update-handoff"
 
 func main() {
+	setSecureProcessUmask()
 	if err := run(); err != nil {
 		log.Fatal(err)
 	}
+}
+
+func setSecureProcessUmask() {
+	// Keep the Unix socket and updater-owned artifacts private from creation,
+	// including when the binary is started outside the hardened systemd unit.
+	// The umask is process-global, so establish it before run starts goroutines.
+	syscall.Umask(0o077)
 }
 
 func run() error {
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -17,6 +18,8 @@ import (
 )
 
 var version = "dev"
+
+const selfUpdateHandoffFlag = "self-update-handoff"
 
 func main() {
 	if err := run(); err != nil {
@@ -43,6 +46,7 @@ func run() error {
 	stateDir := flag.String("state-dir", defaultStateDir, "Durable updater state directory")
 	socketPath := flag.String("socket-path", defaultSocketPath, "Unix socket path")
 	selfUpdatePath := flag.String("self-update-path", defaultSelfUpdatePath, "Installed updater binary path to atomically refresh")
+	selfUpdateHandoff := flag.String(selfUpdateHandoffFlag, "", "Internal one-shot rollback path for a refreshed updater")
 	showVersion := flag.Bool("version", false, "Print version and exit")
 	flag.Parse()
 
@@ -60,7 +64,7 @@ func run() error {
 		SelfUpdatePath: *selfUpdatePath,
 	})
 	if err != nil {
-		return fmt.Errorf("initialize updater: %w", err)
+		return handleSelfUpdateStartupFailure(*selfUpdateHandoff, fmt.Errorf("initialize updater: %w", err))
 	}
 	defer func() {
 		if err := manager.Close(); err != nil {
@@ -70,13 +74,33 @@ func run() error {
 	server := updater.NewServer(manager)
 	errs := make(chan error, 1)
 	go func() {
-		log.Printf("proto-fleet-updater %s listening on %s", version, *socketPath)
 		errs <- server.Serve(*socketPath)
 	}()
 
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 	defer signal.Stop(signals)
+
+	// A refreshed process remains rollback-eligible only until it proves real
+	// startup with the production config/state and a bound, secured socket.
+	// Signals are intentional shutdowns, not evidence that the candidate is bad.
+	select {
+	case <-server.Ready():
+		log.Printf("proto-fleet-updater %s listening on %s", version, *socketPath)
+		*selfUpdateHandoff = ""
+	case sig := <-signals:
+		log.Printf("received %s during startup, shutting down", sig)
+		if err := shutdownUpdater(server, manager); err != nil {
+			log.Printf("shutdown: %v", err)
+		}
+		return nil
+	case err := <-errs:
+		if errors.Is(err, http.ErrServerClosed) {
+			return nil
+		}
+		return handleSelfUpdateStartupFailure(*selfUpdateHandoff, fmt.Errorf("serve updater API: %w", err))
+	}
+
 	select {
 	case sig := <-signals:
 		log.Printf("received %s, shutting down", sig)
@@ -92,7 +116,8 @@ func run() error {
 		if canonicalSelfUpdatePath == "" {
 			return fmt.Errorf("self-update completed without a configured executable path")
 		}
-		if err := syscall.Exec(canonicalSelfUpdatePath, os.Args, os.Environ()); err != nil {
+		execArgs := selfUpdateExecArgs(os.Args, canonicalSelfUpdatePath)
+		if err := syscall.Exec(canonicalSelfUpdatePath, execArgs, os.Environ()); err != nil {
 			rollbackErr := manager.RollbackSelfUpdate()
 			if rollbackErr != nil {
 				return errors.Join(
@@ -109,6 +134,39 @@ func run() error {
 		}
 		return nil
 	}
+}
+
+func handleSelfUpdateStartupFailure(handoffPath string, startupErr error) error {
+	if handoffPath == "" {
+		return startupErr
+	}
+	if rollbackErr := updater.RollbackSelfUpdateExecutable(handoffPath); rollbackErr != nil {
+		return errors.Join(
+			startupErr,
+			fmt.Errorf("restore previous updater after replacement startup failure: %w", rollbackErr),
+		)
+	}
+	return fmt.Errorf("start refreshed updater; previous executable restored: %w", startupErr)
+}
+
+func selfUpdateExecArgs(args []string, handoffPath string) []string {
+	if len(args) == 0 {
+		return []string{"proto-fleet-updater", "--" + selfUpdateHandoffFlag + "=" + handoffPath}
+	}
+	cleaned := make([]string, 0, len(args)+1)
+	cleaned = append(cleaned, args[0], "--"+selfUpdateHandoffFlag+"="+handoffPath)
+	for index := 1; index < len(args); index++ {
+		arg := args[index]
+		if arg == "--"+selfUpdateHandoffFlag || arg == "-"+selfUpdateHandoffFlag {
+			index++
+			continue
+		}
+		if strings.HasPrefix(arg, "--"+selfUpdateHandoffFlag+"=") || strings.HasPrefix(arg, "-"+selfUpdateHandoffFlag+"=") {
+			continue
+		}
+		cleaned = append(cleaned, arg)
+	}
+	return cleaned
 }
 
 func shutdownUpdater(server *updater.Server, manager *updater.Manager) error {
--- a/server/cmd/fleet-updater/main.go
+++ b/server/cmd/fleet-updater/main.go
@@ -125,6 +125,18 @@ func run() error {
 		if err := shutdownUpdater(server, manager); err != nil {
 			log.Printf("shutdown: %v", err)
 		}
+		rolledBackPath, err := rollbackSelfUpdateQueuedDuringShutdown(
+			manager.SelfUpdateReady(),
+			manager.RollbackSelfUpdate,
+		)
+		if err != nil {
+			return fmt.Errorf("restore previous updater after shutdown interrupted self-restart: %w", err)
+		}
+		if rolledBackPath != "" {
+			log.Printf(
+				"host updater refresh completed during shutdown; restored the previous updater and deferred the refresh until a future upgrade",
+			)
+		}
 		return nil
 	case canonicalSelfUpdatePath := <-manager.SelfUpdateReady():
 		log.Printf("activating refreshed host updater")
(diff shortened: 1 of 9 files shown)


mcharles-square:
🤖 Follow-up on the automated security review after the latest push:

- Interrupted activation is addressed by durable staged-tree/parent fsyncs, checked restoration, and startup reconciliation that either restores the untouched previous deployment or preserves the forward deployment with a derived recovery command (`35c0c319a`).
- Duplicate daemons are addressed by a lifetime host-state `flock` acquired before state load plus bind-first, fail-closed stale-socket handling (`e133661b4`).
- Command lifecycle is addressed with phase deadlines, whole-process-group cancellation, bounded logs, graceful operation draining, and lock retention through activation (`8b3199b08`). Tests exercise shutdown at both sides of the activation boundary, activation timeout, cleanup timeout, and real descendant termination.
- Independent artifact signing is a valid broader supply-chain capability, but it is not safely solved by adding a key check only inside this engine PR. Independence requires an end-to-end release design for offline/protected private-key custody, public-key bootstrap, rotation, revocation, and release manifest publication; a key controlled by the same GitHub release workflow would not address the stated compromise model. This PR instead pins production to the official release origin and explicitly treats the same-origin SHA-256 sidecar only as corruption detection. Independent signing should be designed across the release pipeline and packaging boundary, rather than treated as a blocker for this currently uninstalled engine.

mcharles-square:
🤖 Follow-up on the latest automated findings (`8d57c50c8`):

- Failed/interrupted swaps now use an atomically installed, fsynced activation marker. Pre-command failures reconcile before terminalization; startup restores `deployment.previous` only when that marker proves the two-rename gap, and persists a guarded recovery command if automatic repair cannot finish.
- Crash-abandoned archives, checksums, updater copies, state temps, and staging trees now have exact operation-derived names and are cleaned under the daemon lock after layout reconciliation, without following symlinks or touching unrelated paths.
- A verified updater replacement now signals only after durable success. The command drains HTTP, waits for operation cleanup and lock release, then execs the replacement; new triggers are barred atomically with the success transition.

The full server suite, lint, vet, Linux build, 20 repeated updater test runs, and five race-detector runs pass. Two independent final audits found no remaining blocker.

mcharles-square:
🤖 Follow-up on the latest security review:

- The packaging observation is accurate for this isolated diff but is intentionally non-blocking here: #836 is dormant and cannot configure a deployment. #837 reports one-click unavailable without the socket; #838 builds the architecture-specific binary into the exact `deployment/updater/proto-fleet-updater` bundle path and adds the unit/overlay; #839 installs and configures the service; #840 exposes the action only when the runtime capability is live. A production-bundle contract test is worthwhile follow-up work in #838, where that artifact is created.
- Invalid self-update candidates are now rejected before replacement. The protected bytes are copied into a deterministic sibling slot, the host kernel must execute `--version` within a deadline, and its bounded output must exactly match the target tag. Replacement retains the working executable as a hard-linked rollback slot; an unexpected exec failure atomically restores it.
- Operation logs now use exact SHA-derived names under a stable confined directory handle. Startup, pre-trigger, and best-effort terminal pruning preserve the current log while enforcing eight files and 256 MiB total, with one log/64 MiB reserved before accepting a trigger. Symlinked log roots fail closed, byte accounting saturates safely, and a queue-persist failure preserves the prior terminal status.

Focused and full server tests, 20 repeated updater runs, five race-detector runs, lint, vet, and the Linux build pass; independent audits found no remaining blocker.

mcharles-square:
🤖 Addressed the two PR-local filesystem findings in `7a37f41ce`: updater state is now accepted only through validated lexical and canonical directory chains (root/daemon ownership, safe write modes, no final symlink) before the lock or pathname-reopened artifacts are used; the Unix-socket boundary applies the same chain checks and explicitly sets the bound socket's daemon UID/GID and `0660` mode. The intended production caller remains root Fleet API, so this closes the local filesystem-authentication gap without adding a second application-auth protocol. Added adversarial mode, ancestor, symlink, canonicalization, setgid, ownership, and socket-mode tests. Validation is green across repeated updater tests, race tests, full server tests/lint, vet, and a Linux `amd64` build. The same-origin checksum observation remains a valid broader release-publisher threat model, but independent signing requires release-key custody, bootstrap, rotation, and revocation across the publishing pipeline; it should be designed separately rather than adding a PR-local trust anchor that does not establish independent provenance.

mcharles-square:
🤖 Followed up on the latest security pass:

- Ancestor symlinks now receive the same root/daemon ownership validation as directory components, and the accepted socket directory is canonicalized once; bind, ownership, and mode operations use that canonical path, so retargeting the configured symlink cannot redirect them.
- A failed terminal-success write now restores the last durable activating state and recovery command in memory, blocks retriggering, suppresses the self-restart signal, and reconciles fail-closed on restart.
- The checksum-sidecar and bundled-updater findings are valid against #836 in isolation but are stack-scope false positives: this PR deliberately leaves the engine dormant. #838 already builds the versioned binary at the exact `deployment/updater/proto-fleet-updater` path and publishes `<archive>.sha256` for stable, RC, and nightly assets; #839 installs/enables it. I’ve made that contract explicit in this PR description. The worthwhile production-bundle contract test should be added with #838’s packaging workflow, where it can verify the real tar contents, checksum, and `--version` output.

Combined validation remains green across repeated updater tests, race tests, full server tests/lint, vet, and the Linux `amd64` build.

mcharles-square:
🤖 Addressed the final privileged-path finding in `5ed93e42f`. `InstallRoot` now validates both configured and canonical chains, rejects unsafe modes/final symlinks, and permits root plus at most one consistent installation-admin UID; existing `deployment` and `deployment.previous` owners join that same policy before their ownership can be propagated to staged code. `SelfUpdatePath` is stricter: the executable and parent chains must be root/daemon-owned and non-writable by group/world, the final file must be a real nonempty executable, and the durable-ready signal carries the canonical path that `main` now passes directly to `syscall.Exec`. Tests retarget configured ancestor symlinks and prove staging, replacement, rollback, and exec signaling remain on the validated canonical paths. I’m pushing back on a full `openat2` rewrite: with every parent protected, an unrelated UID cannot rename a component; the one accepted deployment owner is already host-admin-equivalent under #839’s same-rootful-Docker eligibility contract. #839 should still normalize extraction ownership and explicitly enforce root ownership for the updater install directory/files when that packaging PR is reviewed. Repeated updater/race runs, full server tests/lint, vet, and the Linux `amd64` build are green.

mcharles-square:
🤖 Follow-up on the fresh security review, including the `5083109b2` fix:

- The HA finding was valid and is fixed in `5083109b2`. Activation now carries the operator-owned `ha/node.env` into the staged deployment, preserves its restrictive mode, and fails closed for regular or dangling symlinks. The end-to-end upgrade test verifies content and mode `0600`; repeated focused runs, updater race tests, the full server suite, and lint pass.
- The checksum-sidecar and bundled-updater findings are stack-scope observations, not defects in this dormant engine PR. #838 creates and uploads `<bundle>.sha256` for release channels and packages the architecture-specific updater at `deployment/updater/proto-fleet-updater`; #839 installs/enables it. The production-bundle contract test belongs in #838, where the real asset is assembled.
- Independent publisher authentication remains a valid broader supply-chain design, but a PR-local signature check would not establish independence without key custody, bootstrap, rotation, revocation, and release-manifest publication across the release pipeline. This PR pins the official origin and deliberately treats its same-origin SHA-256 sidecar as corruption detection; independent provenance should be designed separately.

There are no unresolved review threads on #836.

mcharles-square:
🤖 Addressed the new failed-state persistence finding in `a08d9ccf3`. Terminal failure now follows the same durability invariant as terminal success: the manager snapshots the prior nonterminal operation, attempts the failed-state write, and restores the prior recovery-bearing state if persistence fails. That keeps retriggers blocked and preserves durable recovery context until restart reconciliation can record the failure. The regression injects the write failure after the activation command fails and verifies memory/disk stay activating, recovery remains intact, `CompletedAt` stays unset, no false terminal log is emitted, a second trigger is rejected, and restart persists the reconciled failure. Repeated focused tests, the updater race suite, full server tests, lint, and pre-push checks pass. The independent-signing finding remains a separate release-pipeline architecture decision for the reasons documented above.

mcharles-square:
🤖 Addressed the replacement-startup rollback finding in `cf1b2f36f` with a deliberately bounded one-shot handoff. The old process now execs the canonical updater path with an internal rollback marker; the replacement remains rollback-eligible only until the real manager initializes and the Unix socket is bound with its final ownership/mode. A returned manager/socket startup error validates that marker, restores the hard-linked `.previous` executable, and exits nonzero so #839's `Restart=on-failure` restarts the restored binary. Failed `exec` is still restored by the old process, while ordinary starts, intentional signals, and post-readiness failures cannot trigger a downgrade.

I intentionally did not add a second supervisor or a duplicate non-mutating startup probe: either would add lifecycle complexity without exercising the production lock/socket path. SIGKILL/power loss inside the narrow pre-readiness window would require a durable cross-process marker and remains a separate design choice rather than being hidden here. Added argument-handoff, rollback-scope, and secured-socket readiness tests; focused tests, race tests, full server tests, vet, lint, and the Linux static build are green.

mcharles-square:
🤖 Follow-up on the final security review (`c5c8287b4`):

- Fixed the interrupted self-update finding with a protected, fsynced sibling handoff marker created after retaining `.previous` and before replacing the executable. The matching argv now identifies only the first candidate attempt; returned startup errors restore immediately, while signals/kills/power loss leave durable eligibility for the next supervisor start to reconcile before manager initialization. Secured-socket readiness durably clears the marker. Restoration retains the hard link and is idempotent across pre-rename, post-restore, marker-clear, and missing-destination crash states.
- Fixed the staged-tree sync finding by keeping tree and staging-parent durability in `Preflight`, under an activation-length context, before creating the activation marker or entering the non-cancelable phase. Cancellation/deadline is checked throughout and again after the parent `fsync`; the command retains its separate activation deadline. A single in-kernel `fsync` is inherently non-preemptible, but no unbounded traversal can now sit inside the activation boundary.
- The release-asset finding is valid for the complete stack but belongs to #838, which builds/version-injects the updater, packages it at `deployment/updater/proto-fleet-updater`, generates `.sha256` sidecars, and updates release uploads; its production bundle-contract test should land there. #836 remains a dormant engine, and moving publishing into it would defeat the reviewable stack boundary and publish assets before installation/exposure gating.
- The same-origin checksum observation remains a valid broader supply-chain risk, not a safe `manager.go` patch. Independent authenticity needs a release manifest plus independently controlled key custody, CI signing, bootstrap, rotation/revocation, nightly policy, and updater trust-root migration. Adding a verifier without that publisher/lifecycle design would either brick updates or create a nominal trust anchor with no independent provenance.

Validation is green across full server tests, focused and repeated updater tests, race tests, vet, golangci-lint, pre-push lint, Linux `amd64` static build, and three independent lifecycle/crash-safety audits.

server/internal/updater/server.go
```diff
@@ -0,0 +1,314 @@
...
+		return
+	}
+	var operation updaterapi.Operation
+	if request.OperationID == "" {
+		// Backward compatibility for an older fleetd paired with a newer host
+		// updater. New callers always supply an operation ID for exact
+		// lost-response reconciliation.
+		operation, err = s.manager.Trigger(request.TargetVersion)
+	} else {
+		operation, err = s.manager.TriggerWithID(request.TargetVersion, request.OperationID)
```
rl-block: Should we require `operation_id` and retain accepted IDs beyond the current operation? Otherwise a lost-response retry can run again after `m.operation` is replaced. Perhaps we could remove the fallback and persist a bounded ID-to-operation registry.

server/cmd/fleet-updater/main.go
```diff
@@ -0,0 +1,196 @@
...
+		return handleSelfUpdateStartupFailure(selfUpdateStartup, fmt.Errorf("serve updater API: %w", err))
+	}
+
+	select {
+	case sig := <-signals:
+		log.Printf("received %s, shutting down", sig)
+		if err := shutdownUpdater(server, manager); err != nil {
+			log.Printf("shutdown: %v", err)
+		}
+		return nil
```
rl-block, line 140: Perhaps worth draining `SelfUpdateReady()` after shutdown before returning, because activation may finish while shutdown waits, leaving a successful replacement queued and then rolled back on restart. We could complete or explicitly roll back the pending handoff.

server/internal/updater/manager.go
```diff
@@ -0,0 +1,2750 @@
...
+		m.failActivation(operationID, targetVersion, fmt.Errorf("complete activation swap: %w", err), logFile)
+		return
+	}
+
+	if err := m.runCommand(ctx, m.cfg.ActivationTimeout, currentDeployment, commandOutput, "/bin/bash", "./run-fleet.sh", "--non-interactive", "--skip-build"); err != nil {
+		// run-fleet may have applied forward-only migrations before returning an
+		// error. Keep the new deployment active for forward recovery; restoring
+		// the previous binaries could make the schema and application incompatible.
+		m.fail(operationID, fmt.Errorf("new stack failed to start: %w", err), recovery)
+		return
```
rl-block, line 1163: Should recovery be derived after `run-fleet.sh` fails? It may remove an invalid preflight proof, but this path still records a `--skip-build` command that will immediately fail. Perhaps recovery could rerun preflight when the proof is absent.

server/internal/updater/manager.go
```diff
@@ -0,0 +1,2750 @@
...
+		selfUpdateReady: make(chan string, 1),
+		processLock:     processLock,
+		logRoot:         logRoot,
+	}
+	if err := m.loadState(); err != nil {
+		_ = processLock.Close()
+		_ = logRoot.Close()
+		return nil, err
+	}
+	if err := m.cleanupStaleArtifacts(); err != nil {
```
rl-block, line 652: Perhaps have stale-artifact cleanup run before the reconciled state write? Crash artifacts can cause ENOSPC, making startup exit before reaching the cleanup that would free space.

rl-block (approved):

server/cmd/fleet-updater/main.go
mcharles-square: 🤖 Updated by Codex: Addressed in 6a8b988c4. Signal-driven shutdown now waits for activation, non-blockingly drains a queued SelfUpdateReady, and explicitly rolls the handoff back. The self-update arm also stops signal forwarding and checks for a queued SIGINT or SIGTERM after drain and immediately before exec, so a stop that races readiness rolls back instead of launching a replacement. I chose rollback over exec because an intentional supervisor or operator stop should remain a stop; the normal readiness path still execs when no stop is pending.

server/internal/updater/server.go
mcharles-square: 🤖 Updated by Codex: Addressed the practical part in 6a8b988c4: operation_id is now required, the target-only fallback is gone, and missing, empty, or invalid IDs return 400. I am intentionally not adding a persisted historical ID registry here. Current-operation retention covers the normal lost-response retry window; the residual case requires a stale retry to outlive acceptance of another operation that did not advance the installed version. With a local trusted caller that serializes upgrades, that edge does not justify adding registry persistence, migration, and eviction semantics to this engine PR.

server/internal/updater/manager.go
mcharles-square: 🤖 Updated by Codex: Addressed in 6a8b988c4. After activation fails, recovery is rederived from the post-command proof state: a regular proof keeps --skip-build, a missing proof reruns --preflight-only before --skip-build, and a non-regular or unreadable proof fails closed with no misleading recovery command. The persisted status and regression tests cover all three cases.

server/internal/updater/manager.go
mcharles-square: 🤖 Updated by Codex: Addressed narrowly in 6a8b988c4. If the reconciled-state write fails specifically with ENOSPC, startup removes only exact updater-owned stale artifacts and retries that small write once; other persistence errors preserve the artifacts. I intentionally did not move cleanup ahead of deployment-layout reconciliation because staged or swap artifacts can be required recovery evidence, and broad early cleanup could destroy it before the layout is understood.

server/internal/updater/manager.go
```diff
@@ -0,0 +1,2808 @@
...
+	if marker != nil {
+		return updaterapi.Operation{}, newTriggerError(
+			errTriggerBusy,
+			fmt.Sprintf(
+				"activation recovery for operation %s is pending; restart the updater after completing recovery",
+				marker.OperationID,
+			),
+		)
+	}
+	currentVersion, err := readInstalledVersion(filepath.Join(m.cfg.InstallRoot, "deployment", "version.txt"))
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Codex: Addressed in 60a77091c. The installed version is now read and compared inside the final admission lock, after the closing, nonterminal, and operationRunning checks, so a trigger cannot carry a stale version across completion of another upgrade. Same-ID retries still return before host-state validation. The deterministic regression parks a v1.1 request, completes a v1.2 upgrade, and verifies the stale request is rejected without launching another worker. Repeated, race, full-server, and lint validation are green.

PR #837, opened 2026-07-30
feat(updates): bridge Fleet API to host updater

mcharles-square:
Reviewable diff: +537/-27 across 8 files (excludes generated, test, and story files).

## Summary

Adds the permission-gated Fleet RPC contract and Unix-socket client that mediate access to the host updater. Trigger requests now use the idempotent operation-ID contract introduced by #836, so Fleet can safely recover from a lost acknowledgement without starting a duplicate upgrade. When the daemon socket is absent, Fleet reports `one_click_available=false`, preserving the manual update experience.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → **#837** → [#838](https://github.com/block/proto-fleet/pull/838) → [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840). This is **3/6** of the one-click phase; its diff is relative to #836. The parent supplies the privileged updater and accepts both legacy target-only and new ID-bearing requests. Client exposure remains in #838–#840; unsupported hosts continue using the manual flow.

## How it works

`GetUpdateStatus` performs a bounded status probe and advertises one-click only when the independent executor is reachable. `TriggerUpgrade` re-derives the organization’s currently eligible release, rejects stale or arbitrary targets, creates an operation ID, and sends only that ID plus the validated tag over the Unix socket. A lost or malformed acknowledgement is retried once with the same ID, then reconciled against exact durable status; an unconfirmed outcome remains `Unavailable` rather than encouraging a second upgrade path. Confirmed triggers are activity-audited once using a cancellation-detached bounded context.

```mermaid
sequenceDiagram
  participant C as "Future capability-gated client"
  participant F as "Fleet InstanceUpdateService"
  participant U as "Host updater from #836"
  C->>F: "TriggerUpgrade(target)"
  F->>F: "Authorize and revalidate eligible target"
  F->>U: "POST operation ID + target"
  alt "Acknowledgement received"
    U-->>F: "Durable operation"
  else "Outcome is ambiguous"
    F->>U: "Retry same operation ID"
    F->>U: "GET durable status if still unconfirmed"
    U-->>F: "Exact operation ID or no confirmation"
  end
  F-->>C: "Operation or conservative Unavailable"
```

## Areas of the code involved

|---|---|---|
| `proto/instance/v1` | Added capability, trigger/status RPCs, phases, and durable operation model | Public compatibility and lifecycle contract |
| `server/generated`, client generated API | Regenerated Connect and protobuf bindings | Generated — skip |
| `server/internal/domain/updates` | Bounded Unix-socket client, exact-ID retry/reconciliation, target revalidation, audit, and status mapping | Trust and reliability boundary between org policy and privileged execution |
| `server/internal/handlers/updates` | RPC authorization, cancellation-code preservation, protobuf mapping, and explicit denial tests | Permission and error-contract enforcement |
| `server/internal/handlers/interceptors` | Redacts update and operation responses and keeps mutation RPCs session-only | Prevents host details from entering debug logs or API-key paths |
| `server/internal/handlers/middleware` | Classifies both RPCs under org-wide `instance:update` | Prevents unclassified or site-scoped access |
| `server/cmd/fleetd` | Supplies activity logging to the update service | Makes confirmed privileged mutations auditable |

- Revalidate the eligible target at mutation time rather than trusting the browser’s version value.
- Retry ambiguous triggers with the same operation ID and reconcile only an exact ID/target match rather than guessing by release tag.
- Preserve an unknown outcome as `Unavailable` rather than suggesting the install command, which could launch a second upgrade path.
- Derive capability from a bounded live socket probe rather than a static flag that can drift from host state.
- Expose only a narrow Unix-socket protocol to Fleet rather than mounting the Docker socket into the application.

## Testing & validation

- Focused update-domain, updater, handler, interceptor, middleware-contract, fleetd, and updater API tests.
- Race detector for the update domain and host updater, including concurrent same-ID admission.
- Real Unix-socket tests for request shape, status mapping, response bounds, cancellation, lost acknowledgements, and delayed admission.
- Targeted Updates component test: 28 passing; client TypeScript check passing.
- Proto lint, generated-output verification, Go imports, canonical server lint, and pre-push hooks passing.


proto/instance/v1/updates.proto
```diff
@@ -40,6 +62,9 @@ message GetUpdateStatusResponse {
     // Full copy-paste installer invocation for latest_eligible; empty when
     // update_available is false.
     string install_command = 5;
+    // True only when the independently installed host executor is reachable.
+    // False keeps the copy-command fallback visible.
+    bool one_click_available = 6;
     // False until release discovery succeeds and after its latest list fetch
     // fails. Callers must not interpret update_available=false as up to date
     // while status is unavailable.
```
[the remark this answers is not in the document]
mcharles-square: This is not necessary since the code is still unreleased

server/internal/domain/updates/executor.go
```diff
@@ -0,0 +1,182 @@
...
+	response, err := c.http.Do(request)
+	if err != nil {
+		if errors.Is(err, errExecutorUnavailable) {
+			return errExecutorUnavailable
+		}
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return &executorTransportError{cause: ctxErr}
+		}
+		return &executorTransportError{cause: err}
+	}
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Codex: Addressed in 4e41adfbf. When http.Client.Do fails, the request context is now checked before the generic updater-unavailable marker, so cancellation and deadline errors remain executor transport errors instead of being misclassified as a missing socket. The regression models the exact mixed transport error for both cancellation and deadline cases. Repeated, race, updates-domain, handler, and lint validation are green.

server/internal/domain/updates/service.go
```diff
@@ -119,6 +151,151 @@ func (s *Service) GetUpdateStatus(ctx context.Context, organizationID int64) (Up
...
+	}
+	if err := ctx.Err(); err != nil {
+		return updaterapi.Operation{}, fmt.Errorf("trigger canceled before host mutation: %w", err)
+	}
+
+	// Authorization and target validation are complete. Detach the privileged
+	// local mutation from browser cancellation, but keep a hard bound so a
+	// wedged socket cannot hold the RPC forever.
+	operationID := uuid.NewString()
+	operation, err := s.triggerUpgrade(context.WithoutCancel(ctx), operationID, targetVersion)
```
rl-block: `context.WithoutCancel` also removes the HA active-lifetime cancellation carried by this request. Do we want to preserve activation cancellation while detaching only the browser disconnect, perhaps through a separate active-lifetime context? A demotion regression test would help verify the request still drains within the shutdown budget.

server/internal/domain/updates/service.go
```diff
@@ -119,6 +151,151 @@ func (s *Service) GetUpdateStatus(ctx context.Context, organizationID int64) (Up
...
+		return fmt.Errorf("host updater rejected trigger with HTTP status %d", httpErr.StatusCode)
+	}
+	var transportErr *executorTransportError
+	if errors.As(err, &transportErr) || errors.Is(err, context.DeadlineExceeded) {
+		return fleeterror.NewUnavailableErrorf("host updater did not confirm the upgrade; check its status before retrying")
+	}
+	if errors.Is(err, context.Canceled) {
+		return fleeterror.NewUnavailableErrorf("host updater canceled the upgrade request before confirming it")
+	}
+	return fmt.Errorf("trigger host upgrade: %w", err)
```
rl-block: Protocol failures are treated as ambiguous above, but this fallback turns them into `Internal`. Perhaps we map `executorProtocolError` to the same `Unavailable`/check-status result as transport ambiguity. That could probably keep a possibly accepted upgrade from looking like a definitive failure.

rl-block (approved):

server/internal/domain/updates/service.go
mcharles-square: 🤖 Updated by Codex: Addressed in 3ee3abe51. Upgrade work now carries a separate active-runtime lifetime, so browser cancellation can be detached while demotion or shutdown still cancels the initial trigger, retry, reconciliation, and audit. Gate and update-domain regressions verify cancellation and prompt admission draining.

server/internal/domain/updates/service.go
mcharles-square: 🤖 Updated by Codex: Addressed in 7c3aa7050. After same-ID retry and exact status reconciliation are exhausted, malformed, oversized, or identity-mismatched successful responses now remain Unavailable with check-status guidance, matching transport ambiguity and avoiding leaked protocol detail. Focused package and race tests plus server lint pass.

PR #838, opened 2026-07-30
build(updates): package host updater artifacts

mcharles-square:
Reviewable diff: +129/-2 across 5 files (excludes generated, test, and story files).

## Summary

Packages the host updater into amd64 and arm64 deployment artifacts, publishes paired SHA-256 sidecars, and includes the hardened systemd unit plus narrow socket overlay required by the merged Fleet API bridge. The host binary is statically linked and verified before packaging, while release publication fails closed unless both architecture bundles and correctly bound checksums are present. These assets remain dormant until installer activation lands in #839.

*Stack:* Prerequisites [#841](https://github.com/block/proto-fleet/pull/841) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → [#837](https://github.com/block/proto-fleet/pull/837) are merged. This is **4/6**, and its diff is relative to `main`. Remaining descendants are [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840): #839 installs and activates these assets, while #840 exposes one-click execution in the client.

## How it works

The artifact workflow builds a CGO-disabled `fleet-updater` on each native architecture runner, rejects dynamically linked output, and verifies the embedded version. The server artifact carries that binary into the deployment bundle, where it is placed beside its systemd unit and a Compose overlay that exposes only the updater Unix-socket directory to fleet-api.

Before upload, the workflow verifies the bundle checksum and required updater paths. Stable/RC and nightly publishers then require both amd64 and arm64 pairs, regenerate the expected sidecar for each explicit bundle, and compare the complete digest-and-filename line before creating or updating a GitHub release. The service unit preserves the socket directory inode across stops and restarts, remains available through Docker outages, and gives an in-progress activation enough bounded time to drain safely.

```mermaid
flowchart LR
  B["Build static fleet-updater per architecture"] --> V["Verify ELF linkage and embedded version"]
  V --> S["Server artifact"]
  S --> D["Deployment bundle with unit and socket overlay"]
  D --> C["Verify required paths and SHA-256 sidecar"]
  C --> A["Upload amd64 and arm64 artifacts"]
  A --> P["Release/nightly publisher re-verifies each explicit pair"]
```

## Areas of the code involved

|---|---|---|
| `.github/workflows/proto-fleet-artifact-build.yml` | Builds and validates the static updater, packages runtime assets, creates and checks sidecars | Review architecture naming, static-host portability, and the bundle contract |
| `.github/workflows/release.yml`, `nightly-builds.yml` | Require both architecture pairs and bind each sidecar to its explicit bundle before publication | Prevents partial, mismatched, or swapped release assets |
| `deployment-files/docker-compose.updater.yaml` | Mounts only the read-only updater socket directory into fleet-api | Security boundary preventing Docker-socket exposure |
| `deployment-files/updater/proto-fleet-updater.service` | Defines hardening, soft Docker ordering, bounded activation drain, and runtime-directory preservation | Review root-service permissions, recovery, and lifecycle behavior |

- Build the host updater with `CGO_ENABLED=0` and inspect its ELF headers, avoiding host glibc coupling at the cost of requiring pure-Go dependencies.
- Publish same-name `.sha256` sidecars and compare their full content against each explicit amd64/arm64 bundle, keeping updater discovery deterministic and failing closed on partial or swapped artifacts.
- Treat SHA-256 sidecars as corruption and pairing checks, not independent publisher authentication; the GitHub release origin remains the trust anchor, while separately pinned release signing is intentionally out of scope.
- Use `KillMode=mixed` with a 50-minute stop bound so systemd signals the daemon first and preserves the updater's 45-minute activation safety window.
- Preserve the runtime directory across service stops and restarts so Docker's bind mount stays attached to the socket directory inode.
- Soft-order the updater after Docker instead of requiring Docker continuously; trigger-time preflight remains the authority for daemon readiness.
- Keep packaging separate from activation so this PR remains merge-safe; installation and enablement are isolated in #839.

## Testing & validation

- `go test ./cmd/fleet-updater ./internal/updater`
- CGO-disabled build and embedded-version smoke check
- Host profile, HA profile, and complete non-interactive upgrade-safety harnesses
- Base plus updater Compose rendering
- Workflow/Compose YAML parsing, positive checksum-pair simulation, swapped-sidecar rejection, required-path checks, and `git diff --check`
- Pre-push client typecheck plus server and plugin lint hooks; final review-fix push reran server lint

End-to-end deployment artifact assembly is not run by PR CI; the reusable workflow executes only for release and nightly builds. The new in-workflow assertions make that explicit path fail before publication, but a real release/nightly run remains the integration coverage.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -107,6 +107,27 @@ jobs:
       - name: List all release assets
         run: find /tmp/release-assets -type f | sort
 
+      - name: Verify deployment bundle checksums
+        env:
+          VERSION: ${{ github.ref_name }}
+        run: |
+          cd /tmp/release-assets/bundles
+          for arch in amd64 arm64; do
+            bundle="proto-fleet-${VERSION}-${arch}.tar.gz"
+            if [[ ! -f "$bundle" || ! -f "$bundle.sha256" ]]; then
+              echo "::error::Missing deployment bundle or checksum for $arch"
+              exit 1
+            fi
+            # Bind both the digest and filename; sha256sum -c trusts the
+            # filename stored inside the sidecar.
+            expected_sidecar="$RUNNER_TEMP/proto-fleet-${arch}.sha256"
+            sha256sum "$bundle" > "$expected_sidecar"
+            if ! cmp -s "$expected_sidecar" "$bundle.sha256"; then
+              echo "::error::Deployment bundle checksum is not bound to $bundle"
+              exit 1
+            fi
+          done
+
       - name: Create or refresh draft release and upload assets
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -117,6 +138,7 @@ jobs:
           shopt -s failglob
           files=(
             /tmp/release-assets/bundles/proto-fleet-"$TAG_NAME"-*.tar.gz
+            /tmp/release-assets/bundles/proto-fleet-"$TAG_NAME"-*.tar.gz.sha256
             /tmp/release-assets/windows/installer.exe
             /tmp/release-assets/windows/uninstall.exe
             /tmp/release-assets/server/proto-fleet-server-"$TAG_NAME"-*.tar.gz
--- a/deployment-files/updater/proto-fleet-updater.service
+++ b/deployment-files/updater/proto-fleet-updater.service
@@ -1,0 +1,38 @@
+[Unit]
+Description=Proto Fleet host updater
+Documentation=https://github.com/block/proto-fleet
+After=docker.service network-online.target
+# Docker is needed only when an upgrade runs; keep status available while the
+# daemon is down and recover automatically when Docker returns.
+Wants=docker.service network-online.target
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/proto-fleet/updater.env
+ExecStart=/usr/local/libexec/proto-fleet/proto-fleet-updater
+Restart=on-failure
+RestartSec=5s
+# Signal only the daemon first so it can safely drain an activation that has
+# already stopped Fleet. Its internal activation deadline is 45 minutes.
+KillMode=mixed
+TimeoutStopSec=50min
+NoNewPrivileges=true
+PrivateTmp=true
+ProtectSystem=full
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -107,6 +107,20 @@ jobs:
...
+          VERSION: ${{ github.ref_name }}
+        run: |
+          cd /tmp/release-assets/bundles
+          for arch in amd64 arm64; do
+            bundle="proto-fleet-${VERSION}-${arch}.tar.gz"
+            if [[ ! -f "$bundle" || ! -f "$bundle.sha256" ]]; then
+              echo "::error::Missing deployment bundle or checksum for $arch"
+              exit 1
+            fi
+            sha256sum -c "$bundle.sha256"
```
rl-block: Worth binding this check to the loop's `$bundle` instead of trusting the filename stored in the sidecar? Because both architecture tarballs are present, swapped sidecars can still pass here, while the updater later rejects the mismatched pair. Maybe compare the complete sidecar with `sha256sum "$bundle"` and add a swapped-sidecar case; the nightly check has the same pattern.

deployment-files/updater/proto-fleet-updater.service
```diff
@@ -0,0 +1,37 @@
+[Unit]
+Description=Proto Fleet host updater
+Documentation=https://github.com/block/proto-fleet
+After=docker.service network-online.target
+Wants=network-online.target
+Requires=docker.service
```
rl-block: Should this be a soft `Wants=docker.service` dependency while keeping `After=`? With `Requires=`, a Docker boot failure or later stop can leave the updater inactive even after Docker recovers, so one-click updates need manual service intervention. A Docker stop/start lifecycle check might be worth adding too.

rl-block (approved):

PR #839, opened 2026-07-30
feat(updates): bootstrap host updater installs

mcharles-square:
Reviewable diff: +907/-32 across 6 files (excludes generated, test, and story files).

## Summary

Bootstraps the packaged host updater on supported Linux/systemd installations with rootful Docker, then exposes only its secured Unix socket to Fleet. Unsupported hosts and any bootstrap path that cannot reach a verified safe state explicitly retain the existing copy-command experience.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → [#837](https://github.com/block/proto-fleet/pull/837) → [#838](https://github.com/block/proto-fleet/pull/838) → **#839** → [#840](https://github.com/block/proto-fleet/pull/840). This is **5/6** of the one-click phase. All ancestors are merged, so this diff is directly against `main`; those PRs provide release discovery, capability reporting, the privileged updater, and the server trigger path. This PR installs and connects that updater, while #840 adds the operator-facing one-click action. Manual upgrades remain available throughout.

## How it works

The installer validates the requested release name, downloads the bundle and matching SHA-256 sidecar into a private temporary directory, checks the exact filename and digest, and captures any legacy optional-overlay choices before replacing the deployment. On a compatible host it installs the updater binary, root-only environment, and systemd unit; verifies root and the installer use the same Docker daemon; and waits up to 60 seconds for an authenticated request over the Unix socket to succeed. Only then does `run-fleet.sh` persist `ENABLE_ONE_CLICK_UPDATES=true` and layer the socket-only Compose overlay. Every unsupported or safely recoverable failure persists `false`; cleanup uncertainty aborts the install. Uninstall stops and verifies the service before touching Docker, then removes exact updater files, state, runtime sockets, and hash-named staging directories.

```mermaid
flowchart LR
  I["Installer"] --> D["Private bundle download"]
  D --> V["Validate version, sidecar name, and SHA-256"]
  V --> H{"Compatible systemd and rootful Docker host?"}
  H -->|"yes"| S["Install service and verify Unix-socket readiness"]
  S --> O["Persist enabled state and layer updater socket"]
  H -->|"no or safe fallback"| M["Persist disabled state and keep copy command"]
  O --> F["Start Fleet"]
  M --> F
```

```mermaid
sequenceDiagram
  participant I as Installer
  participant G as GitHub Releases
  participant U as Host updater service
  participant R as run-fleet
  participant F as fleet-api
  I->>G: Download bundle and checksum sidecar
  I->>I: Verify exact asset digest and migrate legacy flags
  I->>U: Install, enable, restart
  I->>U: Poll status over secured Unix socket
  alt updater is ready
    I->>R: Enable one-click overlay
    R->>F: Start with read-only socket mount
  else unsupported or safely disabled
    I->>R: Disable one-click overlay
    R->>F: Start with manual fallback only
  end
```

## Areas of the code involved

|---|---|---|
| `deployment-files/install.sh` | Release validation, private checksum download, legacy option capture, daemon compatibility, service bootstrap/readiness, and fail-closed fallback | Highest-risk host mutation and capability boundary |
| `deployment-files/run-fleet.sh` | Persists the one-click setting and conditionally layers the updater socket overlay | Prevents the UI capability from diverging from the active Compose model |
| `deployment-files/uninstall.sh` | Serializes service shutdown ahead of teardown and removes exact updater/staging artifacts | Review destructive scope, privilege use, and failure ordering |
| `deployment-files/README.md`, `README.md` | Supported-host contract, exact-version install flow, trust boundary, logs, and recovery | Defines operator expectations and fallback behavior |
| `.github/workflows/deployment-config-checks.yml` | Adds individual shell syntax validation | Catches installer/uninstaller syntax outside sourced test helpers |
| Focused deployment tests | Covers migration ambiguity, CLI enable/disable persistence, overlay rendering, activation handoff, and cleanup verification | Tests — review alongside the lifecycle paths |

- One-click is enabled only when root and the installer see the same Docker daemon, avoiding an unsafe rootless-Docker privilege bridge.
- Service readiness requires both a successful socket status request and an active unit before Fleet receives the socket mount.
- Failures before readiness must produce a verified inactive/disabled service; once readiness succeeds, the service stays available if later deployment work fails so persisted enabled state never points at a dead socket.
- The SHA-256 sidecar detects corruption and binds the installer to the named release asset; it is not independent publisher authentication. GitHub Releases remains the publisher trust anchor, and detached signing is outside this PR.
- Uninstall removes only fixed paths and the updater's exact lowercase 64-hex staging-name contract; alternate/multi-instance ownership remains outside the supported single-install model.

## Testing & validation

- Individual `bash -n` validation for installer, runner, uninstaller, and deployment test scripts.
- `./deployment-files/tests/test-install-overlay-migration.sh`
- `./deployment-files/tests/test-profiles.sh`
- `./deployment-files/tests/test-run-fleet-upgrade.sh`
- `git diff --check origin/main..HEAD`
- Pre-commit and pre-push hooks passed.
- A real systemd/rootful-Docker clean-install and upgrade matrix was not run locally; that remains explicit release-host validation.

### DIFF
--- a/deployment-files/install.sh
+++ b/deployment-files/install.sh
@@ -1105,10 +1105,11 @@ complete_disabled_updater_fallback_after_run() {
 
 reconcile_updater_after_failed_deployment() {
   local env_file="$1"
-  local expected_state="$2"
-  local reenable_after_failure="$3"
-  local restart_after_failure="$4"
-  shift 4
+  local previous_state="$2"
+  local desired_state="$3"
+  local reenable_after_failure="$4"
+  local restart_after_failure="$5"
+  shift 5
   local privilege=("$@")
   local restored_state
 
@@ -1129,8 +1130,9 @@ reconcile_updater_after_failed_deployment() {
     return 1
   fi
   [ "$restored_state" != "missing" ] || restored_state=false
-  if [ "$restored_state" != "$expected_state" ]; then
-    echo "❌ The failed deployment did not restore the prior one-click update setting." >&2
+  if [ "$restored_state" != "$previous_state" ] \
+    && [ "$restored_state" != "$desired_state" ]; then
+    echo "❌ The failed deployment left an unknown one-click update setting." >&2
     disable_updater_service_with \
       ${privilege[@]+"${privilege[@]}"} || true
     return 1
@@ -1140,7 +1142,16 @@ reconcile_updater_after_failed_deployment() {
       ${privilege[@]+"${privilege[@]}"}; then
     return 1
   fi
-  if [ "$expected_state" = "true" ]; then
+  if [ "$restored_state" = "$desired_state" ] \
+    && [ "$desired_state" != "$previous_state" ]; then
+    if ! ${privilege[@]+"${privilege[@]}"} systemctl enable \
+        proto-fleet-updater.service >/dev/null 2>&1 \
+      || ! restart_updater_service_with "$UPDATER_SOCKET_PATH" \
+        ${privilege[@]+"${privilege[@]}"}; then
+      echo "❌ Could not retain the host updater after the deployment topology changed." >&2
+      return 1
+    fi
+  elif [ "$previous_state" = "true" ]; then
     if [ "$reenable_after_failure" = "1" ] \
       && ! ${privilege[@]+"${privilege[@]}"} systemctl enable \
         proto-fleet-updater.service >/dev/null 2>&1; then
@@ -2084,7 +2095,8 @@ if [ "$NON_INTERACTIVE" = "1" ]; then
   RUN_FLEET_ARGS+=(--non-interactive)
 fi
 RUN_FLEET_STATUS=0
-if ./run-fleet.sh "${RUN_FLEET_ARGS[@]}"; then
+if PROTO_FLEET_INSTALLER_MANAGED_RUN=1 \
+  ./run-fleet.sh "${RUN_FLEET_ARGS[@]}"; then
   RUN_FLEET_STATUS=0
 else
   RUN_FLEET_STATUS=$?
@@ -2111,6 +2123,7 @@ if [ "$RUN_FLEET_STATUS" -ne 0 ]; then
     && ! reconcile_updater_after_failed_deployment \
       "$INSTALL_DIR/$DEPLOYMENT_DIR/.env" \
       "$PREVIOUS_ONE_CLICK_UPDATES" \
+      true \
       "$UPDATER_REENABLE_AFTER_FAILED_RUN" \
       "$UPDATER_RESTART_AFTER_FAILED_RUN" \
       ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"}; then
--- a/deployment-files/install.sh
+++ b/deployment-files/install.sh
@@ -8,6 +8,7 @@ UPDATER_CLEANUP_FAILED=0
 UPDATER_PRIVILEGE=()
 UPDATER_PRIVILEGE_AVAILABLE=1
 UPDATER_DISABLE_ON_EXIT=0
+UPDATER_REENABLE_ON_EXIT=0
 UPDATER_RESTART_ON_EXIT=0
 UPDATER_SOCKET_PATH="/run/proto-fleet-updater/updater.sock"
 UPDATER_VALIDATION_SOCKET_PATH="/run/proto-fleet-updater/installer-validation.sock"
@@ -625,6 +626,52 @@ prepare_fresh_install_path() {
   printf '%s\n' "$canonical"
 }
 
+# Recognize an explicitly selected, stopped deployment without weakening the
+# fresh-install path boundary. Docker cannot identify a deployment after its
+# containers have been removed, so the on-disk marker is accepted only after
+# the root, deployment directory, and marker have all passed the same trusted
+# owner and non-symlink checks used by the privileged updater.
+canonical_existing_install_path() {
+  local selected="$1"
+  local normalized canonical trusted_uid deployment_dir marker metadata owner_uid mode
+  normalized=$(lexical_absolute_path "$selected") || return 1
+  reject_dangerous_install_root "$normalized" || return 1
+  if [ -L "$normalized" ] || [ ! -d "$normalized" ]; then
+    echo "❌ Existing installation root must be a real directory, not a symlink: $normalized" >&2
+    return 1
+  fi
+
+  canonical=$(cd "$normalized" 2>/dev/null && pwd -P) || {
+    echo "❌ Existing installation root cannot be accessed: $normalized" >&2
+    return 1
+  }
+  reject_dangerous_install_root "$canonical" || return 1
+  trusted_uid=$(install_admin_uid) || {
+    echo "❌ Could not determine the invoking installation administrator." >&2
+    return 1
+  }
+  validate_install_directory_chain "$canonical" "$trusted_uid" 1 || return 1
+
+  deployment_dir="$canonical/$DEPLOYMENT_DIR"
+  marker="$deployment_dir/docker-compose.yaml"
+  if [ -L "$deployment_dir" ] || [ ! -d "$deployment_dir" ]; then
+    echo "❌ Existing deployment directory must be a real directory, not a symlink: $deployment_dir" >&2
+    return 1
+  fi
+  validate_install_directory_chain "$deployment_dir" "$trusted_uid" 1 || return 1
+  if [ -L "$marker" ] || [ ! -f "$marker" ]; then
+    echo "❌ Existing deployment marker must be a regular, non-symlink file: $marker" >&2
+    return 1
+  fi
+  metadata=$(install_path_metadata "$marker") || {
+    echo "❌ Could not inspect existing deployment marker metadata: $marker" >&2
+    return 1
+  }
+  read -r owner_uid mode <<< "$metadata"
+  validate_install_path_metadata "$marker" "$owner_uid" "$mode" "$trusted_uid" 1 || return 1
+  printf '%s\n' "$canonical"
+}
+
 resolve_selected_install_path() {
   local selected="$1"
   local discovered="${2:-}"
@@ -775,6 +822,17 @@ resolve_updater_privilege() {
   fi
 }
 
+arm_existing_updater_restoration() {
+  local active_state="$1"
+  local unit_file_state="$2"
+  case "$unit_file_state" in
+    enabled|enabled-runtime) UPDATER_REENABLE_ON_EXIT=1 ;;
+  esac
+  case "$active_state" in
+    active|activating|reloading) UPDATER_RESTART_ON_EXIT=1 ;;
+  esac
+}
+
 # Stop the existing updater before the release tree can be replaced. systemd's
 # inactive state proves the supervised process has exited, which also releases
 # the updater's lifetime flock. A missing unit needs no privilege or prompt.
@@ -785,7 +843,7 @@ prepare_existing_updater_service() {
     return 0
   fi
 
-  local load_state
+  local load_state active_state unit_file_state
   if ! load_state=$(systemctl show --property=LoadState --value \
       proto-fleet-updater.service 2>/dev/null); then
     echo "❌ Could not inspect the existing host updater before installation." >&2
@@ -800,7 +858,23 @@ prepare_existing_updater_service() {
     UPDATER_CLEANUP_FAILED=1
     return 1
   fi
-  disable_updater_service_with ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"}
+  if ! active_state=$(systemctl show --property=ActiveState --value \
+      proto-fleet-updater.service 2>/dev/null) \
+    || ! unit_file_state=$(systemctl show --property=UnitFileState --value \
+      proto-fleet-updater.service 2>/dev/null); then
+    echo "❌ Could not inspect the existing host updater state before installation." >&2
+    UPDATER_CLEANUP_FAILED=1
+    return 1
+  fi
+  # Arm restoration before the stop/disable call. Even if its final-state
+  # verification fails after systemd performed the mutation, EXIT cleanup can
+  # still return the service to the state observed above.
+  arm_existing_updater_restoration "$active_state" "$unit_file_state"
+  if ! disable_updater_service_with \
+      ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"}; then
+    return 1
+  fi
+  return 0
 }
 
 write_updater_environment_file() {
@@ -887,6 +961,14 @@ installer_exit_cleanup() {
     echo "🔒 Disabling the host updater after interrupted validation..." >&2
     if ! disable_updater_service_with \
       ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"}; then
+      status=1
+    fi
+  fi
+  if [ "${UPDATER_REENABLE_ON_EXIT:-0}" = "1" ]; then
+    echo "🔄 Restoring host updater boot enablement after the interrupted installation..." >&2
+    if ! ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"} systemctl enable \
+        proto-fleet-updater.service >/dev/null 2>&1; then
+      echo "❌ Could not restore host updater boot enablement." >&2
       status=1
     fi
   fi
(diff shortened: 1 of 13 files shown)


deployment-files/install.sh
```diff
@@ -345,7 +675,13 @@ get_default_install_dir() {
 }
 
 echo "🔍 Checking for previous ProtoFleet installations via Docker..."
-detect_previous_install || true
+if [ -z "$REQUESTED_INSTALL_DIR" ]; then
+  detect_previous_install || true
+else
+  PREVIOUS_INSTALL_DIR=""
+  PREVIOUS_INSTALL_NEEDS_SUDO=0
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Codex in bc5004a9fd. Explicit --install-dir installs now retain existing-deployment discovery and the Docker-daemon ownership and safety probe; regression coverage exercises the unprivileged-to-sudo root-daemon path.

deployment-files/install.sh
```diff
@@ -447,5 +822,208 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+      current_docker_id=$(docker info --format '{{.ID}}' 2>/dev/null || true)
+      [ -z "$current_docker_id" ] || break
+      sleep 0.25
+    done
+  fi
+  root_docker_id=$(${privilege[@]+"${privilege[@]}"} docker info --format '{{.ID}}' 2>/dev/null || true)
+  if [ -z "$current_docker_id" ] || [ "$current_docker_id" != "$root_docker_id" ]; then
+    echo "ℹ️  Root does not manage this installation's Docker daemon; keeping copy-command upgrades."
+    disable_updater_service_with ${privilege[@]+"${privilege[@]}"} || true
+    return 1
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Codex assessment: no code change. The packaged updater supports one Proto Fleet installation per host. On that supported path, disabling the machine-global updater when the selected deployment uses a different or rootless daemon is the safe fail-closed behavior. Reading another install root from privileged global state would add ownership and path-canonicalization complexity solely for an unsupported multi-instance topology, which the PR documents as out of scope.

deployment-files/uninstall.sh
```diff
@@ -457,6 +460,171 @@ remove_systemd_units() {
...
+
+  local privilege=()
+  if (( ${#HOST_UPDATER_PRIVILEGE[@]} )); then
+    privilege=("${HOST_UPDATER_PRIVILEGE[@]}")
+  fi
+  if ! ${privilege[@]+"${privilege[@]}"} rm -f \
+    /etc/systemd/system/proto-fleet-updater.service \
+    /usr/local/libexec/proto-fleet/proto-fleet-updater \
+    /usr/local/libexec/proto-fleet-updater \
+    /etc/proto-fleet/updater.env; then
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Codex assessment: no code change. Uninstall owns the machine-global updater under the supported one-install-per-host model. Retaining it based on a different deployment path would formalize unsupported multi-instance semantics and add privileged state and path handling without improving the supported flow; the PR documents alternate and multi-instance ownership as out of scope.

deployment-files/install.sh
```diff
@@ -447,5 +822,208 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+  fi
+
+  if ! ${privilege[@]+"${privilege[@]}"} install -d -m 0755 /usr/local/libexec/proto-fleet /etc/proto-fleet; then
+    rm -f "$env_temp"
+    disable_updater_service_with ${privilege[@]+"${privilege[@]}"} || true
+    return 1
+  fi
+  if ! ${privilege[@]+"${privilege[@]}"} install -m 0755 updater/proto-fleet-updater /usr/local/libexec/proto-fleet/proto-fleet-updater \
+    || ! ${privilege[@]+"${privilege[@]}"} install -m 0644 updater/proto-fleet-updater.service /etc/systemd/system/proto-fleet-updater.service \
+    || ! ${privilege[@]+"${privilege[@]}"} install -m 0600 "$env_temp" /etc/proto-fleet/updater.env; then
```
rl-block: Could be worth keeping privileged bootstrap inputs in the private verified directory and normalize deployment ownership before copying them. The extracted tree may preserve an unrelated archive UID/GID or be writable by another UID, causing either trust-check failure or a modified payload crossing into root execution.

deployment-files/install.sh
```diff
@@ -406,21 +742,60 @@ if [ "${PREVIOUS_INSTALL_SUDO_BLOCKED:-0}" = "1" ]; then
...
-# Without this, stdin is the curl pipe (already consumed) and the reads
-# silently hit EOF, leaving the responses empty — defaulting users into the
-# happy path with no chance to redirect.
-read -p "   Use this location? (Y/n): " use_suggested < /dev/tty
-if [[ "$use_suggested" =~ ^[Nn]$ ]]; then
-  read -p "   Enter installation directory [${DEFAULT_INSTALL_DIR}]: " custom_dir < /dev/tty
-  INSTALL_DIR="${custom_dir:-$DEFAULT_INSTALL_DIR}"
-else
+if [ -n "$REQUESTED_INSTALL_DIR" ]; then
+  INSTALL_DIR="$REQUESTED_INSTALL_DIR"
```
rl-block, line 1908: Not sure if we want to reject a requested path that differs from the canonical Docker-discovered installation? Because if there is a typo it can replace the active stack through the shared `deployment` Compose project.

deployment-files/install.sh
```diff
@@ -406,21 +742,60 @@ if [ "${PREVIOUS_INSTALL_SUDO_BLOCKED:-0}" = "1" ]; then
...
+if [ "${PREVIOUS_INSTALL_NEEDS_SUDO:-0}" = "1" ]; then
+  CAPTURE_PRIVILEGE=(sudo -n)
+fi
+if ! capture_previous_run_options "$INSTALL_DIR" "$EXISTING_DEPLOYMENT" \
+  ${CAPTURE_PRIVILEGE[@]+"${CAPTURE_PRIVILEGE[@]}"}; then
+  echo "❌ Existing deployment options could not be migrated safely; no files were replaced." >&2
+  exit 1
+fi
+
+extract_and_cd "$TAR_PATH" "$INSTALL_DIR"
```
rl-block, line 1980: We should probably stop and drain the matching updater before extraction. An activation can rename `deployment` while this installer writes it, producing a mixed deployment or running from `deployment.previous`.

deployment-files/install.sh
```diff
@@ -447,5 +822,208 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+    # fail the daemon-identity comparison below.
+    ${privilege[@]+"${privilege[@]}"} systemctl start docker.service >/dev/null 2>&1 || true
+    for attempt in {1..20}; do
+      current_docker_id=$(docker info --format '{{.ID}}' 2>/dev/null || true)
+      [ -z "$current_docker_id" ] || break
+      sleep 0.25
+    done
+  fi
+  root_docker_id=$(${privilege[@]+"${privilege[@]}"} docker info --format '{{.ID}}' 2>/dev/null || true)
+  if [ -z "$current_docker_id" ] || [ "$current_docker_id" != "$root_docker_id" ]; then
```
rl-block: Might be worth probing the actual Docker environment the systemd service receives? For UID 0, both probes inherit shell Docker selectors that the unit does not, so later upgrades can target a different daemon.

deployment-files/uninstall.sh
rl-block: We might also want to remove and verify the updater's self-update siblings (`.previous`, `.candidate`, `.handoff`, `.handoff.tmp`, and `.restore`)? Otherwise uninstall can report success while privileged artifacts remain.

deployment-files/install.sh
```diff
@@ -447,5 +822,208 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+    echo "⚠️  This release does not contain the host updater payload."
+    return 1
+  fi
+
+  local privilege=()
+  if [ "$(id -u)" -ne 0 ]; then
+    command -v sudo >/dev/null 2>&1 || {
+      echo "⚠️  sudo is unavailable; one-click upgrades cannot be enabled."
+      return 1
+    }
```
rl-block: Would it be worth routing these early fallbacks through the same checked updater cleanup as the later failure paths? On an upgraded host, returning here can leave the existing updater enabled even though Fleet persists one-click updates as disabled.

rl-block (approved):

deployment-files/install.sh
```diff
@@ -447,5 +822,208 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+    # one-click upgrades. A rootless DOCKER_HOST remains distinct and will
+    # fail the daemon-identity comparison below.
+    ${privilege[@]+"${privilege[@]}"} systemctl start docker.service >/dev/null 2>&1 || true
+    for attempt in {1..20}; do
+      current_docker_id=$(docker info --format '{{.ID}}' 2>/dev/null || true)
+      [ -z "$current_docker_id" ] || break
+      sleep 0.25
+    done
+  fi
+  root_docker_id=$(${privilege[@]+"${privilege[@]}"} docker info --format '{{.ID}}' 2>/dev/null || true)
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Fixed in 317bdc485. The installer now compares the caller daemon with a selector-free service probe, and the unit explicitly unsets Docker endpoint and TLS selectors. Added UID 0 selector regression coverage.

deployment-files/install.sh
mcharles-square: 🤖 Fixed in 317bdc485. The binary and unit are now extracted directly from the checksum-verified archive into the private 0700 download directory, and privileged copies use only those sources. Deployment extraction also uses --no-same-owner to avoid release-builder UID/GID drift without recursively changing operator-owned files.

deployment-files/install.sh
mcharles-square: 🤖 Fixed in 317bdc485. The final selected and discovered roots are physically canonicalized, and mismatches—including nonexistent typo paths—fail before updater shutdown or extraction. Relocation and multi-instance behavior remain intentionally unsupported.

deployment-files/install.sh
mcharles-square: 🤖 Fixed in 317bdc485. An existing updater is now disabled and verified inactive before extraction; systemd process exit releases the updater lifetime flock. Unsafe or query failures abort before files are replaced.

deployment-files/install.sh
mcharles-square: 🤖 Fixed in 317bdc485. The service-side Docker probe now clears daemon selectors before resolving the Docker identity, and the unit explicitly unsets those selectors. Added the UID 0 regression case described here.

deployment-files/uninstall.sh
mcharles-square: 🤖 Fixed in 317bdc485. Uninstall discovery, exact-path deletion, and post-cleanup verification now cover .previous, .candidate, .handoff, .handoff.tmp, and .restore for both canonical and legacy binary paths. Focused tests also verify lookalike paths are preserved.

deployment-files/install.sh
mcharles-square: 🤖 Fixed in 317bdc485. The existing service is drained before extraction, and every Linux/systemd fallback now routes through the checked disable and verification path before one-click updates are persisted disabled.

deployment-files/install.sh
```diff
@@ -406,21 +917,73 @@ if [ "${PREVIOUS_INSTALL_SUDO_BLOCKED:-0}" = "1" ]; then
   echo "      curl -fsSL https://fleet.proto.xyz/install.sh | sudo bash -s -- ${QUOTED_VERSION})"
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Fixed in 04362a3bd. After the final install path is selected, an unprivileged run now aborts before migration, updater shutdown, or extraction when sudo blocked root-daemon discovery and that path contains an existing deployment. Fresh paths retain the warning-only behavior. Added regression coverage for complete persisted overlay values and an ordering assertion.

deployment-files/install.sh
```diff
@@ -447,5 +1031,166 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+  echo "✅ Host updater installed; one-click upgrades are enabled."
+  RUN_FLEET_ARGS+=(--enable-one-click-updates)
+else
+  # Keep every fallback fail-closed, including returns that happen before the
+  # service payload is copied. Unsupported non-systemd hosts have no system
+  # updater to reconcile.
+  if [ "$(uname -s)" = "Linux" ] \
+    && command -v systemctl >/dev/null 2>&1 \
+    && [ -d /run/systemd/system ]; then
+    disable_updater_service_with ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"} || true
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Codex: Fixed in 8f23d469d. Fallback cleanup now reads LoadState without elevation and returns safely when the updater unit is not found, so unavailable sudo -n no longer aborts after extraction. Known units still require privileged cleanup and verified final state. Added regressions for both paths.

deployment-files/install.sh
```diff
@@ -379,13 +1508,26 @@ fi
 # this really is a ProtoFleet install (and not some unrelated 'deployment/'
 # tree the user happened to create).
 if [ -z "${PREVIOUS_INSTALL_DIR:-}" ] \
+  && [ -n "$REQUESTED_INSTALL_DIR" ] \
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Addressed in c0635d2ef: interactive selections are promoted through canonical_existing_install_path after prompting and before fresh-path validation. Regression coverage exercises a stopped custom-path deployment selected interactively.

deployment-files/install.sh
```diff
@@ -447,5 +1723,220 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+    && [ -d /run/systemd/system ]; then
+    disable_updater_service_with ${UPDATER_PRIVILEGE[@]+"${UPDATER_PRIVILEGE[@]}"} || true
+  fi
+  if [ "$UPDATER_CLEANUP_FAILED" = "1" ]; then
+    echo "❌ Installation stopped because the host updater could not be left in a safe state." >&2
+    exit 1
+  fi
+  UPDATER_DISABLE_ON_EXIT=0
+  UPDATER_REENABLE_ON_EXIT=0
+  UPDATER_RESTART_ON_EXIT=0
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Addressed in 51a5baedb: fallback restoration remains armed while run-fleet executes and is cleared only after the persisted .env explicitly records ENABLE_ONE_CLICK_UPDATES=false. A failed run before persistence restores the prior updater, with regression coverage for both states.

deployment-files/install.sh
```diff
@@ -447,5 +1824,236 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+fi
+if [ "$PREVIOUS_SYSTEM_MONITORING" = "1" ]; then
+  RUN_FLEET_ARGS+=(--enable-system-monitoring)
+fi
+if [ "$PREVIOUS_TRACING" = "1" ]; then
+  RUN_FLEET_ARGS+=(--enable-tracing)
+fi
+if install_updater_service; then
+  echo "✅ Host updater installed and validated; one-click upgrades will be enabled after deployment."
+  RUN_FLEET_ARGS+=(--enable-one-click-updates)
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Pushing back: stable/RC artifacts predating these flags lack the required checksum and are already routed to their matching legacy installer. The documented exact-version and nightly flows also fetch install.sh from the selected release, preserving the installer/runner pairing. The only matching edge is mixing a future installer with intermediate August nightlies produced while this unshipped stack was landing; carrying runtime feature detection for those transient artifacts would add a compatibility shim outside the supported release contract. No code change planned.

deployment-files/install.sh
```diff
@@ -447,5 +1895,251 @@ for plugin in "${REQUIRED_PLUGINS[@]}"; do
...
+
+UPDATER_RESTART_FAILED=0
+if [ "$UPDATER_START_AFTER_RUN" = "1" ]; then
+  echo "🔄 Starting the host updater after the manual deployment run..."
+  # run-fleet is finished, so the old API can no longer race this deployment.
+  # systemd owns the service lifecycle once the restart is issued.
+  UPDATER_DISABLE_ON_EXIT=0
+  UPDATER_RESTART_ON_EXIT=0
+  UPDATER_REENABLE_ON_EXIT=0
+  UPDATER_START_AFTER_RUN=0
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Codex: fixed in 35a2caf2b. The installer now keeps cleanup armed until production updater readiness succeeds; on failure it redeploys with one-click updates disabled, verifies the persisted fallback, and leaves updater disablement armed if that fallback also fails.

PR #841, opened 2026-07-30
feat(updates): add release discovery engine

mcharles-square:
Reviewable diff: +1537/-2 across 13 files (excludes generated, test, and story files).

## Summary

Adds a dormant GitHub release-discovery engine that identifies the newest canonical stable and release-candidate builds. Production discovery is fixed to the Proto Fleet repository, runs at most hourly with subtract-only jitter, and exposes no operator experience until downstream PRs wire it into fleetd.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → [#837](https://github.com/block/proto-fleet/pull/837) → [#838](https://github.com/block/proto-fleet/pull/838) → [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840). This is 1/5 of phase one, relative to `main`. Contract/storage, backend delivery, client surfaces, and host execution remain downstream.

## How it works

Each cycle makes two baseline requests to the fixed `/releases/latest` endpoint and page 1 of `/releases?per_page=100&page=1`. Responses are capped at 8 MiB and the list is stream-decoded, rejecting the 101st entry. Exact tag grammar and semver validation select the maximum stable and RC candidates; prerelease metadata and release-note links are derived from that classification and the fixed repository rather than upstream body fields.

Baseline responses are cached by ETag and reused on `304`; an optional GitHub token makes unchanged authenticated polls quota-free. A reported rate limit suppresses requests until reset. If a higher cached candidate ages out of the bounded responses, the checker conditionally revalidates it through `/releases/tags/{tag}` before keeping it eligible. A 404 or channel reclassification drops it; a transient, malformed, or mismatched response retains the data for retry but marks only that channel unavailable, with one warning after three consecutive failures.

Stable and RC availability are exposed through invariant-preserving accessors, allowing a fresh RC-only snapshot to remain useful to Stable+RC operators. Startup checks immediately, then repeats every configured hour minus 10–20% jitter (48–54 minutes). Each polling iteration contains panics so a defect in this best-effort feature cannot crash fleetd, and the next scheduled cycle still runs.

```mermaid
flowchart LR
  G["Fixed Proto Fleet GitHub endpoints"] --> B["Two bounded baseline response readers"]
  B --> V["Canonical stable and RC validation"]
  V --> S["Snapshot: releases, fetched-at, per-channel availability"]
  S --> R["Conditional exact-tag revalidation for aged-out cached candidates"]
  R --> S
  S --> D["Dormant until backend wiring"]
```

## Areas of the code involved

|---|---|---|
| `server/internal/domain/updates/github.go` | Fixed baseline client, authenticated ETag caching, rate-limit suppression, body/count bounds, strict JSON handling | Network, quota, withdrawal, and allocation boundary |
| `server/internal/domain/updates/checker.go` | Hourly jitter, semver selection, snapshot accessors, cache revalidation visibility, panic containment | Release eligibility and outage semantics |
| `server/internal/domain/updates/config.go` | Hourly default, optional GitHub token, and allowlisted download origin | Configuration and command trust boundary |
| `server/internal/domain/updates/*_test.go`, `testdata/` | Hostile payload, grammar, ETag/auth, rate-limit, panic, withdrawal, outage, and lifecycle coverage | Tests/fixtures |
| `plugin/asicrs/Dockerfile.build` | Package-aware cleanup of the dummy dependency-warmup artifact | Ensures Cargo rebuilds the real plugin across nightly toolchain changes |
| `docs/plans/...release-update-notifications-plan.md` | Updated end-to-end design | Review context; no runtime effect |

- Use two fixed baseline requests and at most one conditional exact-tag lookup per aged-out channel candidate, chosen over unbounded historical pagination.
- Treat 404 or channel reclassification as confirmed withdrawal; treat transient or malformed revalidation as unknown and suppress only that channel until retry.
- Keep the 8 MiB and 100-entry guards even with a trusted host, because the remote response remains outside fleetd’s control.
- Use subtract-only jitter so simultaneous restarts spread out without exceeding the one-hour maximum interval.
- Reject negative and sub-five-minute nonzero overrides so jitter cannot truncate into a busy loop.
- Retain cached release data on failure for later retry, but exclude unverified candidates from downstream offers.
- Warn once after three failed exact-tag revalidations instead of dropping a candidate that has not been confirmed withdrawn.
- Cache baseline responses by ETag and support optional authentication; unauthenticated polling remains functional for zero-config installs.
- Derive RC classification and release-note links from canonical tags and the fixed repository rather than trusting upstream presentation fields.
- Recover per polling iteration so the next cycle continues; recovered programming defects log at Error with a stack trace while ordinary fetch failures remain Debug-only.
- Emit generic configuration errors and strip request URLs from transport errors so rejected configuration cannot leak.
- Use `cargo clean --release -p asicrs-plugin` after dependency warmup so Cargo owns package invalidation while retaining the warmed dependency graph.

## Testing & validation

- `go test -race ./internal/domain/updates -count=1`
- `golangci-lint run ./internal/domain/updates`
- Exact-tag tests cover confirmed withdrawal, reclassification, transient failure, tag mismatch, and trailing JSON.
- Cache tests cover authenticated `304` reuse, stable/RC preservation, rate-limit reset suppression, and independent channel degradation.
- Lifecycle tests cover per-cycle panic containment and continuation.
- Required GitHub checks pass for the updated head.


docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Implementation Units
+
+### U1. Updates proto contract and generated code
+
+- **Goal:** Define the `updates.v1.UpdatesService` Connect RPC contract and regenerate Go/TS code.
+- **Requirements:** R2, R5, R6, R9 (contract surface).
+- **Dependencies:** none.
+- **Files:** `proto/updates/v1/updates.proto` (new); regenerated `server/generated/grpc/updates/v1/**` and `client/src/protoFleet/api/generated/updates/v1/**` via `just gen` (never hand-edited).
+- **Approach:** Mirror `proto/alerts/v1/alerts.proto` for file/service/message structure; take the `option idempotency_level` idiom from `proto/ping/v1/ping.proto` and `proto/serverlog/v1/serverlog.proto` (`NO_SIDE_EFFECTS` on reads — alerts.proto carries none). Service: `GetVersion` → `{current_version}`; `GetUpdateStatus` → `{current_version, channel, latest_eligible (ReleaseInfo), update_available, install_command}`; `SetReleaseChannel(channel)`. `ReleaseChannel` enum: `STABLE`, `STABLE_AND_RC` (+ unspecified zero value per proto style); `ReleaseInfo`: `version`, `release_notes_url`, `published_at`, `prerelease`.
```
ankitgoswami: Can we remove GetVersion? GetUpdateStatus already returns current_version

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
```diff
@@ -0,0 +1,361 @@
...
+- **Permission `instance:update`, SUPER_ADMIN-held, ADMIN-excluded.** One catalog key covers seeing prompts, changing the channel, and (phase 2) triggering upgrades. `SUPER_ADMIN` acquires it automatically via `ReconcileFull`/`AllPermissions()`; it is explicitly excluded from `adminSeedPermissions()` (mirroring the existing `role:manage` exclusion) so ADMIN roles opt in through the existing role editor. No seed migration is needed: boot-time `authz.Reconcile` upserts every catalog entry before the listener starts, so existing databases receive the `instance:update` row from the `catalog.go` change alone (the 000088-style seed migrations existed for their ADMIN `role_permission` backfills — the exact grant this feature deliberately omits). The ADMIN exclusion is forward-looking either way: additive reconcile never re-asserts seeds on existing orgs, so upgraded deployments' ADMIN roles gain the key only through the role editor — which is the intent.
```
ankitgoswami: I thought we couldn't edit roles via the UI. I don't get the opt-in mechanism mentioned here.

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
```diff
@@ -0,0 +1,361 @@
...
...ors before logging. Checker implements `runtimejobs.Lifecycle` mirroring the diagnostics closer (idempotent Start/Stop, ticker + subtract-only jitter, one non-blocking startup check). HTTP via stdlib client (10s timeout, per `server/internal/domain/alerts/grafana_client.go`), headers: identifying `User-Agent` (`fleetd/<version>`), `X-GitHub-Api-Version`, `If-None-Match` from cached ETags. Each cycle independently fetches `GET .../releases/latest` (optional stable fallback) and paginated `GET .../releases?per_page=100` (authoritative stable and RC candidates), so a fallback failure does not suppress list-derived offers; stream-decode at most 100 raw entries per page and retain at most 1,000 entries across the ten-page crawl; require exact stable `vX.Y.Z` or RC `vX.Y.Z-rc.N` grammar plus `semver.IsValid`, and require stable releases to have GitHub's `prerelease` flag unset; semver-max each channel across the complete bounded set; store snapshot under mutex. Any release-list failure (network, non-2xx, rate limit, malformed JSON, or more than 100 entries on one page) retains the previous snapshot and logs at `slog.Debug` only; 304 reuses cache; malformed individual entries are skipped.
```
ankitgoswami: I feel like we could start with something simpler here. Like using fixed Proto Fleet GitHub URLs and checking once every 24 hours. Fetch /releases/latest for stable and only /releases?per_page=100&page=1 for RCs, with a response-size limit followed by an ordinary JSON decode.

If I understand correctly, the ten-page crawl mainly protects against a hypothetical future where a relevant RC is buried behind hundreds of nightlies?

Also, with 2 requests/day we could avoid the ETag stuff I think.

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
```diff
@@ -0,0 +1,361 @@
...
...e client bundle); when a newer eligible release exists — its version, release-notes link, and the install command with the same copy control U5 uses (independent of the nav callout's dismissal state, so a dismissed prompt's command stays retrievable); an explicit up-to-date state ("You're on the latest version") when `update_available` is false or no snapshot exists yet; skeleton on initial load and Network.tsx-style error handling when the status RPC fails; and a Stable / Stable + RC control (matching nearby settings idiom) that calls `SetReleaseChannel`. Include one line of helper copy on the RC option noting RC installs cannot downgrade until the next stable. Registering a settings page takes three steps, all required: (1) export the lazy importer in `routePrefetch.ts` **and** append it to the `settingsRoutePrefetch` list; (2) add a `createRoute("/settings/updates", ...)` entry wrapped in `SettingsLayout` in `router.tsx`; (3) add the `secondaryNavItems` entry in `navItems.ts` with `parent: "/settings"`, a `section`, and `requiredPermission: 'instance:update'` — mirror the Integrations entry (`apikey:manage`). The settings tab strip renders from `secondaryNavItems` automatically.
```
ankitgoswami: We should probably have a "Update status unavailable" state for when github goes down.

server/internal/domain/updates/github.go
```diff
@@ -0,0 +1,265 @@
...
+		userAgent:  "fleetd/" + serverVersion,
+		logger:     logger,
+		httpClient: &http.Client{Timeout: githubHTTPTimeout},
+	}
+}
+
+// fetchLatestStable returns the newest non-prerelease release, exactly as
+// GitHub's /releases/latest endpoint defines it.
+func (c *githubClient) fetchLatestStable(ctx context.Context) (githubRelease, error) {
+	resp, err := c.get(ctx, c.baseURL+"/releases/latest", c.latestETag)
```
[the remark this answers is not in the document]
ankitgoswami: ah, good catch

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
mcharles-square: but Github never goes down... :)

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
mcharles-square: the wording here is weird but effectively admin here is a short hand for a group of permissions. from the ui you could create a role with the same as admin plus the `instance:update`

docs/plans/2026-07-27-001-feat-release-update-notifications-plan.md
mcharles-square: I felt like 24hrs is tool long. I generally agree that this got a bit over engineered due to the fact that we dont have clearly documented release practices for things like patch releases (gh `releases/latest` is creation time based).

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	runCtx, cancel := context.WithCancel(ctx)
+	done := make(chan struct{})
+	c.cancel = cancel
+	c.done = done
+	go func() {
+		defer close(done)
+		c.run(runCtx)
+	}()
```
rongxin-liu, line 153: If a check cycle ever panics (e.g. a nil deref on an odd API response), it takes the whole server down, and this loop parses external input on a timer, forever. Worth adding a `defer recover()` (with a Debug log) inside this goroutine before a later unit wires it into server startup? A background nicety should never be able to crash the server.

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+// check runs one fetch cycle. A list failure retains the previous release data
+// but marks both channels unavailable. The latest endpoint is only a stable
+// fallback, so its failure must not hide releases discovered from the
+// authoritative list page. Cached releases that age out of those bounded
+// responses are revalidated by exact tag before they remain eligible. Nothing
+// is logged above Debug — update notification is best-effort and must never
+// look like a server problem.
+func (c *Checker) check(ctx context.Context) {
+	latest, latestErr := c.client.fetchLatestStableFallback(ctx)
+	list, err := c.client.fetchReleases(ctx)
```
rongxin-liu, line 234: Steady state is 2 unauthenticated requests per cycle per server, drawn from GitHub's shared 60/hr/IP pool, which I think is fine for one deployment, but several fleets behind the same NAT could exhaust it and all quietly go dark (failures only log at Debug). Probably fine to merge as-is since nothing runs the checker yet, but maybe leave a TODO here (conditional requests via ETag, or optional token support) so it's settled before a later unit enables this by default.

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+		isEligibleRCRelease,
+	)
+
+	snapshot := Snapshot{
+		LatestStable:    stable,
+		LatestRC:        rc,
+		FetchedAt:       time.Now(),
+		StableAvailable: stableComplete && stable != nil,
+		RCAvailable:     rcComplete,
+	}
```
rongxin-liu, line 269: Two things maybe worth pinning down here before downstream units start consuming the snapshot:

- The flags are asymmetric: `StableAvailable` includes `stable != nil`, but `RCAvailable` can be true while `LatestRC` is nil. Should we make them symmetric, or document what each means ("channel data is trustworthy" vs "a release exists")?
- Using this safely requires checking flag + pointer + `FetchedAt.IsZero()` together. Perhaps add small accessors (e.g. `Snapshot.EligibleStable() (*Release, bool)`) or a doc comment spelling out the invariants, so the API-layer unit can't misread it.

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+	if cached == nil ||
+		(current != nil && semver.Compare(cached.Version, current.Version) <= 0) {
+		return current, true
+	}
+
+	rel, found, err := c.client.fetchReleaseByTag(ctx, cached.Version)
+	if err != nil {
+		c.logger.Debug("cached release revalidation unavailable", "channel", channel, "error", err)
+		return cached, false
+	}
```
rongxin-liu, line 300: If revalidation keeps erroring (persistent rate limit, odd 5xx), the cached candidate is retained forever and the channel stays unavailable indefinitely that could quietly suppressing all update notifications for that channel. Maybe bound consecutive failures (after N misses, drop the cached candidate and fall back to the verified `current`), or at least log when the checker is stuck in this state?

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+	return stableTagPattern.MatchString(tag) && semver.IsValid(tag)
+}
+
+func isCanonicalRCTag(tag string) bool {
+	return rcTagPattern.MatchString(tag) && semver.IsValid(tag)
+}
+
+func isCanonicalReleaseTag(tag string) bool {
+	return isCanonicalStableTag(tag) || isCanonicalRCTag(tag)
+}
```
rongxin-liu: I suppose this is for the next PRs but sometimes AI agent would keep flagging it, so maybe add a short comment pointing at the consuming unit so it doesn't land as unexplained dead code and get swept by a future cleanup pass.

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+
+func newRelease(rel githubRelease) *Release {
+	return &Release{
+		Version:     rel.TagName,
+		NotesURL:    safeNotesURL(rel.HTMLURL),
+		PublishedAt: rel.PublishedAt,
+		Prerelease:  rel.Prerelease,
+	}
+}
```
rongxin-liu, line 402: `Prerelease` copies GitHub's flag verbatim, but channel selection is grammar-based, so a `v1.2.3-rc.1` that isn't marked prerelease on GitHub flows through with `Prerelease: false`. Do we want to derive it from the tag instead (`Prerelease: isCanonicalRCTag(rel.TagName)`) so the field always agrees with the channel the release was selected for?

server/internal/domain/updates/checker.go
```diff
@@ -0,0 +1,345 @@
...
+// misconfigured releases API could otherwise smuggle a javascript: href into
+// the UI. Anything short of https is dropped and the client hides the link.
+// This extends Config.Validate's https-only stance to body-derived fields.
+func safeNotesURL(raw string) string {
+	u, err := url.Parse(raw)
+	if err != nil || u.Scheme != "https" || u.Host == "" {
+		return ""
+	}
+	return raw
+}
```
rongxin-liu: https-only blocks `javascript:` hrefs, but this still accepts any https host (including userinfo tricks like `https://github.com@something.example/...`), and the client renders it as a clickable link. Since eligible releases always come from the configured repo, should we also require `u.Host == "github.com"` (or derive the allowed host from config) and reject URLs carrying userinfo?

rongxin-liu (approved):

PR #842, opened 2026-07-30
feat(updates): add update contract and release channel state

mcharles-square:
Reviewable diff: +146/-24 across 9 files (excludes generated, test, and story files).

## Summary

Defines the additive `InstanceUpdateService` contract, persists one release-channel choice per organization, and introduces the `instance:update` authorization key. No handler or client consumes the contract yet, so merging this PR exposes no incomplete experience.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841) (merged) → [#842](https://github.com/block/proto-fleet/pull/842) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → [#837](https://github.com/block/proto-fleet/pull/837) → [#838](https://github.com/block/proto-fleet/pull/838) → [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840). This is 2/5 of phase one and is now relative to `main`, which contains #841’s release-discovery engine. Backend delivery, client surfaces, and host execution remain downstream.

## How it works

Migration 000131 creates a constrained organization-scoped release-channel row, and sqlc owns all reads and upserts. Boot-time authorization reconciliation grants `instance:update` to `SUPER_ADMIN` while built-in `ADMIN` remains unchanged; custom roles can opt in. `GetUpdateStatus` returns the running version, eligible release, command, channel, and explicit discovery availability. `SetReleaseChannel` changes the persisted channel. The redundant `GetVersion` RPC is intentionally absent.

```mermaid
flowchart LR
  P["InstanceUpdateService proto"] --> G["Generated Go and TypeScript"]
  M["Migration 000131"] --> Q["Prepared sqlc queries"]
  A["instance:update catalog entry"] --> R["SUPER_ADMIN reconciliation"]
  G --> U["Consumed by #843"]
  Q --> U
  R --> U
```

## Areas of the code involved

| Area / package / file | What changed | Why it matters for review |
|---|---|---|
| `proto/instance/v1/updates.proto` | Status/channel RPCs and release messages | Public compatibility contract |
| Generated Go, Connect, TypeScript | Regenerated bindings | Generated — skip |
| `server/migrations/000131_*` | Release-channel table and rollback | Schema constraints and reversibility |
[...]

PR #843, opened 2026-07-30
feat(updates): serve release update status

mcharles-square:
Reviewable diff: +420/-20 across 12 files (excludes generated, test, and story files).

## Summary

Wires release discovery into fleetd and implements the permission-gated `InstanceUpdateService` backend for update status and release-channel selection. The API distinguishes an unavailable comparison from an up-to-date instance, invalidates availability after polling failures or recovered panics, and only serves offers from verified channel data. Existing clients make no calls yet, so this remains merge-safe.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841) (merged) → [#842](https://github.com/block/proto-fleet/pull/842) (merged) → [#843](https://github.com/block/proto-fleet/pull/843) → [#844](https://github.com/block/proto-fleet/pull/844) → [#845](https://github.com/block/proto-fleet/pull/845) → [#835](https://github.com/block/proto-fleet/pull/835) → [#836](https://github.com/block/proto-fleet/pull/836) → [#837](https://github.com/block/proto-fleet/pull/837) → [#838](https://github.com/block/proto-fleet/pull/838) → [#839](https://github.com/block/proto-fleet/pull/839) → [#840](https://github.com/block/proto-fleet/pull/840). This is 3/5 of phase one and is relative to `main`, which contains #841’s release-discovery engine and #842’s API contract, release-channel persistence, and authorization key. Operator surfaces and host execution remain downstream.

## How it works

fleetd runs one release checker and exposes its verified snapshot through the updates domain service. A failed fetch or recovered panic retains cached metadata but marks it ineligible until a successful cycle; `UPDATES_ENABLED=false` disables polling in packaged deployments. For each status request, the service loads the organization’s channel, requires a semver-valid running version, and uses only verified candidates. Stable+RC can offer a verified RC while stable discovery is unavailable only when that RC is newer than the running version; it cannot claim the instance is current from a partial snapshot.

The handler maps that result into `InstanceUpdateService`, requires an authenticated session plus org-wide `instance:update`, and redacts update-status response bodies from debug logs. Installer commands are composed only after rechecking the exact download allowlist and canonical tag. GitHub polling remains anonymous at the low, jittered hourly cadence, with ETag caching and rate-limit-reset backoff.

```mermaid
flowchart LR
  C["Verified checker snapshot"] --> S["Updates domain service"]
  O["Organization channel"] --> S
  V["Running fleetd version"] --> S
  S --> H["Session and instance:update gate"]
  H --> R["Redacted status response or channel mutation"]
```

## Areas of the code involved

|---|---|---|
| `server/internal/domain/updates` | Panic/failure invalidation, channel gating, semver comparison, anonymous polling, command composition | Primary availability and security logic |
| `server/internal/handlers/updates` | Authorization, protobuf mapping, and both channel mappings | Org isolation and response fidelity |
| `server/internal/handlers/interceptors` | Session-only classification and response redaction | Authentication and logging boundary |
| `server/cmd/fleetd` | Checker lifecycle and service wiring | Runtime behavior |
| `deployment-files/docker-compose.yaml` | Forwards `UPDATES_ENABLED` with a true default | Operator control of outbound polling |
| `docs/plans/` | Records the conservative partial-snapshot and panic semantics | Architectural contract |

- Retain cached release metadata after fetch failures or panics, but make it ineligible until the affected view recovers.
- Treat non-semver running builds as status-unavailable rather than a successful no-update result.
- Allow an RC-only partial snapshot to offer a newer verified RC, but never to claim the instance is up to date.
- Re-check the exact download-base allowlist and canonical tag during command construction.
- Poll anonymously at the low hourly cadence; defer token configuration until observed rate limits justify the extra secret-management surface.
- Keep the running fleetd version in `GetUpdateStatus`, eliminating the separate `GetVersion` procedure.

## Testing & validation

- `go test ./internal/domain/updates -count=1`
- `go test ./internal/handlers/updates ./cmd/fleetd -count=1`
- Targeted update interceptor tests passed; the unrelated database-backed interceptor suite timed out locally waiting for its test database.
- `golangci-lint run ./internal/domain/updates ./internal/handlers/interceptors ./internal/handlers/updates ./cmd/fleetd`
- `UPDATES_ENABLED=false docker compose --project-directory . -f deployment-files/docker-compose.yaml config --format json` forwards `"false"` to fleet-api.
- Pre-push server lint passed.


server/cmd/fleetd/config.go
```diff
@@ -54,6 +55,7 @@ type Config struct {
 	Plugins        plugins.Config               `embed:"" prefix:"plugins-" envprefix:"PLUGINS_"`
 	IPScanner      ipscanner.Config             `embed:"" prefix:"ipscanner-" envprefix:"IPSCANNER_"`
 	Diagnostics    diagnostics.Config           `embed:"" prefix:"diagnostics-" envprefix:"DIAGNOSTICS_"`
+	Updates        updates.Config               `embed:"" prefix:"updates-" envprefix:"UPDATES_"`
```
rongxin-liu, line 58: Should we also forward `UPDATES_ENABLED` through `deployment-files/docker-compose.yaml`? The packaged service currently omits it, so operators cannot disable polling even though this embedded config exposes the switch.

server/internal/domain/updates/service.go
```diff
@@ -0,0 +1,222 @@
...
+func (s *Service) GetUpdateStatus(ctx context.Context, organizationID int64) (UpdateStatus, error) {
+	channel, err := s.releaseChannel(ctx, organizationID)
+	if err != nil {
+		return UpdateStatus{}, err
+	}
+	snapshot := s.snapshots.Snapshot()
+	status := UpdateStatus{
+		CurrentVersion:  s.currentVersion,
+		Channel:         channel,
+		StatusAvailable: channelStatusAvailable(channel, snapshot),
```
rongxin-liu: Since this status now trusts the checker's availability flags, perhaps `checkSafely` should probably clear `StableAvailable` and `RCAvailable` when it recovers a panic. Otherwise a failed refresh can keep serving a stale cached offer; worth adding a prime -> panic -> unavailable -> recover test.

server/internal/domain/updates/service.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if !status.StatusAvailable {
+		return status, nil
+	}
+
+	if !semver.IsValid(s.currentVersion) {
+		s.logger.Debug("update check skipped: running version is not semver", "version", s.currentVersion)
+		return status, nil
```
rongxin-liu, line 97: Worth setting `StatusAvailable` to false before returning for non-semver builds? As written, `dev` or nightly builds can look like a successful no-update result rather than an unavailable comparison.

server/internal/domain/updates/service.go
```diff
@@ -0,0 +1,222 @@
...
+// when stable discovery is unavailable, but an incomplete RC view makes that
+// broader channel unavailable.
+func channelStatusAvailable(channel Channel, snap Snapshot) bool {
+	_, stableAvailable := snap.EligibleStable()
+	rc, rcAvailable := snap.EligibleRC()
+	switch channel {
+	case ChannelStable:
+		return stableAvailable
+	case ChannelStableAndRC:
+		return rcAvailable && (stableAvailable || rc != nil)
```
rongxin-liu: Maybe an RC-only partial snapshot should be considered available only when the verified RC is newer than the running version. If stable discovery failed and the RC is equal or older, this can report up to date without knowing the latest stable.

server/internal/domain/updates/service_test.go
```diff
@@ -0,0 +1,523 @@
...
+// A candidate outside the canonical stable/RC grammar must never yield an
+// install command, and update_available stays false — even though the checker
+// should never cache such a release.
+func TestNoncanonicalCandidateNeverOffered(t *testing.T) {
+	t.Parallel()
+
+	snaps := &fakeSnapshots{snap: Snapshot{
+		// "0.3.0" (no leading v) is invalid for golang.org/x/mod/semver.
+		LatestStable:    rel("0.3.0"),
+		LatestRC:        rc("nightly-20260727-abc"),
```
rongxin-liu: Perhaps use a semver-valid, higher but noncanonical tag here, such as `v0.3.0+build.1`. `0.3.0` exits at `semver.Compare` before `installCommand`, so this service-path test likely does not reach the command guard; worth separating the untrusted-base assertion too.

server/internal/handlers/interceptors/config.go
```diff
@@ -155,6 +156,13 @@ var SessionOnlyProcedures = []string{
 	alertsv1connect.MaintenanceWindowServiceUpdateMaintenanceWindowProcedure,
 	alertsv1connect.MaintenanceWindowServiceDeleteMaintenanceWindowProcedure,
 	alertsv1connect.HistoryServiceListAlertsProcedure,
+	// The updates surface is session-only across both procedures —
+	// uniform surface, same rationale as the authz entries above. Update
+	// status and the install command describe the instance's patch level,
+	// SetReleaseChannel changes which builds every operator is offered,
+	// and no API-key automation consumes this service.
+	instancev1connect.InstanceUpdateServiceGetUpdateStatusProcedure,
```
rongxin-liu, line 167: Since this response includes patch-level metadata and a copy-paste install command, should we also add `GetUpdateStatus` to `RedactedResponseProcedures` and cover it in the config tests? Session-only auth does not prevent debug body logging.

server/internal/handlers/updates/handler_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	_, err := h.SetReleaseChannel(ctx, connect.NewRequest(&instancev1.SetReleaseChannelRequest{
+		Channel: instancev1.ReleaseChannel_RELEASE_CHANNEL_STABLE_AND_RC,
+	}))
+	require.NoError(t, err)
+	assert.Equal(t, "stable_and_rc", store.channels[7], "the session's org id must key the row")
+
+	resp, err := h.GetUpdateStatus(ctx, connect.NewRequest(&instancev1.GetUpdateStatusRequest{}))
+	require.NoError(t, err)
+	assert.Equal(t, instancev1.ReleaseChannel_RELEASE_CHANNEL_STABLE_AND_RC, resp.Msg.GetChannel())
```
rongxin-liu: Could this be table-driven for both `STABLE` and `STABLE_AND_RC`? That would prove the enum-to-storage mapping for the default stable path, not only the RC opt-in.

rongxin-liu (approved):

PR #844, opened 2026-07-30
feat(updates): add release settings page

mcharles-square:
Reviewable diff: +370/-2 across 7 files (excludes generated, test, and story files).

## Summary

Adds Settings → Updates for permission holders, showing the running server version, eligible release, exact manual install command, and Stable + RC opt-in. Update discovery failures remain distinct from the up-to-date state, while request sequencing and bounded mutation lifecycle coordination prevent stale channels or install commands during saves, navigation, and ambiguous network failures. If update authority is revoked, the page reports it, removes the stale client permission, and redirects to a settings route the operator can access; delayed failures perform global auth cleanup only for the session that originated them.

*Stack:* [#841](https://github.com/block/proto-fleet/pull/841), [#842](https://github.com/block/proto-fleet/pull/842), and [#843](https://github.com/block/proto-fleet/pull/843) are merged; this PR is relative to `main`. Remaining notification and host-execution work continues in [#845](https://github.com/block/proto-fleet/pull/845), [#835](https://github.com/block/proto-fleet/pull/835), [#836](https://github.com/block/proto-fleet/pull/836), [#837](https://github.com/block/proto-fleet/pull/837), [#838](https://github.com/block/proto-fleet/pull/838), [#839](https://github.com/block/proto-fleet/pull/839), and [#840](https://github.com/block/proto-fleet/pull/840).

## How it works

The permission-gated route loads `GetUpdateStatus`. Available status renders either the eligible release and command or the up-to-date state; unavailable discovery gets a neutral state. A channel change persists through `SetReleaseChannel` with a 30-second deadline, keeps channel and copy controls disabled, and then refetches authoritative status. Request sequence numbers ignore late status responses, while a route remount waits for any save started by its predecessor; the deadline guarantees that shared barrier settles. Ambiguous non-auth save failures, including deadlines, refetch because the server may have committed before the response was lost. Each request captures the auth session installed at login. Same-session failures still pass through global auth handling after unmount or request replacement, while obsolete page state and toasts remain suppressed; failures from a replaced session are discarded so they cannot log out or remove permissions from its successor. Mounted permission denial also shows a revocation notice and redirects through the permission-aware settings landing helper.

```mermaid
sequenceDiagram
  participant O as Operator
  participant U1 as Updates page
  participant S as InstanceUpdateService
  participant U2 as Remounted page
  O->>U1: Choose Stable or RC
  U1->>U1: Disable channel and copy controls
  U1->>S: SetReleaseChannel with 30s deadline
  U1--xU1: Unmount
  U2->>U2: Wait for bounded in-flight save
  S-->>U2: Save settles or deadline expires
  U2->>S: GetUpdateStatus
  S-->>U2: Authoritative channel and command
```

## Areas of the code involved

|---|---|---|
| `features/settings/components/Updates.tsx` | Status states, request sequencing, bounded save lifecycle coordination, reconciliation, session-bound auth cleanup, permission invalidation, safe fallback routing, release notes, and command copy | Primary operator workflow, authorization response, and concurrency boundary |
| `features/updates/copyInstallCommand.ts` | Shared clipboard feedback | Manual fallback behavior |
| `api/clients.ts` | InstanceUpdateService client | Backend connection |
| `navItems.ts`, `router.tsx`, `routePrefetch.ts` | Permission-gated lazy route and idle-time prefetch | Route completeness |
| `docs/plans/...release-update-notifications-plan.md` | Documents request ordering, bounded remount behavior, reconciliation, and session-bound auth handling | Keeps the implementation plan aligned |
| Tests and visual snapshot | Cover rendering, permissions, protobuf fixtures, mutation/refetch ordering, deadline recovery, remounts, late and cross-session auth failures, denial handling, fallback routing, and layout | Regression coverage; snapshot is generated test output |

- Sequence status requests in the component so stale success and ordinary error responses cannot replace the fresh UI.
- Bind every request to the login-installed session-expiry object so delayed session-A failures cannot mutate session B, even for the same username.
- Process same-session authentication and permission failures globally after unmount while suppressing obsolete component state and toast work.
- Read current permissions when processing same-session delayed denial so an old request cannot overwrite newer authority changes.
- Coordinate the single in-flight channel save across route instances so remounts cannot load an obsolete command, and bound it with a 30-second RPC deadline so a stuck request cannot block the page for the tab lifetime.
- Reconcile non-auth save failures because a lost response cannot prove the server rejected the mutation; skip the extra request for auth or permission failures.
- Invalidate only `instance:update` after permission denial so stale update controls disappear without discarding unrelated cached authority.
- Use the shared settings landing helper after permission loss so Network is selected only with `fleet:read`; otherwise Preferences is the safe fallback.
- Keep channel and copy controls disabled through persistence and refetch so displayed commands always match confirmed server state.
- Show unavailable separately from up to date rather than treating failed discovery as success.

## Testing & validation

- `npm run test -- --run src/protoFleet/features/settings/components/Updates.test.tsx` (28 tests)
- `npm exec eslint -- src/protoFleet/features/settings/components/Updates.tsx src/protoFleet/features/settings/components/Updates.test.tsx`
- `npm exec --no -- tsc --noEmit`
- Pre-commit formatting and pre-push client typecheck


client/src/protoFleet/features/settings/components/Updates.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+    statusAvailable: true,
+    updateAvailable: true,
+    installCommand: INSTALL_COMMAND,
+    latestEligible: {
+      version: "v1.3.0",
+      releaseNotesUrl: RELEASE_NOTES_URL,
+      prerelease: false,
+    },
+    ...overrides,
+  }) as unknown as GetUpdateStatusResponse;
```
rongxin-liu: Could be worth building this fixture with `create(GetUpdateStatusResponseSchema, ...)` and `ReleaseInfoSchema` instead of casting through `unknown`? That would probably keep it aligned with the generated protobuf shape and catch schema changes at compile time.

client/src/protoFleet/features/settings/components/Updates.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+            releaseNotesUrl: "https://github.com/block/proto-fleet/releases/tag/v1.4.0-rc.1",
+            prerelease: true,
+          } as GetUpdateStatusResponse["latestEligible"],
+        }),
+      );
+      await staleRequest.promise;
+    });
+    expect(queryByText("v1.4.0-rc.1")).not.toBeInTheDocument();
+    expect(getByRole("checkbox", { name: RC_CHECKBOX_NAME })).not.toBeChecked();
+  });
```
rongxin-liu, line 344: Perhaps worth adding a sibling case where the older request rejects after the newer request succeeds. The catch guard is separate from the success guard, so asserting that the fresh UI remains and no load error appears would cover that path.

client/src/protoFleet/features/settings/components/Updates.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+    mockGetUpdateStatus.mockResolvedValueOnce(buildStatus({ channel: ReleaseChannel.STABLE }));
+    mockGetUpdateStatus.mockReturnValueOnce(refetch.promise);
+    mockSetReleaseChannel.mockResolvedValue({} as never);
+
+    const { findByRole, getByRole } = render(<Updates />);
+    const checkbox = await findByRole("checkbox", { name: RC_CHECKBOX_NAME });
+    const copyButton = getByRole("button", { name: "Copy install command" });
+    fireEvent.click(checkbox);
+
+    await waitFor(() => expect(mockGetUpdateStatus).toHaveBeenCalledTimes(2));
```
rongxin-liu: Do we want to keep `setReleaseChannel` deferred as well as the refetch? It resolves immediately here, so this currently checks only the refetch-pending phase. A deferred save would verify both controls stay disabled throughout the mutation and subsequent refresh.

client/src/protoFleet/features/settings/components/Updates.tsx
```diff
@@ -0,0 +1,193 @@
...
+    try {
+      await instanceUpdateClient.setReleaseChannel({ channel: nextChannel });
+      setChannel(nextChannel);
+      pushToast({
+        message: "Release channel updated",
+        status: STATUSES.success,
+      });
+      // The eligible release differs per channel, so refetch to keep the
+      // offered version and install command in sync with the new channel.
+      await fetchStatus();
```
rongxin-liu: Perhaps we could make this mutation lifecycle-aware and have remounts wait for any in-flight save before loading status. Right now looks like navigating away mid-save can still toast/refetch after unmount, while a quick remount can fetch the old channel and expose a stale install command.

client/src/protoFleet/features/settings/components/Updates.tsx
```diff
@@ -0,0 +1,193 @@
...
+    } catch (err) {
+      handleAuthErrors({
+        error: err,
+        onError: () => {
+          pushToast({
+            message: getErrorMessage(err, "Failed to update release channel"),
+            status: STATUSES.error,
+          });
+        },
+      });
```
rongxin-liu: Worth separating mutation failures from refresh failures and reconcile status after non-auth mutation errors? A lost response after the server commits leaves the checkbox and install command stale, while a failed refetch after a successful save is reported as a save failure. We could refetch after ambiguous save errors and test both sequences.

rongxin-liu (approved):

PR #853, opened 2026-07-31
Sync curtailment fallback configuration to Proto rigs

negarn:
Reviewable diff: +1404/-53 across 34 files (excludes generated, test, and story files).

## Summary
This PR keeps Proto rigs' local curtailment fallback configuration aligned with the MQTT sources managed in Fleet. Settings changes and newly paired rigs durably schedule the latest complete configuration without making those user-facing workflows wait for fleet-wide command fan-out, and the Curtailment settings page explains the fallback behavior. Only fully paired Proto rigs receive broker credentials.

image

rig config matches fleet:
```
{"enabled":true,"fail_policy":"closed","restore_policy":"respect_manual_stop","nats_url":"nats://localhost:4222","mcdd_grpc_addr":"127.0.0.1:2122","status_publish_interval":"15s","providers":[{"name":"maestro","type":"maestro_mqtt","enabled":true,"brokers":["10.90.0.30","10.90.0.31"],"port":1883,"username":"operator","password":"secret","topic":"maestro/target","qos":1,"stale_after":"4m0s","reconnect_backoff":"5s"}]}
```

(tested both with and without nodes)

## How it works
When an operator creates, updates, enables, disables, or deletes an MQTT source, the source mutation and a new per-organization desired generation are committed atomically. The request returns after that durable write; server-side MQTT runtime reload is best effort, while a runtime job claims due rig-config work under a database lease, rebuilds the latest snapshot from enabled TCP MaestroOS sources, and enqueues it through the command service. Failed builds or enqueues remain pending for retry, and a newer settings write during an active claim becomes the next generation instead of racing an older snapshot.

On upgrade, the migration seeds one due reconciliation request for every organization that already has an enabled MQTT source, so existing fleets converge without requiring a settings edit or another pairing event.

Successful pairing requests the same durable reconciliation rather than launching an organization-wide goroutine. FleetNode reconnects do not create new config commands: the existing durable command queue retains work for offline devices and resumes delivery when their route is available.

The command selector admits only `PAIRED` Proto devices. Locally connected rigs receive a service-key-encrypted queue payload; FleetNode-connected rigs receive a device-bound envelope produced with ephemeral X25519, HKDF-SHA256, and AES-256-GCM. The FleetNode decrypts and validates the target binding, invokes the optional plugin capability, and the Proto plugin replaces the rig config through `PUT /api/v1/curtailment/config`.

## Diagrams

```mermaid
flowchart TD
  A["Operator changes an MQTT source"] --> B["Atomic source write and desired generation"]
  P["Rig pairs successfully"] --> C["Durable reconciliation request"]
  B --> C
  C --> D["Leased background reconciler"]
  D --> E["Build latest complete config"]
  E --> F["Select fully paired Proto rigs"]
  F --> G["Durable command queue"]
  G --> H["Local encrypted route"]
  G --> I["FleetNode device-bound encrypted route"]
  H --> J["SDK ApplyCurtailmentConfig"]
  I --> J
  J --> K["Proto PUT /api/v1/curtailment/config"]
```

```mermaid
sequenceDiagram
  participant API as MQTT settings API
  participant DB as Settings and reconciliation state
  participant Worker as Rig config reconciler
  participant Queue as Command queue
  participant Rig as Proto rig
  API->>DB: Commit source change and generation G
  DB-->>API: Durable success
  Worker->>DB: Lease latest due generation
  Worker->>DB: Read latest enabled sources
  Worker->>Queue: Enqueue replacement config for paired Proto rigs
  Queue-->>Worker: Durable enqueue complete
  Worker->>DB: Advance enqueued generation to G
  Queue->>Rig: Retry delivery until command completion
  Rig-->>Queue: Terminal failure after bounded retries
  Queue->>DB: Reopen a later desired generation
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/protoFleet/features/settings/components/Curtailment/` | Adds an informational Proto rig fallback callout and focused coverage. | Confirms the operator-facing behavior communicated by the settings page. |
| `server/migrations/000132_*` and `server/sqlc/queries/curtailment_mqtt.sql` | Adds per-organization desired/enqueued generations, retry state, and a lease; seeds existing enabled-source organizations; source mutations update it in the same SQL statement. | This is the durability, upgrade convergence, and concurrent-write ordering boundary. |
| `server/internal/domain/curtailment/mqttingest/` | Builds the complete snapshot and runs the leased retrying reconciliation lifecycle. | Owns source filtering, defaults, credential decryption, audit identity, and retry behavior. |
| `server/internal/domain/command/` and `commandtype/` | Adds config dispatch, paired-only Proto selection, route-specific payload preparation, execution, activity labels, and filter bypass. | Main orchestration and credential-admission path. |
| `server/internal/domain/fleetnode/curtailmentconfig/` | Adds the device-bound X25519/HKDF/AES-GCM envelope. | Security boundary for MQTT credentials crossing the FleetNode transport. |
| Pairing services and FleetNode gateway | Pairing records durable work; reconnect no longer starts organization-wide fan-out. | Prevents request-path amplification and repeated reconnect pushes. |
| `server/internal/domain/miner/`, `server/internal/domain/plugins/`, and `server/cmd/fleetnode/` | Carries encrypted payloads through local and remote miner implementations. | Enforces the distinction between local plugin dispatch and FleetNode transport. |
| `server/sdk/v1/` and `server/sdk/v1/pb/driver.proto` | Adds the optional SDK config capability and gRPC conversions. | Defines the plugin contract while allowing unsupported drivers to return `Unimplemented`. |
| `plugin/proto/` and `plugin/asicrs/` | Proto replaces the rig-local config over HTTP; asicrs explicitly reports unsupported. | Device-specific behavior and compatibility boundary. |
| `proto/fleetnodegateway/v1/fleetnodegateway.proto` | Adds the encrypted FleetNode command action. | Wire contract for delivering secrets without plaintext node transport. |
| Generated SQL and protobuf outputs | Regenerated from SQL and protobuf sources — generated, skip. | Required artifacts; review the source migration, queries, and protos instead. |

- Persist a latest-value generation outbox instead of fan-out in the settings request, so an API error no longer ambiguously follows a committed source change and queue outages remain retryable.
- Serialize organizations with a database lease and preserve an active lease when a newer generation arrives, so older snapshots enqueue before their trailing replacement across concurrent writers and fleetd replicas.
- Let the existing per-device command queue own normal execution retries, and atomically reopen the organization generation when a config command becomes terminal or is reaped, rather than duplicating command state in the settings domain.
- Remove reconnect-triggered reapply because queued FleetNode commands survive disconnects; pairing still schedules work so devices created after the latest settings generation receive the current snapshot.
- Restrict config commands to `PAIRED` Proto rigs instead of inheriting the general command default that also admits `DEFAULT_PASSWORD`, keeping broker credentials away from factory-password devices.
- Include enabled TCP MaestroOS sources only because the current rig API does not support the configured TLS transport; excluded sources remain active in Fleet's server-side runtime.
- Encrypt each FleetNode payload with an ephemeral X25519-derived AES-GCM key and bind it to the device identifier instead of sending MQTT credentials in plaintext or allowing cross-device replay.
- Model rig configuration as an optional SDK capability instead of expanding the required plugin surface; unsupported drivers fail explicitly without blocking existing plugins.

## Testing & validation

- Ran `just gen`, including protobuf lint and regenerated Go, TypeScript, Python, and SQL outputs.
- Passed client ESLint/Prettier validation from `just lint`; earlier focused TypeScript checks, Curtailment settings tests, and the ProtoFleet production build remain green for the unchanged client portion.
- Passed server `golangci-lint`, compiled every server package, and passed focused MQTT reconciliation, generation-coalescing, paired-only selector, terminal-failure requeue, command dispatch, pairing, handler, and runtime-job tests.
- Passed the existing Proto plugin config replacement test and plugin build, plus Python SDK tests, strict MyPy, and Ruff, for the unchanged plugin/SDK portion.
- Database-backed server tests could not run against the local TimescaleDB instance because its configured `fleet` credentials were rejected; SQLC generation validated the migration/query shape, but the new migration was not integration-tested locally.
- Rust SDK and asicrs checks were not run locally because Cargo is unavailable; CI provides their format, Clippy, build, and test coverage.
- E2E testing was not run.


server/internal/domain/curtailment/mqttingest/settings_service.go
```diff
@@ -311,6 +328,119 @@ func (s *SettingsService) TestConnection(ctx context.Context, req TestSourceConn
...
+		if reconcileErr != nil {
+			slog.Error("runtime reload and Proto rig config delivery both failed after MQTT settings write",
+				"org_id", orgID,
+				"runtime_error", reconcileErr,
+				"delivery_error", applyErr,
+			)
+		}
+		return applyErr
+	}
+	return reconcileErr
```
rl-block: This puts the full org-wide fan-out (per-source decrypts, per-rig encrypts, queue inserts) inside the request path, and by the time it runs the settings row has already committed, so a delivery failure comes back as an API error for a change that is already durable, and a retried `Create` would then hit the name-uniqueness error. Since there's no durable retry, a rig keeps stale credentials until an unrelated trigger (pairing/reconnect) happens to fire.

Should we decouple rig delivery from the mutation result (async/best-effort like the reapply path) and add a reconciliation pass so failed deliveries eventually converge?

server/internal/domain/curtailment/mqttingest/settings_service.go
```diff
@@ -311,6 +328,119 @@ func (s *SettingsService) TestConnection(ctx context.Context, req TestSourceConn
...
+	if s.rigConfigApplier == nil {
+		return nil
+	}
+	applyCtx, cancel := context.WithTimeout(detachedContext(ctx), s.reconcileTimeout)
+	defer cancel()
+	config, err := s.buildRigCurtailmentConfig(applyCtx, orgID)
+	if err != nil {
+		return err
+	}
+	return s.rigConfigApplier.ApplyCurtailmentConfigToProtoRigs(applyCtx, config)
```
rl-block: Looks like nothing serializes the sources read against the dispatch, so two concurrent settings writes can interleave: A builds its config, B persists and dispatches, then A dispatches, A's stale snapshot is enqueued last and wins (per-device queue delivery is `created_at`-ordered, and the envelope carries no per-org generation).

The fleetnode pairing service already has a per-org mutex + trailing-run pattern for exactly this; it might be worth reusing it here, or adding a config generation number.

server/internal/handlers/fleetnode/gateway/handler.go
```diff
@@ -706,6 +706,7 @@ func (h *Handler) ControlStream(ctx context.Context, stream *connect.BidiStream[
 	}}); sendErr != nil {
 		return fleeterror.NewInternalErrorf("send accepted: %v", sendErr)
 	}
+	h.pairing.FleetNodeConnectedBestEffort(ctx, subject.FleetNodeID, subject.OrgID)
```
rl-block: Every ControlStream reconnect triggers a full-org config push, so a flapping node (reconnect backoff caps around 30s) would re-push to every Proto rig in the org indefinitely, even if it owns no Proto rigs. The per-org coalescing bounds concurrency but not repeat rate. Maybe we could scope the reconnect reapply to the reconnecting node's devices, or add a debounce / config-version no-op check?

server/internal/domain/pairing/service.go
```diff
@@ -1381,6 +1389,18 @@ func (s *Service) pairDevices(ctx context.Context, r *pb.PairRequest, allowAllFa
...
+	if s.rigConfigReapplier == nil {
+		return
+	}
+	baseCtx := context.WithoutCancel(ctx)
+	go func() {
+		reapplyCtx, cancel := context.WithTimeout(baseCtx, rigConfigReapplyTimeout)
+		defer cancel()
+		s.rigConfigReapplier(reapplyCtx, orgID, userID)
+	}()
+}
```
rl-block, line 1396: Unlike the coalesced reapply in `fleetnode/pairing`, this spawns an unguarded goroutine per call, bulk pairing fans out N concurrent org-wide dispatches (full source reads, per-source decrypts, per-rig encrypts, queue inserts). Perhaps we can do the same per-org coalescing here?

server/internal/domain/command/service.go
```diff
@@ -1645,6 +1734,38 @@ func (s *Service) Uncurtail(ctx context.Context, deviceSelector *pb.DeviceSelect
...
 
+// ApplyCurtailmentConfigToProtoRigs replaces the rig-local fallback config on
+// every paired Proto rig in the caller's organization. Operators do not manage
+// a separate coverage list.
+func (s *Service) ApplyCurtailmentConfigToProtoRigs(ctx context.Context, config sdk.CurtailmentConfig) error {
+	allProtoRigs := &pb.DeviceSelector{
+		SelectionType: &pb.DeviceSelector_AllDevices{
+			AllDevices: &pb.DeviceFilter{Manufacturers: []string{"Proto"}},
+		},
+	}
```
rl-block: With `PairingStatus` unset, this selector inherits the default `PAIRED + DEFAULT_PASSWORD` set, and the rig's curtailment-config PUT isn't gated by the default-password check, so the plaintext broker credentials get delivered to rigs Fleet has flagged as still on factory-default passwords. Should we narrow this selector to `PAIRED` only?

rl-block (approved):

PR #874, opened 2026-08-03
HA: make command recovery restart-safe

ankitgoswami:
Reviewable diff: +649/-505 across 18 files (excludes generated, test, and story files).

## Summary

Fleet now recovers command processing safely after an active-instance restart. Commands that never left the queue remain eligible for execution, while commands interrupted after dispatch began are failed because their device outcome is unknown. Ownership loss ends the active runtime instead of attempting an in-process demotion and reacquisition.

Standalone Fleet keeps its existing admission and graceful-shutdown behavior.

**Stack**

- [#874 - make command recovery restart-safe](https://github.com/block/proto-fleet/pull/874) **(current)**
- [#875 - wire the opt-in production runtime](https://github.com/block/proto-fleet/pull/875)

#875 is stacked directly on this PR and adds the production configuration that selects HA or standalone mode. This PR defines command recovery and active-runtime lifecycle behavior without changing how deployments enable HA.

## How it works

Patroni selects the writable PostgreSQL server and publishes it through etcd. Fleet separately verifies that writer and acquires its database lease before it starts active-only work.

On startup or failover:

1. The coordinator verifies the current database writer and acquires the Fleet lease.
2. It renews the lease once more after the closing writer proof, then publishes the active lifetime.
3. Command execution starts by running the existing reaper once before workers start.
4. The reaper preserves `PENDING` rows because no device attempt has started. It changes interrupted `PROCESSING` rows to `FAILED` with an unknown-outcome reason, then finishes affected terminal batches in bounded pages.
5. Active jobs start and the request gate opens.
6. The normal periodic reaper continues handling stale `PROCESSING` work.

On ownership loss, the coordinator cancels the active lifetime and returns a fatal error. The runtime closes admission, aborts active jobs, and returns that error to `fleetd`. The deployment supervisor can then restart Fleet as a fresh passive candidate.

Command writes use expected-state updates such as `PROCESSING` to `SUCCESS`. A late result therefore affects zero rows after recovery has already failed that command and is treated as a benign lost race.

Command enqueue uses one bounded transaction and a bulk insert. If the client sees an ambiguous enqueue error, Fleet checks whether the transaction committed and only fails the batch when it can prove no queue rows exist.

Recovered `DownloadLogs` batches no longer depend on the callback from the old process. Once the caller's organization-scoped batch is finished, the download endpoint lazily creates the missing log bundle from persisted results. First-time creation is serialized so concurrent requests reuse one complete bundle.

```mermaid
flowchart LR
    DCS["etcd publishes Patroni primary"] --> OBS["Fleet verifies writer"]
    PG["Connected PostgreSQL identity"] --> OBS
    OBS --> LEASE["Acquire and renew Fleet lease"]
    LEASE --> REAP["Run startup reaper"]
    REAP --> PENDING["Keep PENDING queued"]
    REAP --> PROCESSING["Fail interrupted PROCESSING"]
    PENDING --> JOBS["Start active jobs"]
    PROCESSING --> JOBS
    JOBS --> GATE["Open request gate"]
```

```mermaid
sequenceDiagram
    participant Old as Former active Fleet
    participant Lease as Fleet lease
    participant New as Replacement Fleet
    participant DB as Command tables
    participant Supervisor as Service supervisor

    Old->>Lease: Renewal fails
    Old->>Old: Close admission and abort active jobs
    Old->>Supervisor: Return fatal ownership error
    New->>Lease: Acquire ownership
    New->>DB: Fail PROCESSING and preserve PENDING
    New->>New: Start jobs and open admission
    Supervisor->>Old: Restart as passive candidate
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/ha/` | Makes active ownership loss terminal, simplifies activation to a final lease renewal, and keeps standalone shutdown graceful | Review ownership lifecycle, cancellation, and the activation race handling |
| `server/internal/domain/command/` | Runs bounded recovery through the existing reaper before workers start and treats zero-row state transitions as lost races | Review the `PENDING` versus `PROCESSING` policy and result consistency |
| `server/internal/infrastructure/queue/` | Bulk-enqueues messages in one bounded transaction and reconciles ambiguous failures | Review atomicity and error classification |
| `server/internal/infrastructure/files/`, command service and handler | Authorizes bundles by organization and command type, then serializes lazy creation after process restart | Review tenant isolation, idempotency, and the finished-batch boundary |
| `server/sqlc/queries/` | Adds bounded reaper queries and expected-state transition predicates | Review row locking, page limits, and status guards |
| `server/internal/infrastructure/db/`, `runtimepolicy/` | Applies the shared five-second transaction bound to command writes | Review timeout scope and rollback behavior |
| `server/generated/sqlc/` and generated mocks | Regenerated from SQL and interface sources | Generated, skip |
| HA, command, queue, and file tests | Covers lifecycle exit, restart recovery, enqueue ambiguity, state races, and lazy log bundles | Focus on recovery boundaries rather than helper-level assertions |

- Preserve `PENDING` work instead of failing all queued work, because no device attempt has started.
- Fail `PROCESSING` work instead of replaying it, because Fleet cannot prove whether the device already acted.
- Reuse the command reaper at startup instead of adding takeover-only cleanup and coordinator sequencing.
- Exit after ownership loss instead of implementing graceful in-process demotion and reacquisition.
- Use local expected-state SQL guards instead of advisory locks or ownership columns on command rows.
- Authorize log bundles through the persisted organization and `DownloadLogs` type before accessing local files.
- Accept a failover edge case in per-device ordering. Failing an interrupted `PROCESSING` row can release a later command before an already-issued request from the former process is provably quiesced. Device-side fencing or a quarantine and reconciliation workflow is out of scope for this PR.
- Accept the rare crash window between batch creation and enqueue. A batch with no queue rows can remain pending after a hard process crash; closing it would require another recovery protocol for little operational value.
- Do not reconstruct process-local completion activity after restart. Persisted command history remains authoritative.

## Testing & validation

- Focused HA, command, queue, file, runtime-job, handler, and `fleetd` tests passed.
- Hermit-backed server lint passed with zero issues.
- sqlc output and interface mocks were regenerated from their sources.
- Database-backed integration tests cover restart recovery, enqueue ambiguity, guarded transitions, cleanup paging, and cross-organization or wrong-type bundle rejection. They were not run locally because the available PostgreSQL credentials do not match the test harness; CI is the verification gate.
- A live multi-process Patroni and etcd failover is not covered here.

- **Logs:** watch for `active Fleet ownership ended`, `Interrupted by Fleet restart`, command-service startup failures, and repeated Fleet process exits.
- **Healthy signals:** exactly one Fleet instance is active; `PENDING` counts fall after failover; interrupted `PROCESSING` rows fail once; replacement workers resume dispatch.
- **Failure signals:** both instances remain passive, restart loops continue, or command backlogs grow after failover.
- **Mitigation trigger:** if a replacement cannot become active or command processing does not resume, disable HA and run one known-good standalone instance while lease and writer state are inspected.
- **Window and owner:** the deploying operator should observe startup, one controlled failover, and the following 24 hours.

## Follow-ups

- Add a chaos E2E that kills the primary mid-curtailment batch and asserts total time to shed stays under 180 seconds.
- Verify that the curtailment caller re-issues failed batches promptly rather than waiting on a slow poll.


server/sqlc/queries/queue.sql
```diff
@@ -67,43 +103,83 @@ WHERE m.status = 'PENDING'
...
+-- name: FinishTerminalCommandBatches :execrows
+WITH candidates AS MATERIALIZED (
+    SELECT batch.id
+    FROM command_batch_log AS batch
+    WHERE batch.status IN ('PENDING', 'PROCESSING')
+      AND EXISTS (
+          SELECT 1
+          FROM queue_message AS message
+          WHERE message.command_batch_log_uuid = batch.uuid
+      )
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > handle zero-message batches in the startup-only cleanup path

Confirmed, but not addressing because this is an explicit tradeoff. A hard crash in the create-before-enqueue gap can leave a history-only `PENDING` batch, but it creates no device work, does not block later commands, and does not affect HA activation. Closing empty batches safely would reintroduce startup-specific query behavior and also needs enqueue protection against a paused former process. The residual edge case is documented in the PR description.

server/internal/infrastructure/queue/service.go
```diff
@@ -58,19 +56,32 @@ func (d DatabaseMessageQueue) EnqueueMany(ctx context.Context, commandBatchLogUU
...
+		inserted, err := q.CreateQueueMessages(ctx, sqlc.CreateQueueMessagesParams{
+			CommandBatchLogUuid: commandBatchLogUUID,
+			CommandType:         commandType.String(),
+			DeviceIds:           deviceIDs,
+			Payloads:            payloads,
+		})
+		if err != nil {
+			return fleeterror.NewInternalErrorf("failed to enqueue messages: %v", err)
+		}
+		if inserted != int64(len(messages)) {
```
mcharles-square: is this check needed because of the join in the select used in the insert query? could this be validated outside of the query?

mcharles-square (approved):
Tradeoffs seems reasonable

server/internal/infrastructure/queue/service.go
ankitgoswami: yea, it's not really useful because both are built from the same messages slice, I'll remove the check

PR #883, opened 2026-08-05
Tag Proto Fleet PR-eligible E2E scenarios

edgars-avotins:
Reviewable diff: +0/-0 across 0 files (excludes generated, test, and story files).

## Summary
This PR marks the Proto Fleet Playwright scenarios we want included in the default pull-request subset by appending `@pr` to their test titles. It does not change product behavior or test logic; it only labels a representative slice of the existing suite so we can wire PR-vs-nightly selection around explicit tags and see how that shape behaves in CI.

## How it works
Each selected scenario keeps its existing flow and assertions, but its title now includes `@pr`. The tagged set covers 61 of the 130 non-setup, non-visual Proto Fleet scenarios, which keeps broad feature coverage while trimming the default PR set down to roughly the 1/3-to-1/2 range we discussed.

## Diagrams
```mermaid
flowchart TD
  A["Proto Fleet spec file"] --> B["Existing scenarios"]
  B --> C["Selected representative scenarios get @pr tag"]
  C --> D["Future PR selection can target tagged scenarios"]
  B --> E["Untagged scenarios remain available for nightly/full runs"]
```

## Areas of the code involved
| --- | --- | --- |
| `client/e2eTests/protoFleet/spec/**` | Added `@pr` suffixes to selected test titles across the Proto Fleet E2E specs. | Review should focus on whether the chosen subset is representative and balanced, since no assertions or page-object behavior changed. |

- Tag individual scenarios instead of splitting files immediately, so we can shape PR coverage without duplicating hooks or setup flows first.
- Keep at least one tagged scenario in each meaningful area/spec where practical, instead of over-optimizing solely for runtime.
- Leave the alerts spec untagged for now because it still depends on its special environment gate and is not part of the normal PR path today.

## Testing & validation
- `./node_modules/.bin/eslint` on the touched Proto Fleet spec files
- Push hook `client-typecheck` passed when the branch was pushed
- No Playwright rerun, because this change only updates test titles and does not alter test behavior


mcharles-square (commented):
I'm not convinced that this approach of adding `@pr` to the test name is the best approach for tagging fast tests. Have you considered using the native test annotations? ie `{ tag: "smoke" }` (using smoke here instead of pr since `smoke` or `fast` is usually how these kinds of tests get tagged)

.github/workflows/protofleet-e2e-tests.yml
```diff
@@ -20,16 +36,42 @@ jobs:
...
 
+          if [ -z "$RUN_SCOPE" ]; then
+            RUN_SCOPE="all"
+          fi
+
           # Setup specs (NN-*.spec.ts) run automatically as Playwright project
           # dependencies in every shard. onboardingVisual.spec.ts runs in its
           # own fresh-env job, so only the remaining target specs are sharded.
-          SPEC_FILES=$( (find . -name "*.spec.ts" | sed 's|^\./||' | sort | grep -Ev '(^[0-9]{2}-.*\.spec\.ts$|^onboardingVisual\.spec\.ts$)' || true) | jq -R . | jq -cs .)
+          ALL_TARGET_SPECS=$(find . -name "*.spec.ts" | sed 's|^\./||' | sort | grep -Ev '(^[0-9]{2}-.*\.spec\.ts$|^onboardingVisual\.spec\.ts$)' || true)
```
mcharles-square, line 51: Can this exclusion also be done via tags and different scripts? is is necessary to parse out spec files here?

.github/workflows/protofleet-e2e-tests.yml
```diff
@@ -20,16 +36,42 @@ jobs:
...
+
+          if [ "$RUN_SCOPE" = "smoke" ]; then
+            SPEC_FILES=$(printf '%s\n' "$ALL_TARGET_SPECS" | while IFS= read -r spec; do
+              if [ -n "$spec" ] && grep -q '@smoke' "$spec"; then
+                printf '%s\n' "$spec"
+              fi
+            done | jq -R . | jq -cs .)
+            TEST_GREP="@smoke"
```
mcharles-square: `TEST_GREP` feels like the wrong name here. If this is just the target tag/tag filter probably just call it that

mcharles-square (approved):

PR #885, opened 2026-08-05
feat(deployment): route Fleet VIP by active health

ankitgoswami:
Reviewable diff: +437/-49 across 15 files (excludes generated, test, and story files).

## Summary

HA deployments now expose Fleet through one stable LAN address that follows the active Fleet process. Keepalived advertises the address only when Fleet holds its database lease and reports active health. Fleet stops lease renewal and exits if the routing layer loses the address or stops reporting liveness, so another node can take over.

## How it works

Patroni elects the PostgreSQL writer, and the existing Fleet HA coordinator grants one Fleet process an application lease on that writer. Only that process opens its command gate and returns success from `/health/active`.

`ha-a` and `ha-b` are the two Fleet hosts; this deployment profile also runs PostgreSQL and Patroni on them. A small HA Compose overlay starts only `fleet-api` and `fleet-client`, removes Fleet's dependency on the bundled standalone database, and supplies the Patroni-backed database and etcd settings. The API container receives only the service CA, read-only etcd observer password, and endpoint heartbeat directory.

Keepalived runs beside Fleet on both hosts. Once per second, its local health script calls `/health/active` through the client-facing HTTPS nginx proxy. Two consecutive failed checks withdraw the address; a successful response makes that host advertise the shared virtual IP through unicast VRRP and refresh an ephemeral heartbeat under `/run/proto-fleet-ha`. The passive host remains in backup and does not advertise the address. A systemd drop-in restarts keepalived after failure and removes a leftover VIP whenever the daemon exits.

Fleet independently checks that the virtual IP is assigned to a local interface and that the heartbeat timestamp is neither stale nor in the future. It allows five seconds for the initial VRRP election. After endpoint ownership has been observed once, three consecutive unhealthy samples fail closed: Fleet closes admission, cancels its lease coordinator before job cleanup, and exits through the existing fatal HA path. A healthy sample resets that short failure count.

On `ha-a` and `ha-b`, deployment preflight requires the VIP to be a usable address on the configured interface, rejects gateway routes, and uses ARP duplicate-address detection before startup. The etcd-only `ha-c` witness still validates its own address, peer routes, secrets, ports, and firewall, but it can live outside the VIP's L2 network. The firewall accepts VRRP only on the configured interface, from a Fleet-host address, to the local host. It also keeps Fleet's port 4000 on loopback; clients reach the existing HTTPS `fleet-client` proxy through the VIP. This narrows the trusted-LAN boundary but does not authenticate raw-packet peers.

```mermaid
flowchart LR
    P["Patroni and etcd"] --> C["Fleet HA coordinator"]
    C --> G["Active command gate"]
    G --> H["/health/active returns 200"]
    H --> T["HTTPS fleet-client proxy"]
    T --> K["Local keepalived health check"]
    K --> V["Shared Fleet VIP"]
    K --> B["Heartbeat in /run/proto-fleet-ha"]
    V --> E["Fleet endpoint monitor"]
    B --> E
    E --> X["Cancel lease and exit on loss"]
    N["LAN clients"] --> V
    V --> T
    T --> A["Fleet API on loopback"]
```

```mermaid
sequenceDiagram
    participant F as Fleet
    participant K as keepalived
    participant P as HTTPS proxy
    participant N as LAN clients
    participant L as Fleet lease
    F->>L: Acquire active lease
    F->>F: Open command gate
    K->>P: GET /api-proxy/health/active locally
    P->>F: Proxy active-health request
    F-->>P: 200 active
    P-->>K: 200 active through proxy
    K->>N: Advertise shared VIP
    K->>F: Refresh local heartbeat
    F->>F: Verify VIP and heartbeat
    K--xF: VIP or heartbeat remains unhealthy
    F->>F: Close command gate
    F->>L: Stop lease renewal
    F->>F: Abort active jobs and exit
```

## Areas of the code involved

|---|---|---|
| `deployment-files/ha/keepalived.conf.tmpl`, systemd drop-in, and health script | Define unicast VRRP, proxy-aware health tracking, restart, and VIP cleanup | Verify only a reachable active Fleet host can advertise the VIP and abnormal exit cannot strand it |
| `deployment-files/ha/firewall.nft.tmpl` and `node.env.example` | Add the shared VIP, interface, VRRP peer boundary, and loopback-only Fleet API rule | Verify the assumed same-L2 topology and allowed traffic |
| `deployment-files/ha/fleet-compose.yaml` | Overlay the regular deployment with HA-only settings, no standalone database dependency, and two individual secret mounts | Verify standalone Compose stays unchanged and Fleet cannot read substrate credentials |
| `server/internal/ha/deployment` | Render host-specific keepalived config and extend clean-host preflight | Verify unsafe or unreachable VIP configurations fail before startup |
| `server/internal/ha` | Monitor local VIP ownership and heartbeat; cancel renewal before fatal cleanup | Verify startup timing and fail-closed ordering |
| `server/cmd/fleet-ha` | Expose `render-keepalived` to deployment operators | Verify the CLI stays a small local installation primitive |
| `deployment-files/ha/README.md` | Document installation order and Fleet settings | Verify the operator flow matches the runtime dependency cycle |

- The Fleet database lease remains the authority for active work. Keepalived only routes traffic, instead of becoming a second ownership system.
- Keepalived runs on the host instead of in a privileged third-party container. This reduces container privileges and image lifecycle work, but adds one host package.
- The heartbeat uses one fixed file under a bind-mounted `/run/proto-fleet-ha` directory. This avoids configuration drift and durable disk writes, at the cost of no custom path override.
- The VIP must be directly connected on each Fleet host's configured interface. The etcd-only witness need not share that L2 network.
- VRRP assumes a trusted L2 network. Interface, source, and destination firewall checks reduce accidental exposure but do not provide cryptographic peer authentication.
- Port 4000 remains loopback-only in HA and clients use the existing HTTPS proxy. Standalone deployments retain their current default listener.
- HA Fleet uses a small Compose overlay rather than changing the standalone deployment model. Operators must start the two Fleet services with both files.
- Fleet receives only the CA and etcd observer password as individual mounts. Patroni, database-administrator, server-key, and token-signing secrets remain outside the API container.
- A five-second startup allowance covers the VRRP election and remains below the ten-second Fleet lease. Once healthy, bounded consecutive-failure thresholds tolerate brief probe blips without masking sustained endpoint loss.

## Testing & validation

- Focused Go tests cover endpoint config, fresh/stale/future heartbeat checks, transient recovery, sustained endpoint loss, lease cancellation before blocking cleanup, keepalived rendering, subnet boundaries, local/remote VIP conflict rejection, and an off-L2 etcd witness.
- `just _lint-server`
- `deployment-files/ha/tests/test-profile.sh`
- Rendered the merged regular plus HA Compose model and verified Fleet has no standalone database dependency and only two HA secret files.
- Rendered the real template and passed `keepalived --config-test` in an Alpine container with the production health script mounted.
- Not covered here: a real two-host VRRP election, gratuitous ARP convergence, or a full failover on the target LAN. Those require deployment validation.

- For 15 minutes after installation and during one controlled failover, verify exactly one Fleet host owns the VIP and the VIP serves `/health/ready`.
- Search Fleet logs for `active Fleet endpoint is unavailable` and `HA Fleet runtime aborted`. Search keepalived logs for repeated `MASTER`, `BACKUP`, or `FAULT` transitions.
- Healthy signal: `/health/active` succeeds only on the VIP owner, the peer stays passive, and the VIP moves within the Fleet lease window after the active process stops.
- Mitigate if both hosts advertise the VIP, neither owns it for more than one lease window, or transitions flap. Stop keepalived, route clients to one node-specific address, and disable HA on the selected Fleet instance while the network configuration is corrected.
- Owner: the operator performing the HA rollout, with the service on-call for the validation window.

---


server/internal/ha/deployment/preflight.go
```diff
@@ -71,13 +73,40 @@ func preflight(ctx context.Context, envPath, firewallTemplatePath string, host h
...
+	}
 	listeners, err := host.runCommand(ctx, "ss", "-H", "-lnt")
 	if err != nil {
 		return NodeConfig{}, fmt.Errorf("HA preflight failed: inspect listening ports: %s", commandError(listeners, err))
 	}
 	ports := []int{2379, 2380}
 	if config.isDatabaseNode() {
-		ports = append(ports, 5432, 8008)
+		ports = append(ports, 4000, 5432, 8008)
 	}
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Add the client-facing ports to this check for `ha-a`/`ha-b` as well.

Addressed in `6db3b72d23`: Fleet-host preflight now rejects occupied TCP ports 80 and 443, with coverage for both conflicts while preserving witness behavior.

ankitgoswami:
🤖

> Use a configurable startup deadline with sufficient margin beyond the complete probe and VRRP election window. Keep heartbeat-staleness detection separate if rapid post-start failure detection is required, and test delayed elections.

Addressed differently in `6db3b72d23`: startup now has a fixed 10-second endpoint-ownership allowance, separate from the 5-second heartbeat-freshness limit. This gives VRRP election adequate margin without reintroducing operator-facing timer configuration, and the existing monitor test now covers an election completing after nine seconds.

> Include ports 80 and 443 in Fleet-host preflight validation and add coverage for both conflicts.

Addressed in `6db3b72d23`: Fleet-host preflight now rejects occupied TCP ports 80 and 443, with focused coverage for both conflicts.

ankitgoswami:
🤖

> HA nodes can start with incompatible signing and encryption keys.
>
> The VIP is not given a stable TLS identity.

Addressed differently in `27ffa0cacd`: the offline HA generator now creates one shared Fleet authentication/encryption environment for both Fleet hosts and issues each proxy a certificate for the VIP under the common HA CA. Fleet-host preflight requires the generated environment, validates its key formats without logging values, verifies the VIP SAN and CA trust, and checks the proxy certificate/key pair. The Compose overlay mounts the generated proxy identity, and the runbook loads the shared environment after the deployment `.env` and tells clients to trust the common CA.

ankitgoswami:
🤖

> HA health probe disables TLS verification

Addressed in `66adc40dcc`. The keepalived probe now validates the certificate for the configured VIP using the generated service CA, pins the connection to the local nginx listener with `--resolve`, and bypasses proxy environment variables with `--noproxy '*'`.

> HA construction bypasses runtime dependency validation

Addressed in `66adc40dcc`. HA runtime construction now rejects a missing runtime job group or critical health check before reading secrets, loading TLS, or opening etcd/database resources. Focused coverage verifies both failure paths.

server/cmd/fleet-ha/main.go
```diff
@@ -1,70 +1 @@
...
-		return usageError()
-	}
-}
-
-func usageError() error {
-	return errors.New("usage: fleet-ha <generate-secrets|preflight|bootstrap-etcd-auth> ...")
-}
+m�
눧��bu��g��ǫ���rg~W��������,j�jǺ�7an{��)ߊW���_�W�n�����jG�r�^v����n�)��X�z�����7]y�
yל�ע������ج�V�����,j��zn�)����^}����+��b��.�����i����^}����+��h�
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Restore the fleet-ha source files

Addressed in `5d1311a266`. All six corrupted source, test, and documentation blobs were restored from the validated local worktree and their Git blob hashes were verified byte-for-byte before the branch was updated. The targeted Go tests and `git diff --check` pass.

deployment-files/ha/README.md
```diff
@@ -96,15 +110,72 @@ rm /secure/proto-fleet-ha-secrets/etcd-root-password
...
+generated secrets or node identity:
+
+```bash
+./ha/fleet-ha compose \
+  --env-file .env \
+  --env-file /etc/proto-fleet/ha/fleet.env \
+  --env-file ha/node.env \
+  --file docker-compose.yaml \
+  --file ha/fleet-compose.yaml \
+  up -d --no-build fleet-api fleet-client
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Add an HA image-prep step, including the HTTPS nginx config, before this no-build startup or let this command build the two Fleet services.

Addressed in `66adc40dcc`. The clean-host flow now copies the HTTPS nginx template into the Docker build context and uses `up --build` for `fleet-api` and `fleet-client`, so it does not depend on missing or stale images.

ankitgoswami:
🤖

> Endpoint health accepts the VIP on any interface

Addressed in `76fc81a4ae`. Fleet now verifies that the VIP is assigned specifically to `FLEET_HA_ENDPOINT_INTERFACE`, which the HA Compose overlay derives from `HA_NETWORK_INTERFACE`. A VIP stranded on loopback or another interface no longer keeps endpoint health green. Focused tests cover both the configured and incorrect-interface cases.

deployment-files/ha/README.md
```diff
@@ -96,15 +110,74 @@ rm /secure/proto-fleet-ha-secrets/etcd-root-password
...
+
+```bash
+cp client/nginx.https.conf client/nginx.conf
+./ha/fleet-ha compose \
+  --env-file .env \
+  --env-file /etc/proto-fleet/ha/fleet.env \
+  --env-file ha/node.env \
+  --file docker-compose.yaml \
+  --file ha/fleet-compose.yaml \
+  up -d --build fleet-api fleet-client
```
rl-block, line 167: This HA path appears to start the production `fleet-api` image, whose `server/Dockerfile` runs Delve on `:40000`. Because the service uses host networking and the firewall does not drop that port, should we perhaps run `fleetd` directly in production and block 40000 as a backstop?

deployment-files/ha/keepalived-systemd.conf.tmpl
```diff
@@ -0,0 +1,5 @@
+[Service]
+Restart=on-failure
+RestartSec=2s
+# Remove a stranded VIP after keepalived exits; '-' tolerates an absent address.
+ExecStopPost=-/usr/sbin/ip address delete ${HA_VIRTUAL_IP}/32 dev ${HA_NETWORK_INTERFACE}
```
rl-block: Seems like the leading `-` makes systemd ignore every failure from `ip address delete`, not just an absent address. Should we perhaps use a small helper that treats only an absent VIP as success, so real cleanup failures do not leave a stale address?

deployment-files/ha/README.md
```diff
@@ -96,15 +110,74 @@ rm /secure/proto-fleet-ha-secrets/etcd-root-password
...
+sudo systemctl enable --now keepalived
+```
+
+Start Fleet on both Fleet hosts from the deployment directory. The deployment
+`.env` supplies the database and other deployment settings. The generated
+`fleet.env` is passed after it so the cluster-wide authentication and encryption
+values override any independently created values. `node.env` supplies the local
+HA identity. Use the HA wrapper so exported parent variables cannot override the
+generated secrets or node identity. Select the HTTPS proxy configuration before
+building and starting the Fleet images:
```
rl-block: Do we want this flow to set `DB_DSN` explicitly? HA validation requires a multi-host DSN, but `.env`, generated `fleet.env`, and `node.env` do not currently provide one, so following these steps appears to fail at startup. We could generate it or add a setup step here.

server/internal/ha/runtime.go
```diff
@@ -223,10 +230,48 @@ func (r *Runtime) waitWhileHealthy(parent, activeCtx context.Context) error {
 			if !r.healthCheck() {
 				return errCriticalRuntimeUnhealthy
 			}
+			if err := endpoint.check(time.Now()); err != nil {
```
rl-block, line 233: Endpoint health is first checked after this host acquires the lease, and `fleet-api` restarts automatically. With a persistent local endpoint fault, could this host restart and win the lease again during each startup grace period? Might be worth carrying a local quarantine or precheck across restarts so the healthy peer gets a chance.

rl-block (approved):

deployment-files/ha/README.md
ankitgoswami: 🤖

> run `fleetd` directly in production and block 40000 as a backstop
>
> could this host restart and win the lease again during each startup grace period?

Addressed in `742ba48fb0`. The HA overlay now waits 15 seconds before running `/app/fleetd` directly, bypassing Delve and giving the continuously passive peer several lease-acquisition attempts after a restart. The HA firewall also explicitly drops TCP 40000.

server/internal/ha/runtime.go
ankitgoswami: 🤖

> run `fleetd` directly in production and block 40000 as a backstop
>
> could this host restart and win the lease again during each startup grace period?

Addressed in `742ba48fb0`. The HA overlay now waits 15 seconds before running `/app/fleetd` directly, bypassing Delve and giving the continuously passive peer several lease-acquisition attempts after a restart. The HA firewall also explicitly drops TCP 40000.

deployment-files/ha/README.md
ankitgoswami: 🤖

> Do we want this flow to set `DB_DSN` explicitly?

Addressed in `742ba48fb0`. Secret generation now writes the complete multi-host `DB_DSN` into the shared mode-0600 `fleet.env` using the generated Fleet database password, both database hosts, writer selection, and the HA service CA. Preflight requires it, and the Compose wrapper rejects parent `DB_DSN` overrides.

deployment-files/ha/keepalived-systemd.conf.tmpl
ankitgoswami: 🤖

> the leading `-` makes systemd ignore every failure from `ip address delete`, not just an absent address

Addressed differently in `742ba48fb0`: `ExecStopPost` now uses an unignored `ip address flush to ...`. The flush succeeds when the VIP is already absent, while permission, interface, and other real cleanup failures remain visible to systemd.

PR #888, opened 2026-08-07
Guide HA installation on supported apt-based hosts

ankitgoswami:
Reviewable diff: +2415/-255 across 22 files (excludes generated, test, and story files).

## Summary

Adds a guided `fleet-ha install` flow for dedicated Debian-based HA hosts. Operators no longer create `node.env`, arrange secrets, or set file permissions by hand: `ha-a` prepares protected, role-specific host bundles, and each host installs with one local command. Supported targets are Debian 12/13, Ubuntu 22.04/24.04, 64-bit Raspberry Pi OS based on Debian 12/13, and compatible apt/systemd derivatives on amd64 or arm64 with 4096-byte pages.

**Stack:** **#888** -> [#914](https://github.com/block/proto-fleet/pull/914) -> [#890](https://github.com/block/proto-fleet/pull/890) -> [#891](https://github.com/block/proto-fleet/pull/891). This PR targets `main` and provides the installed paths, services, and local operator workflow used by the HA promotion and application-update PRs above it. Merged [#887](https://github.com/block/proto-fleet/pull/887) supplies local HA status and readiness checks. Host replacement, unattended installation, resume, rollback, SSH orchestration, and exact-artifact qualification remain outside this PR.

## How it works

The operator transfers the release archive and official checksum to `ha-a`, verifies the archive before extraction, and runs `fleet-ha install` through `ssh -t`. The wizard prompts for the `ha-b`, `ha-c`, and VIP addresses, derives `ha-a`'s LAN address and interface, validates the packaged runtime and dedicated host, shows the complete topology and planned changes, and requires the operator to type `INSTALL` before mutation.

The wizard creates three architecture-independent JSON host bundles and the public service CA certificate. Each bundle contains only the topology, release identity, and credentials required by its role; only `ha-a` receives the one-time etcd root password. Before mutation, peer installation validates the protected bundle's size, exact fields, role, local host address, release identity, and certificates. The temporary CA private key is discarded because certificate rotation and host replacement are not supported by this initial installer.

```mermaid
flowchart TD
  O["Operator verifies release and runs fleet-ha install on ha-a"] --> T["Enter peer IPs and VIP"]
  T --> V["Validate runtime, topology, network, and dedicated host"]
  V --> C["Operator types INSTALL"]
  C --> B["Generate three role-specific JSON bundles"]
  B --> A["Install ha-a from its bundle"]
  B --> PB["Copy ha-b bundle over SSH"]
  B --> PC["Copy ha-c bundle over SSH"]
  PB --> IB["Run guided local install on ha-b"]
  PC --> IC["Run guided local install on ha-c"]
```

Each host follows stable phases for validation, package setup, configuration, image preparation, service startup, peer waiting, and final readiness. Docker is installed from its official Debian or Ubuntu apt repository when absent. An unused compatible Docker Engine and Compose v2 installation can be reused, as can inactive and unconfigured keepalived. Existing containers, orphaned container-runtime data without a working Docker installation, custom Docker configuration, configured or active keepalived, previous Proto Fleet state, occupied HA ports, and a claimed VIP fail before the release is copied or runtime services are installed.

Release packaging builds architecture-specific Fleet API and client images once, stores them in the checksum-covered `images/fleet.tar.gz`, and pins Compose to the release version. The installer loads and verifies those exact tags; application startup uses `--no-build --pull never`, so target hosts do not rebuild Fleet from mutable registry inputs.

The installer places the release under `/opt/proto-fleet`, root-owned configuration under `/etc/proto-fleet/ha`, and data under `/var/lib/proto-fleet/ha`. Systemd persists firewall and startup ordering: networking and firewall precede Docker and etcd, Patroni starts on database hosts, Fleet starts after its dependencies, and keepalived advertises the VIP only when the local active endpoint is healthy.

```mermaid
sequenceDiagram
  participant O as Operator
  participant I as fleet-ha installer
  participant S as systemd
  participant E as etcd cluster
  O->>I: Confirm INSTALL
  I->>I: Install validation prerequisites and check host firewall
  I->>I: Install runtime packages and protected configuration
  I->>S: Enable and start proto-fleet-ha.service
  S->>E: Start local peer listener, then wait for quorum
  E-->>S: Local member joins a responding quorum
  S->>S: Start Patroni and Fleet on database hosts
  I->>I: Run final local readiness check
```

`ha-a` enables etcd authentication only after quorum. On restart, existing authentication must accept the generated root, Patroni, and Fleet credentials and expose the expected roles and DCS permissions; mismatches fail closed. Once the local etcd peer listener opens within one minute, the service remains enabled and waits without a fixed deadline while missing peers converge. Final startup still requires the local member to join a responding etcd quorum. Fatal local setup or service-start failures stop the incomplete local service.

## Areas of the code involved

| --- | --- | --- |
| `server/cmd/fleet-ha/main.go` | Defines the Kong-based CLI and guided `install [HOST_BUNDLE]` command | Review the operator surface and TTY boundary |
| `server/internal/ha/deployment/guided_install.go` | Implements prompts, JSON bundle generation and validation, role resolution, and bundle consumption | Review credential separation and pre-mutation confirmation |
| `.github/workflows/proto-fleet-artifact-build.yml` | Builds architecture-specific Fleet images and embeds them in the checksummed release bundle | Review the release integrity boundary and image tags |
| `server/internal/ha/deployment/install.go` and `preflight.go` | Detect apt/systemd platforms, validate runtime and L2 requirements, load packaged images, and install protected files | Review platform selection, image verification, conflict rejection, and privilege boundaries |
| `server/internal/ha/deployment/start.go` and `bootstrap.go` | Start role-specific services, bound local etcd startup, wait for quorum, and bootstrap or verify etcd authentication | Review startup ordering and fail-closed authentication |
| `deployment-files/ha/` | Adds persistent firewall/systemd ordering and the guided two-command-per-host runbook | Review reboot persistence and the documented operator flow |
| `*_test.go` and `deployment-files/ha/tests/` | Covers the guided flow, bundle boundary, supported platforms, conflicts, convergence, and profile contracts | Tests; review with the associated behavior |

- Host bundles are JSON documents rather than archives, structurally avoiding path traversal and symlink handling.
- The official release checksum covers prebuilt Fleet and database image archives; target hosts load release-tagged images instead of rebuilding them.
- Bundle transfer relies on SSH integrity plus protected-file, exact-field, release, host, and certificate validation; redundant per-bundle checksum sidecars are omitted.
- Only role-scoped bundles and the public CA leave the generation workspace. The private CA key is discarded instead of creating an unsupported recovery workflow.
- Service certificates use a ten-year lifetime; certificate rotation remains deferred for this fixed deployment.
- Existing package caches, images, volumes, and networks are allowed when a working, container-free Docker installation is present. Orphaned Docker or containerd state is rejected rather than trusted.
- The HA Fleet container drops inherited `NET_ADMIN` and retains only `NET_RAW` for nmap discovery.
- Database peers and the VIP must be directly connected on the configured interface because VRRP/ARP cannot move the VIP across routed topology.
- One apt/systemd path covers the documented targets. RPM systems, non-systemd hosts, 32-bit ARM, and automatic base-release overrides remain out of scope.
- Installation is local and one-shot. There is no SSH controller, enrollment service, resume engine, or rollback state machine.

## Testing & validation

- `go test ./cmd/fleet-ha ./internal/ha/deployment -count=1`
- Hermit-backed `just _lint-server`
- `git diff --check`
- Guided transcript, role-scoped bundle contents, protected modes, exact-field validation, release/host binding, and total-size bounds.
- Debian, Ubuntu, Raspberry Pi OS, Ubuntu-derivative, and generic apt-derivative detection, including unsupported architecture, missing prerequisites, and 16K-page rejection.
- Dedicated-host reuse, orphaned runtime and firewall conflict cases, minimal container capabilities, role-specific startup, etcd authentication bootstrap and restart, and slow or absent peer convergence.
- Real-host installation and failover remain part of exact-artifact qualification.


ankitgoswami:
🤖 Addressed the mutable image-build finding in `9aee9e458b`. The release workflow now builds architecture-specific Fleet API and client images once, stores them in the checksum-covered `images/fleet.tar.gz`, and verifies the expected release tags are present. `fleet-ha install` loads and inspects those images instead of rebuilding on the target; runtime startup already uses `--no-build --pull never`. The stacked update flow now loads the same verified archive as well.

The certificate-lifetime finding remains an accepted initial-profile trade-off: rotation and host replacement are explicitly unsupported, and recovering from key compromise requires rebuilding the fixed cluster credentials and trust root.

ankitgoswami:
🤖 The two current Codex Security Review findings describe accepted reliability tradeoffs for the first guided-installer release:

- **Partial etcd authentication bootstrap:** the finding is accurate that a failure after creating only part of the auth policy is not automatically recoverable. Making this retry-safe would require reconciliation of users, roles, grants, permissions, and generated passwords. That is effectively installer resume/repair machinery, which this PR deliberately does not support. The install runs on a dedicated new host before the HA control plane is considered ready, and the documented recovery for an interrupted partial install is to reimage the dedicated hosts and rerun the guided installation. We are keeping that bounded operational recovery instead of adding a second bootstrap state machine.

- **Certificate/key relationship validation:** host bundles are generated by this installer, decoded through a strict typed JSON schema, checked for the exact role-specific secret set, matched to the local node IP and release, protected by ownership/mode checks, and transferred over SSH. Full X.509 chain, SAN/EKU, key-pair, and JWT-pair validation would detect a corrupted generated bundle earlier, but it would duplicate validation already performed when etcd, Patroni, PostgreSQL, nginx, and Fleet load those identities. It also would not establish bundle provenance against deliberate substitution. For this dedicated-host flow, a malformed bundle failing during startup has the same documented remediation: reimage and rerun.

These are not being treated as false positives. We are accepting the narrower failure behavior to keep this PR focused on the golden installation path and avoid roughly a few hundred lines of recovery and cryptographic prevalidation code. If qualification shows either failure mode occurring in normal operation, that evidence would justify expanding the installer.

ankitgoswami:
🤖 > HA firewall service cannot restart while its table exists

Fixed in c5740ec42. The service now ensures the profile-owned table exists, then validates and applies one nftables transaction that deletes and recreates it. On restart, the existing filtering rules remain active until nft atomically commits the replacement. The Go preflight path uses the same transaction shape.

I kept validation focused rather than adding a privileged CI harness: the deployment profile and targeted Go tests pass, and the replacement transaction was applied twice successfully against nftables v1.1.3 inside an isolated network namespace on the qualification hardware.

server/internal/ha/deployment/start.go
```diff
@@ -0,0 +1,195 @@
...
+	for {
+		enabled, err := etcdAuthEnabled(ctx, config)
+		if err == nil {
+			return enabled, nil
+		}
+		select {
+		case <-ctx.Done():
+			return false, fmt.Errorf("stopped checking etcd authentication: %w", ctx.Err())
+		case <-time.After(2 * time.Second):
+		}
```
rl-block, line 174: I think we could distinguish retryable quorum/transport failures from permanent local TLS or configuration errors here. With `TimeoutStartSec=infinity`, a bad CA or local TLS setup can loop forever and look like peer convergence. Perhaps we could validate the local TLS setup once before this loop, or return non-transient errors, with a focused test.

server/internal/ha/deployment/guided_install.go
```diff
@@ -0,0 +1,576 @@
...
+			return preparedHostBundle{}, fmt.Errorf("host bundle rejected: missing secret %s", name)
+		}
+	}
+	if bundle.Metadata.Role == "ha-a" && len(bundle.EtcdRootPassword) == 0 {
+		return preparedHostBundle{}, errors.New("host bundle rejected: missing etcd root password")
+	}
+	if bundle.Metadata.Role != "ha-a" && len(bundle.EtcdRootPassword) != 0 {
+		return preparedHostBundle{}, errors.New("host bundle rejected: etcd root password is only valid for ha-a")
+	}
+	return bundle, nil
```
rl-block: Should we also reject secret keys outside the expected set for this role? We currently verify required entries, but an over-privileged or malformed bundle can still retain unrelated credentials. Maybe comparing `bundle.Secrets` with `copiedSecretFiles(...)` and adding an extra-key test per role would keep the exact-set contract enforceable.

server/internal/ha/deployment/guided_install.go
```diff
@@ -0,0 +1,576 @@
...
+}
+
+func decodeHostBundle(contents []byte) (preparedHostBundle, error) {
+	var bundle preparedHostBundle
+	decoder := json.NewDecoder(bytes.NewReader(contents))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&bundle); err != nil {
+		return preparedHostBundle{}, fmt.Errorf("host bundle rejected: invalid JSON: %w", err)
+	}
+	if err := validateBundleMetadata(bundle.Metadata); err != nil {
```
rl-block, line 388: Probably worth requiring EOF after this first decode? As written, a valid bundle followed by another JSON value or trailing garbage is accepted. Perhaps a second `Decode` that must return `io.EOF`.

server/internal/ha/deployment/install.go
```diff
@@ -0,0 +1,897 @@
...
+	if err := installValidationPrerequisites(ctx, config.isDatabaseNode(), deps); err != nil {
+		return err
+	}
+	if config.isDatabaseNode() {
+		if err := deps.verifyVIP(ctx, config); err != nil {
+			return err
+		}
+	}
+	if err := rejectIncompatibleNftablesInputChains(ctx, deps); err != nil {
+		return err
```
rl-block, line 130: Should we validate persisted nftables configuration here as well as the live ruleset? If `nftables.service` is inactive, the live rules can be empty while `/etc/nftables.conf` still contains an input chain that loads when the HA firewall starts. Perhaps we could evaluate that persisted config in isolation before `snapshotRelease` and cover the inactive-service case.

rl-block (approved):

server/internal/ha/deployment/start.go
ankitgoswami: 🤖 > I think we could distinguish retryable quorum/transport failures from permanent local TLS or configuration errors here.

Fixed in 3591848c1. Startup now completes a CA-verified TLS handshake with the local etcd client listener before entering the unbounded quorum/auth wait. An absent listener remains retryable for one minute, while an invalid CA or server certificate fails immediately.

server/internal/ha/deployment/guided_install.go
ankitgoswami: 🤖 > Should we also reject secret keys outside the expected set for this role?

Fixed in 3591848c1. Bundle decoding now requires the secret map to contain exactly the role-specific files returned by copiedSecretFiles, so a bundle with any extra credential is rejected.

server/internal/ha/deployment/guided_install.go
ankitgoswami: 🤖 > Probably worth requiring EOF after this first decode?

Fixed in 3591848c1. The decoder now performs a second decode and requires io.EOF, rejecting a second JSON document or trailing garbage.

server/internal/ha/deployment/install.go
ankitgoswami: 🤖 > Should we validate persisted nftables configuration here as well as the live ruleset?

Fixed in 3591848c1. Preflight now loads /etc/nftables.conf in an isolated network namespace, lists that resulting ruleset as JSON, and applies the existing input-chain validation without changing the host firewall.

server/internal/ha/deployment/install.go
```diff
@@ -0,0 +1,901 @@
...
+			{file: "ha/compose.yaml", prefix: "proto-fleet-timescaledb-ha:"},
+			{file: "docker-compose.yaml", prefix: "proto-fleet-api:"},
+			{file: "docker-compose.yaml", prefix: "proto-fleet-client:"},
+		}
+		for _, expected := range images {
+			image, err := composeImage(source, expected.file, expected.prefix, deps.readFile)
+			if err != nil {
+				return err
+			}
+			if output, err := deps.run(ctx, "sudo", "docker", "image", "inspect", image); err != nil {
```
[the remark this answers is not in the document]
ankitgoswami: will handle this in the updater PRs

PR #905, opened 2026-08-10
feat(alerts): scope user alert rules to subsets of miners

illegalprime:
Reviewable diff: +2195/-243 across 28 files (excludes generated, test, and story files).

## Summary

User-created alert rules (offline / hashrate / temperature) can now be scoped to any union of sites, buildings, racks, groups, and explicitly selected miners; unscoped rules stay org-wide, and existing rules are untouched. Scope membership is resolved at evaluation time — moved miners join/leave a scope immediately — via a new owner-privilege view. As a prerequisite, rule configs move from a Grafana rule annotation to a server-side `alert_rule_config` table: Grafana copies annotations onto every alert instance, so a large scope (up to 600 ids) replicated per firing device could push a notification batch past the webhook's 32 MiB body cap and drop the org's whole batch.

Design doc: `docs/plans/2026-08-06-scope-alert-rules-to-subsets-of-miners-tdd.md` (included in this PR).

## How it works

**Write path.** The Add/Edit Rule surface is now a full-screen two-pane view (details + condition on the left, live summary on the right) with an "Apply to" section of five picker rows reusing the Schedules/curtailment pickers. The client sends a `RuleScope` on `RuleConfig`; the handler validates placement ids as live and org-owned (`ScopeLookup`), device ids against the existing maintenance-window pattern, and caps (100 per placement dimension, 500 devices). The service upserts the config into `alert_rule_config` *before* writing the Grafana rule, then compiles the rule SQL.

**Evaluation.** Scoped rules gain one predicate: `device_id IN (SELECT device_id FROM fleet_device_placement WHERE org_id = … AND (site_id IN … OR building_id IN … OR rack_id IN … OR group_id IN … OR device_id IN …))`. `fleet_device_placement` (migration 000134) is an owner-privilege view over `device` / `device_set_membership` / `device_set_rack`, granted to `grafana_ro` in `run-fleet.sh`, so only *current* membership of *live* devices matches — emit-time `site_id` stamps can be up to 10 minutes stale after a move, and soft-deleted miners' retained samples would otherwise keep matching. Scoped rules also add a `site_id` label resolved from current placement. Unscoped rules compile byte-identical to before: zero label/fingerprint churn for existing rules.

**Reads.** `attachConfigs` overlays store rows onto the Grafana rule list (fail-closed, like routing). Rules created before migration 000135 fall back to their legacy `proto_fleet_config` annotation (scope-less, so it cannot bloat instances); their first successful update migrates the config into a store row and strips the annotation.

**Contract hardening on the wire:**
- Scope device ids are redacted unless the caller holds org-**wide** `miner:read` (`device_ids_redacted` flag); writes carrying the flag are rejected since the caller cannot know the redacted list.
- Org-wide writes must send `org_wide=true` explicitly; an update of a scoped rule that omits the scope message entirely is rejected as a stale pre-scope client, so an unrelated edit can never silently widen a rule to every miner.
- `config_out_of_sync` / `config_unknown` flags mirror the existing routing-unknown handling for interrupted saves and failed config reads.
- Orphan config rows from ambiguous create failures are reclaimed by a sweep (1 h slack protects in-flight creates).

## Diagrams

```mermaid
flowchart LR
    UI["AddRuleModal (two-pane, Apply-to pickers)"] --> H["alerts handler: validate ids, caps, redaction"]
    H --> S["alerts service"]
    S -->|"1. upsert config"| T[("alert_rule_config")]
    S -->|"2. compile + write rule SQL"| G["Grafana rule"]
    G -->|"eval: semijoin"| V[("fleet_device_placement view")]
    V --> D[("device / device_set tables")]
    G -->|"firing (site_id label)"| W["webhook → Deliverer (unchanged)"]
```

```mermaid
sequenceDiagram
    participant C as Client
    participant S as alerts service
    participant DB as alert_rule_config
    participant G as Grafana
    Note over S,G: read path for a pre-scope (legacy) rule
    C->>S: ListRules
    S->>G: list rules
    S->>DB: ListConfigs(rule uids)
    DB-->>S: no row
    S->>S: fall back to legacy annotation
    Note over S,G: first update migrates it
    C->>S: UpdateRule
    S->>DB: UpsertConfig (before PUT)
    S->>G: PUT rule (annotation stripped)
```

## Areas of the code involved

| Area | What changed | Why it matters for review |
|---|---|---|
| `proto/alerts/v1/alerts.proto` | `RuleScope` message, `scope` on `RuleConfig`, `config_out_of_sync` / `config_unknown` on `Rule` | Wire contract: redaction + explicit-org-wide semantics live here |
| `server/migrations/000134` | `fleet_device_placement` owner-privilege view | Placement resolution + grafana_ro access boundary; down refuses while scoped rules exist |
| `server/migrations/000135` | `alert_rule_config` table | Config persistence; down refuses while rows exist (pre-table server cannot read them) |
| `server/internal/domain/alerts/user_rules.go` | Scope predicate compilation, `site_id` label column | SQL injection contract (server-formatted ints, regex-validated device ids); unscoped output must stay byte-identical |
| `server/internal/domain/alerts/service.go` | Config store lifecycle, scope validation, legacy-annotation fallback/migration, sweep | Create/update ordering and failure-cleanup semantics |
| `server/internal/handlers/alerts/handler.go` | Scope redaction gate (org-wide `miner:read`), flag plumbing | Permission boundary; also tightens `ListAlerts` device-identity gate to org-wide |
| `server/internal/domain/stores/sqlstores/` | `alert_rule_config.go` store, `ScopeLookup` queries | Straightforward sqlc wrappers |
| `deployment-files/run-fleet.sh` | Grant + smoke check + wait for the new view | Provisioning aborts if the view is missing; ordering matters |
| `client/.../alerts/` | Two-pane `AddRuleModal`, `useAlertScope`, `useScopeSampleMiners`, scope column in `RulesSection`, scope labels | UI reuses Schedules/curtailment pickers; scope editable in edit mode (unlike delivery) |
| `client/.../Schedules/SiteSelectionModal.tsx` | Optional `onSaveSelection` / `initialAllSelected` | Shared component: verify Schedules behavior unchanged |
| `server/generated/`, `client/src/protoFleet/api/generated/` | regenerated | generated — skip |
| `docs/plans/2026-08-06-…-tdd.md` | design doc | Context; alternatives and risks recorded there |

- **Configs persist server-side (`alert_rule_config`) instead of the Grafana annotation** — chosen over keeping the annotation because Grafana replicates annotations onto every alert instance, risking the 32 MiB webhook cap; costs a table + lifecycle code and a legacy fallback path.
- **Placement resolved at eval time via a semijoin on an owner-privilege view** — chosen over filtering on the sample's emit-time `site_id` (up to 10 min stale after a move) or save-time id snapshots (go stale as miners move/pair); soft-deleted miners are excluded by the view.
- **`all_sites` is a live flag, and the miner picker's select-all round-trips as org-wide** — chosen over materializing today's id lists, so future sites are covered and >100-site orgs can save.
- **Explicit `org_wide` write marker; omitting scope on update is rejected** — chosen over absent-means-org-wide, which would let a stale pre-scope client silently widen a scoped rule.
- **Scope device ids gated on org-wide `miner:read` with a `device_ids_redacted` flag** — a plain org-scope check would leak device ids from sites narrowed out of the caller's grant.
- **Down migrations refuse to run while scoped rules / config rows exist** — chosen over silent rollback, which would strand rules erroring (view gone) or unreadable (table gone).
- **Narrow `RuleScope` message, not `MinerListFilter`/`DeviceSelector`** — the placement view can't evaluate predicate filters (model, firmware), so a rich message would promise scoping the SQL can't deliver.
- **Unscoped rules compile byte-identical** — protects every existing rule from fingerprint churn; toggling scope on a rule intentionally changes its alert identity (user-initiated re-fire).

## Testing & validation

- `just lint` clean (buf, eslint --max-warnings 0, golangci-lint ×2).
- Server (`DB_PASSWORD=fleet`, `-p 1`, re-run after rebasing onto current main): `internal/domain/alerts` (golden SQL scope tests incl. byte-identical unscoped assertion, validation caps/cross-org), `internal/domain/stores/sqlstores` (new `alert_rule_config` integration test, 130 s), `internal/handlers/alerts` (scope redaction / permission tests), `internal/handlers/alertmanagerwebhook` — all pass.
- Client: vitest 12 files / 68 tests pass (alerts feature + Schedules pickers).
- `just gen` is idempotent on the branch — no generated-code drift (the only diff was local grpcio-tools version noise, reverted).
- Not covered here: the updated Playwright E2E specs (`alerts.spec.ts`) are included but not executed — run against `just dev` as a follow-up per the E2E convention; manual scoped-fire verification (fake miner in/out of a scoped site) is also outstanding. Contract tests were skipped: the proto change is confined to the alerts API and regen confirmed no plugin SDK surface changed.


server/internal/domain/alerts/user_rules.go
```diff
@@ -211,6 +290,142 @@ func (s *Service) updateRuleSerialized(ctx context.Context, orgID int64, id stri
...
+	if raw == "" {
+		return nil
+	}
+	var cfg RuleConfig
+	if err := json.Unmarshal([]byte(raw), &cfg); err != nil {
+		slog.WarnContext(ctx, "alerts.rule_legacy_config_parse", "org_id", orgID, "rule_id", ruleUID, "error", err)
+		return nil
+	}
+	cfg.Scope = normalizeRuleScope(cfg.Scope)
+	if err := validateRuleConfig(cfg); err != nil {
```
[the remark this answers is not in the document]
illegalprime: 🤖 Fixed in ea0576350 — `legacyRuleConfig` now takes the rule's template and hides a mismatched annotation config exactly like the old inline parser (`cfg.Template() != tmpl` → nil, non-editable). Added a "template disagrees with rule" case to `TestListRulesIgnoresInvalidLegacyAnnotationConfig`.

server/internal/domain/alerts/user_rules.go
```diff
@@ -211,6 +290,151 @@ func (s *Service) updateRuleSerialized(ctx context.Context, orgID int64, id stri
...
+	live, err := s.grafana.GetAlertRule(probeCtx, id)
+	if err != nil {
+		slog.Warn("alerts.user_rule_update_config_probe", "org_id", orgID, "rule_id", id, "error", err)
+		return
+	}
+	if !ruleCarriesUpdate(*live, attempted) {
+		return
+	}
+	if uerr := s.configs.UpsertConfig(probeCtx, orgID, id, cfg); uerr != nil {
+		slog.Warn("alerts.user_rule_update_config_publish", "org_id", orgID, "rule_id", id, "error", uerr)
```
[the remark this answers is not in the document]
illegalprime: 🤖 Fixed in 736efa917 — `updateRuleSerialized` now stages a legacy rule's annotation config as its store row *before* the Grafana PUT, and aborts the update if that stage fails. Every later failure combination (committed-PUT + failed publish, failed reconcile/restore) now leaves the staged row behind, so `ListRules` serves the old config with `ConfigOutOfSync` set instead of rendering the rule config-less. This also restores the rollback guards' invariant that live scoped SQL always has a config row anchoring the 000135 down refusal (the related HIGH finding in the security review). Tests: `TestUpdateRuleAbortsWhenLegacyConfigStageFails`, `TestUpdateRuleKeepsStagedLegacyConfigThroughPublishFailure`.

server/internal/domain/alerts/user_rules.go
```diff
@@ -211,6 +306,151 @@ func (s *Service) updateRuleSerialized(ctx context.Context, orgID int64, id stri
 	return updated, nil
 }
 
+// reconcileFailedConfigWrite handles an UpsertConfig error after a committed
```
ankitgoswami: we already have markConfigOutOfSync which seems to protect against this case, can we simplify this down to single a idempotent upsert retry?

client/src/protoFleet/features/alerts/api/useScopeSampleMiners.ts
```diff
@@ -0,0 +1,119 @@
...
+  // the sample was merged across dimensions (union totals would double-count).
+  total: number | null;
+  loading: boolean;
+}
+
+// A one-shot (per open/scope change) sample of miners the scope currently
+// covers, for the create/edit preview pane. Placement dimensions map onto
+// MinerListFilter facets; facets AND together server-side, so each dimension
+// is fetched separately and merged to preserve the scope's union semantics.
+export function useScopeSampleMiners(enabled: boolean, scope: UseAlertScopeResult): ScopeSampleResult {
```
ankitgoswami, line 24: the TDD says this is out of scope for V1?

client/src/protoFleet/features/alerts/components/AddRuleModal.tsx
```diff
@@ -88,26 +112,66 @@ const describeDuration = (seconds: number): string => {
   return amount === "1" ? `1 ${singular}` : `${amount} ${unit}`;
 };
 
-// Phrased like the server's compiled rule summaries (compileTemplate), so the
-// preview reads the same as the saved rule's Condition column.
+// The preview pane's lead sentence, phrased like the Figma design's examples.
```
ankitgoswami: nit: this is a weird and useless comment hah

ankitgoswami (approved):
LGTM - I would split something like this into a 2 PR stack next time. Maybe the UI changes could have been reviewed independently.

illegalprime:
@ankitgoswami definitely. The UI change (switching modal styles) was only around 500 lines, but that would've made a good PR size.

client/src/protoFleet/features/alerts/api/useScopeSampleMiners.ts
illegalprime: this is more about listing miners in scope rather than evaluating the alerting rules against them.

PR #906, opened 2026-08-10
fix(collection): validate SetRackSlotPosition before writing

flesher:
Reviewable diff: +36/-1 across 1 files (excludes generated, test, and story files).

## Summary

`SetRackSlotPosition` — the single-slot placement RPC behind both the collection and deviceset handlers — accepted a slot write without checking the rack's grid dimensions, taking a rack lock, or confirming the device was a rack member. It would persist a position outside the grid and return success for a non-member, leaving a miner recorded in a cell the UI can never render and cannot be dragged off. This PR brings the single-slot path up to the same write contract the batch assignment path (`AssignDevicesToRack`) already enforces.

## How it works

The change adds three guards inside the RPC's existing transaction, after the collection is confirmed to be a rack and before the slot is written:

1. **Lock the rack row** (`LockRackPlacementForWrite`) up front, so the reads that follow and the eventual write all hold under the same lock. This is the same rack-first lock the resize and batch paths take, so a concurrent resize can no longer shrink the grid out from under a position that just passed its bounds check.
2. **Bounds check** — read the rack's real dimensions and reject any row/column that falls outside the grid with `InvalidArgument`, using the same wording as the batch path. A rack with no dimension row is treated as a broken invariant and fails loudly rather than writing blind.
3. **Membership check** — confirm the device is actually a member before writing. The underlying store statement is an `INSERT … SELECT` over the membership table that silently writes zero rows for a non-member, so without this the RPC returned `200 OK` for a placement that never happened.

Only when all three pass does the slot write proceed. Both handler surfaces call this one service method, so the fix covers them without duplication.

```mermaid
flowchart TD
    A["SetRackSlotPosition RPC"] --> B["Begin transaction"]
    B --> C{"Collection is a rack?"}
    C -->|"no"| E1["Reject: InvalidArgument"]
    C -->|"yes"| D["Lock rack row FOR UPDATE"]
    D --> F["Read rack dimensions"]
    F --> G{"Position within grid?"}
    G -->|"no"| E2["Reject: out of bounds"]
    G -->|"yes"| H{"Device is a member?"}
    H -->|"no"| E3["Reject: not a member"]
    H -->|"yes"| I["Write slot position"]
    I --> J["Commit + log activity"]
```

## Areas of the code involved

|---|---|---|
| `server/internal/domain/collection/service.go` | Added lock, bounds, and membership guards inside `SetRackSlotPosition`'s transaction | The behavioral change — verify the guards run under the lock and before the write |
| `server/internal/domain/collection/service_test.go` | Added unit tests: out-of-bounds rejected, non-member rejected, in-bounds member succeeds | Confirms each guard fires and the happy path still writes |

- **Lock the rack row rather than rely on a bounds check alone** — a check without a lock is still racy against a concurrent resize; the lock matches the ordering the batch and resize paths already use.
- **Reject a missing dimension row as `Internal`** — a rack always has a dimension row, so `nil` signals data corruption and is failed loudly instead of silently skipping the bounds check.
- **Fix in the shared service method, not the two handlers** — both handlers delegate here, so a single change covers both surfaces with no duplicated validation.

## Testing & validation

Added unit tests covering out-of-bounds rejection, non-member rejection, and the in-bounds member happy path; full collection package tests pass, `gofmt` and `go vet` clean. `ClearRackSlotPosition` is intentionally out of scope: it only deletes, so it has no bounds risk and clearing a non-member is a harmless no-op.

### DIFF
--- a/server/internal/domain/collection/service.go
+++ b/server/internal/domain/collection/service.go
@@ -1990,11 +1990,14 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
 	}
 
 	result, err := s.transactor.RunInTxWithResult(ctx, func(ctx context.Context) (any, error) {
-		coll, err := s.collectionStore.GetCollection(ctx, info.OrganizationID, req.CollectionId)
+		// Gate on type before locking. Type is immutable, so this pre-lock read
+		// can't go stale; authoritative label + site for the activity event are
+		// re-read under the lock below.
+		collType, err := s.collectionStore.GetCollectionType(ctx, info.OrganizationID, req.CollectionId)
 		if err != nil {
 			return nil, err
 		}
-		if coll.Type != pb.CollectionType_COLLECTION_TYPE_RACK {
+		if collType != pb.CollectionType_COLLECTION_TYPE_RACK {
 			return nil, fleeterror.NewInvalidArgumentError("slot positions can only be set on rack collections")
 		}
 
@@ -2007,7 +2010,10 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
 		}
 
 		// Reject out-of-bounds cells. A RACK always has a device_set_rack row, so
-		// nil is a broken invariant.
+		// nil is a broken invariant. Guard both bounds so a negative coordinate
+		// (which bypasses the proto interceptor on a direct call) returns
+		// InvalidArgument rather than tripping the SQL CHECK as a 500, matching
+		// the batch and SaveRack paths.
 		rackInfo, err := s.collectionStore.GetRackInfo(ctx, req.CollectionId, info.OrganizationID)
 		if err != nil {
 			return nil, err
@@ -2015,10 +2021,10 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
 		if rackInfo == nil {
 			return nil, fleeterror.NewInternalErrorf("rack %d has no rack extension row", req.CollectionId)
 		}
-		if req.Position.Row >= rackInfo.Rows {
+		if req.Position.Row < 0 || req.Position.Row >= rackInfo.Rows {
 			return nil, fleeterror.NewInvalidArgumentErrorf("slot row %d is out of bounds (rack has %d rows)", req.Position.Row, rackInfo.Rows)
 		}
-		if req.Position.Column >= rackInfo.Columns {
+		if req.Position.Column < 0 || req.Position.Column >= rackInfo.Columns {
 			return nil, fleeterror.NewInvalidArgumentErrorf("slot column %d is out of bounds (rack has %d columns)", req.Position.Column, rackInfo.Columns)
 		}
 
@@ -2035,6 +2041,14 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
 		}
 
 		if err := s.collectionStore.SetRackSlotPosition(ctx, req.CollectionId, req.DeviceIdentifier, req.Position.Row, req.Position.Column, info.OrganizationID); err != nil {
+			return nil, err
+		}
+
+		// Read the collection under the lock so the activity event's label + site
+		// reflect the state we wrote to, not a pre-lock snapshot a concurrent
+		// rename/move could have staled.
+		coll, err := s.collectionStore.GetCollection(ctx, info.OrganizationID, req.CollectionId)
+		if err != nil {
 			return nil, err
 		}
 
--- a/server/internal/domain/collection/service.go
+++ b/server/internal/domain/collection/service.go
@@ -1998,7 +1998,42 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
 			return nil, fleeterror.NewInvalidArgumentError("slot positions can only be set on rack collections")
 		}
 
-		// Device membership is enforced by the store query joining on device_set_membership.
+		// Lock the rack row FOR UPDATE before reading grid + membership so the
+		// checks hold for the write; without it a concurrent resize could shrink
+		// the grid out from under an in-bounds position. Same rack-first lock the
+		// batch path takes.
+		if _, err := s.collectionStore.LockRackPlacementForWrite(ctx, req.CollectionId, info.OrganizationID); err != nil {
+			return nil, err
+		}
+
+		// Reject out-of-bounds cells. A RACK always has a device_set_rack row, so
+		// nil is a broken invariant.
+		rackInfo, err := s.collectionStore.GetRackInfo(ctx, req.CollectionId, info.OrganizationID)
+		if err != nil {
+			return nil, err
+		}
+		if rackInfo == nil {
+			return nil, fleeterror.NewInternalErrorf("rack %d has no rack extension row", req.CollectionId)
+		}
+		if req.Position.Row >= rackInfo.Rows {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot row %d is out of bounds (rack has %d rows)", req.Position.Row, rackInfo.Rows)
+		}
+		if req.Position.Column >= rackInfo.Columns {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot column %d is out of bounds (rack has %d columns)", req.Position.Column, rackInfo.Columns)
+		}
+
+		// Confirm membership: the store's INSERT ... SELECT silently writes zero
+		// rows for a non-member and returns no error, so a bare call would report
+		// a placement that never landed.
+		members, err := s.collectionStore.GetDeviceSiteIDsByMembership(ctx, req.CollectionId, info.OrganizationID)
+		if err != nil {
+			return nil, err
+		}
+		if _, ok := members[req.DeviceIdentifier]; !ok {
+			return nil, fleeterror.NewInvalidArgumentErrorf(
+				"device %q is not a member of rack %d, so its slot cannot be set", req.DeviceIdentifier, req.CollectionId)
+		}
+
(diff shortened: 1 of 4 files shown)


server/internal/domain/collection/service.go
```diff
@@ -1998,7 +1998,42 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
...
+		}
+		if rackInfo == nil {
+			return nil, fleeterror.NewInternalErrorf("rack %d has no rack extension row", req.CollectionId)
+		}
+		if req.Position.Row >= rackInfo.Rows {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot row %d is out of bounds (rack has %d rows)", req.Position.Row, rackInfo.Rows)
+		}
+		if req.Position.Column >= rackInfo.Columns {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot column %d is out of bounds (rack has %d columns)", req.Position.Column, rackInfo.Columns)
+		}
```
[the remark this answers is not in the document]
flesher: 🤖 Fixed in e49fd411c — the bounds checks now guard the lower bound too (`row < 0 || row >= rackInfo.Rows`, same for column), so a negative coordinate returns InvalidArgument instead of tripping the SQL CHECK. Matches the `SaveRack` path. Added `TestService_SetRackSlotPosition_RejectsNegativePosition`.

server/internal/domain/collection/service.go
```diff
@@ -1998,7 +1998,42 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
...
+			return nil, err
+		}
+		if rackInfo == nil {
+			return nil, fleeterror.NewInternalErrorf("rack %d has no rack extension row", req.CollectionId)
+		}
+		if req.Position.Row >= rackInfo.Rows {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot row %d is out of bounds (rack has %d rows)", req.Position.Row, rackInfo.Rows)
+		}
+		if req.Position.Column >= rackInfo.Columns {
+			return nil, fleeterror.NewInvalidArgumentErrorf("slot column %d is out of bounds (rack has %d columns)", req.Position.Column, rackInfo.Columns)
```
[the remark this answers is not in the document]
flesher: 🤖 Fixed in e49fd411c. Added lower-bound checks (`row < 0 || row >= rows`, `column < 0 || column >= columns`) so a direct call with a negative coordinate returns InvalidArgument rather than hitting `rack_slot.valid_position` and surfacing as a 500 — same contract as the batch and SaveRack paths.

flesher:
🤖 **Re: Codex Security Review — collection metadata read before the rack lock (MEDIUM)**

Fixed in e49fd411c. The pre-lock `GetCollection` is gone. The transaction now:
1. reads collection **type** before locking (type is immutable, so a stale read here is harmless) — used only to reject non-rack collections with a friendly InvalidArgument;
2. takes `LockRackPlacementForWrite`;
3. runs the bounds + membership checks and the write;
4. re-reads `GetCollection` **under the lock** to source the activity event's label + site.

So the label/site now reflect the state we actually wrote to, closing the stale-attribution window a concurrent rename/move could otherwise open.

server/internal/domain/collection/service.go
```diff
@@ -1990,19 +1990,68 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
...
 		}
 
-		// Device membership is enforced by the store query joining on device_set_membership.
+		// Lock the rack row FOR UPDATE before reading grid + membership so the
+		// checks hold for the write; without it a concurrent resize could shrink
+		// the grid out from under an in-bounds position. Same rack-first lock the
+		// batch path takes.
+		if _, err := s.collectionStore.LockRackPlacementForWrite(ctx, req.CollectionId, info.OrganizationID); err != nil {
+			return nil, err
+		}
```
rl-block, line 2010: Maybe worth preserving the underlying PostgreSQL error on this new lock path. `LockRackPlacementForWrite` currently wraps non-NotFound errors with `%v`, so `WithTransaction` cannot detect and retry `40001`/`40P01`. Perhaps changing the wrappers exercised here to `%w` and adding a retry test would keep the transaction retry contract intact.

server/internal/domain/collection/service_test.go
```diff
@@ -369,6 +369,122 @@ func TestService_SetRackSlotPosition_RejectsGroupCollection(t *testing.T) {
...
+	mockStore.EXPECT().LockRackPlacementForWrite(gomock.Any(), testCollectionID, testOrgID).
+		Return(interfaces.RackPlacement{}, nil)
+	mockStore.EXPECT().GetRackInfo(gomock.Any(), testCollectionID, testOrgID).
+		Return(&pb.RackInfo{Rows: 2, Columns: 2}, nil)
+	mockStore.EXPECT().GetDeviceSiteIDsByMembership(gomock.Any(), testCollectionID, testOrgID).
+		Return(map[string]*int64{"device-1": nil}, nil)
+	mockStore.EXPECT().SetRackSlotPosition(gomock.Any(), testCollectionID, "device-1", int32(1), int32(1), testOrgID).
+		Return(nil)
+	mockStore.EXPECT().GetCollection(gomock.Any(), testOrgID, testCollectionID).
+		Return(&pb.DeviceCollection{Id: testCollectionID, Type: pb.CollectionType_COLLECTION_TYPE_RACK, Label: "R1"}, nil)
```
rl-block: Since this test is meant to protect the post-write metadata read, should we pin `SetRackSlotPosition` before `GetCollection` with `gomock.InOrder` and assert the activity event's `ScopeLabel`/`SiteID` using the existing activity helper? As written, the test would still pass if the read moved before the write or those fields stopped reaching the event.

server/internal/domain/collection/service.go
```diff
@@ -1990,19 +1990,68 @@ func (s *Service) SetRackSlotPosition(ctx context.Context, req *pb.SetRackSlotPo
...
+		// nil is a broken invariant. Guard both bounds so a negative coordinate
+		// (which bypasses the proto interceptor on a direct call) returns
+		// InvalidArgument rather than tripping the SQL CHECK as a 500, matching
+		// the batch and SaveRack paths.
+		rackInfo, err := s.collectionStore.GetRackInfo(ctx, req.CollectionId, info.OrganizationID)
+		if err != nil {
+			return nil, err
+		}
+		if rackInfo == nil {
+			return nil, fleeterror.NewInternalErrorf("rack %d has no rack extension row", req.CollectionId)
```
rl-block, line 2022: Worth distinguishing a missing rack extension from a missing rack here? The production lock query starts from `device_set_rack`, so a missing extension becomes `NotFound` before this `rackInfo == nil` / `Internal` guard can run. Perhaps locking the parent row independently (or returning a distinct state) plus a store-level test would make the stated invariant behavior reachable.

rl-block (approved):

server/internal/domain/collection/service_test.go
flesher: 🤖 Done in c20b5c93c. The happy-path test now uses `gomock.InOrder` to pin `SetRackSlotPosition` before `GetCollection`, and asserts the emitted event's `ScopeLabel` ("R1") and `SiteID` (77) via the activity mock — so it fails if the read moves ahead of the write or those fields stop reaching the event.

server/internal/domain/collection/service.go
flesher: 🤖 Valid — with `%v` the underlying `pgconn.PgError` is flattened out of the chain, so `IsRetryablePostgresError` (which relies on `errors.As`) can't see a `40001`/`40P01` and `WithTransaction` won't retry.

But this is a pre-existing, store-wide convention rather than something this path introduces: 44 methods in `collection.go` wrap with `%v` vs 12 with `%w`, and the old lock-free `SetRackSlotPosition` write could already deadlock and be swallowed the same way. A one-off `%w` on just `LockRackPlacementForWrite` would leave the reads/write in the same tx still swallowing, so a partial fix here would be inconsistent.

I'd rather do it right as a focused follow-up: sweep the store's `%v`→`%w` wrappers and add retry-contract tests. Filing that separately — happy to prioritize it into this PR instead if you'd prefer.

server/internal/domain/collection/service.go
flesher: 🤖 Correct that in production the `rackInfo == nil` branch is unreachable: `LockRackPlacementForWrite` reads from `device_set_rack`, so a rack with no extension row returns `NotFound` at the lock before this guard runs.

I kept it deliberately as defensive parity with the batch path (`AssignDevicesToRack`, service.go:1669-1671), which has the identical lock → `GetRackInfo` → nil-Internal shape — dropping it only here would make the two rack-write paths diverge.

The deeper suggestion (lock the parent `device_set` row independently so the missing-extension state becomes reachable/distinct) is a broader change to the locking model I'd keep out of this bugfix; if that distinct-state behavior is worth making real, it's better as its own change across both paths.

server/internal/domain/collection/service.go
flesher: 🤖 Went ahead and did the scoped version in ff883023a. Since the codebase already mixes `%v`/`%w`, I converted just the six store methods this transaction exercises — `GetCollectionType`, `GetCollection`, `GetRackInfo`, `LockRackPlacementForWrite`, `GetDeviceSiteIDsByMembership`, `SetRackSlotPosition` — from `%v` to `%w`, leaving the other 40 sites for a separate sweep.

Added `TestLockRackPlacementForWrite_PreservesRetryablePgError`: it injects a `40001` via a stub querier (through the `WithTxQueries` context seam, no DB needed) and asserts `db.IsRetryablePostgresError` still sees it through the store's wrap — so `WithTransaction` retries instead of surfacing a dead 500. Would fail if the wrap regressed to `%v`.

PR #913, opened 2026-08-12
feat(alerts): roll active alerts up per rule with a per-miner drill-in

illegalprime:
Reviewable diff: +1228/-441 across 33 files (excludes generated, test, and story files).

## Summary

During a fleet-wide outage the active-alerts card rendered one row per affected miner, so the view that exists to show an outage was least usable exactly when it mattered, and the Slack digest sent one line per miner. Both surfaces now show one row per firing **rule** with its blast radius ("Miner Offline — 5,000 miners"), and the affected miners load on demand when an operator clicks into a row. A new `ListActiveAlertGroups` RPC serves the rollup and `ListAlerts` gains a paged per-rule drill-in.

## How it works

Nothing changes about how alerts arrive. Alertmanager still posts to the webhook, rows still land in `notification_history`, and a trigger still projects the latest state per alert into `notification_active`. This PR adds two new **reads** of `notification_active` and points the UI and the digest at them.

**The rollup (dashboard card).** The card polls `ListActiveAlertGroups`. The handler resolves the caller's org, then asks a per-org cache for the rollup rows. On a miss the cache runs one aggregate that groups the org's firing rows by `(alert_name, rule_group)` and returns, per group, the alert and distinct-miner counts, the earliest start across the group, and the newest instance's severity/template/summary. Concurrent viewers of the same org collapse onto a single query through a singleflight slot; the result is held for a TTL shorter than the dashboard's poll interval, so a lone viewer never waits an extra tick to see a change. The cache stores the raw store rows rather than the finished response, because whether a summary is visible depends on the caller's `miner:read` grant — redaction is applied per request, after the cache.

**The drill-in (modal).** Clicking a row opens a modal that pages `ListAlerts` with `active_only`, the rule's stored `alert_name`, and its `rule_group`. That branch returns one row per affected miner and pages forward on `alert_key` — the immutable per-alert identity in `notification_active` — rather than on the row id. This matters because Alertmanager re-asserts a firing alert repeatedly: each re-assert rewrites the row's `history_id`, which would lift not-yet-read rows above a row-id cursor and silently drop them from every later page. The modal captures the clicked group by value, so an alert resolving mid-read does not close the modal and discard the pages already loaded.

**Which rows are rows.** A rule that fires on something other than a miner (the operator-only self-monitoring group, host-scoped system rules, MQTT curtailment sources) has no blast radius to rank, and would otherwise render as "0 miners". The client previously classified these with two hardcoded allowlists of rule groups and template names. It now simply branches on the server's `device_count === 0` and renders those as callouts above the table, so adding a new device-less rule scope needs no client change.

**Delivery.** `renderSlack` and `renderWebhook` group through one shared `alertGroup` rollup, so an outage becomes a handful of Slack sections with miner counts instead of thousands of lines. A group names up to three miners inline and then "+N more"; a single-miner group is named inline rather than rendered as "1 miner" plus a list of one. The webhook payload gains `firing_groups` / `resolved_groups` next to the existing per-instance arrays, so existing consumers are unaffected.

**Schema.** The rollup and the drill-in each need an index that `notification_active` did not have. Because those indexes store webhook-supplied label text raw, a bounding migration runs first — see the decisions below.

## Diagrams

Read path, from ingest to the two new queries:

```mermaid
flowchart LR
    AM["Alertmanager webhook"] --> NH["notification_history"]
    NH --> TRG["notification_active_sync trigger"]
    TRG --> NA["notification_active"]
    NA --> AGG["GROUP BY alert_name, rule_group"]
    NA --> KS["keyset scan on alert_key"]
    AGG --> CACHE["activeGroupsCache<br/>singleflight, one query per org per TTL"]
    CACHE --> RPC1["ListActiveAlertGroups"]
    KS --> RPC2["ListAlerts (active_only + alert_name)"]
    RPC1 --> CARD["ActiveAlertsCard"]
    CARD --> MODAL["AlertInstancesModal"]
    MODAL --> RPC2
    NA --> DIG["Slack / webhook digest"]
```

`ListAlerts` now has three branches, each naming its own cursor so the response shape cannot drift from the branch that produced it:

```mermaid
flowchart TD
    REQ["ListAlerts request"] --> Q1{"active_only set?"}
    Q1 -- no --> HIST["History feed<br/>keyset on row id, descending"]
    Q1 -- yes --> Q2{"alert_name or rule_group set?"}
    Q2 -- no --> ALL["Whole firing set<br/>capped, unpaged, no cursor"]
    Q2 -- yes --> DRILL["One rule's instances<br/>keyset on alert_key, ascending"]
    HIST --> CUR["next_cursor = last row id"]
    ALL --> NOCUR["next_cursor empty; has_more means the cap hid rows"]
    DRILL --> CUR2["next_cursor = last alert_key"]
```

Why the drill-in cannot page on `history_id`:

```mermaid
sequenceDiagram
    participant C as Client
    participant DB as notification_active
    participant AM as Alertmanager
    C->>DB: page 1 (rows 1-50)
    DB-->>C: rows + cursor
    AM->>DB: re-assert group (every history_id rewritten)
    C->>DB: page 2 after cursor
    Note over DB: on a history_id cursor the unread rows<br/>now sort above it and are never returned
    Note over DB: alert_key is immutable, so page 2 resumes correctly
```

## Areas of the code involved

| --- | --- | --- |
| `proto/alerts/v1/alerts.proto` | New `ListActiveAlertGroups` RPC + `ActiveAlertGroup`; `ListAlerts` gains `alert_name`, optional `rule_group`, `next_cursor`; `before_id` gains a length bound | The public contract. Check the `optional rule_group` presence semantics and that `next_cursor` is additive |
| `server/internal/handlers/alerts/handler.go` | `ListAlerts` restructured into three cursor-naming branches; new `ListActiveAlertGroups`; `visibleSummary` extracted | **Main review target.** Confirm the unfiltered `active_only` contract is unchanged for old clients |
| `server/internal/handlers/alerts/active_groups_cache.go` | **New.** Per-org TTL cache with singleflight and a detached fetch context | **Main review target.** Concurrency and the store-rows-not-responses choice |
| `server/sqlc/queries/notification_history.sql` | New `ListActiveNotificationsByAlert` and `ListActiveNotificationGroups` | The aggregate shape and the keyset predicate |
| `server/internal/domain/stores/sqlstores/notification_history.go` | Store methods for both queries | Thin mapping layer |
| `server/internal/domain/notificationhistory/notificationhistory.go` | `AlertKey` on `StoredNotification`; new `ActiveAlertGroup`, `ActiveAlertFilter`; two `Lister` methods | Interface surface |
| `server/migrations/000136` | Bounds webhook-controlled labels; adds `rule_group` to the fingerprintless identity | **Review carefully.** Table locks, ordering, and the delete-then-replace-trigger sequence |
| `server/migrations/000137`–`000139` | Three partial indexes, built `CONCURRENTLY` | Index shape vs. the queries above |
| `server/internal/domain/alerts/render.go` | Shared `alertGroup` rollup; `groupLine` replaces `alertLine`; `deviceLabel` moved here | Digest output shape |
| `server/internal/domain/alerts/deliver.go` | `deviceSuffix` removed (superseded by `deviceLabel`) | Deletion only |
| `server/internal/handlers/interceptors/config.go`, `middleware/rpc_permissions.go` | New RPC registered session-only + `alert:read` | One line each — verify the RPC is not left ungated |
| `server/cmd/fleetcli/cmd_alerts.go`, `tools/generate-fleet-cli/commands.json` | New `alerts history active` leaf; new flags on `alerts history list` | Generated from the manifest |
| `client/.../api/useActiveAlertGroups.ts` | **New.** Polls the rollup; stops polling on permission denial | Denial handling |
| `client/.../api/usePagedAlerts.ts` | **New.** Replaces `useActiveAlerts` + `useAlertHistory` with one keyset-paging hook | Both deleted hooks folded into one |
| `client/.../components/ActiveAlertsCard.tsx` | Rewritten: per-rule rows, callouts for device-less rules, opens the modal | The two hardcoded classification allowlists are gone |
| `client/.../components/AlertInstancesModal.tsx`, `PagedAlertTable.tsx`, `lib/alertInstanceColumns.ts`, `lib/alertCountLabels.ts` | **New.** Drill-in modal plus the table/columns it shares with the history feed | Shared config, no per-table drift |
| `client/.../components/HistoryTable.tsx`, `alertColumns.tsx`, `types/index.ts` | Re-pointed at the shared table and hook | Mostly deletion |
| `server/generated/**`, `client/src/protoFleet/api/generated/**` | Regenerated from the proto and sqlc sources | **Generated — skip** |

- **Keyset the drill-in on `alert_key`, not `history_id`.** `history_id` is rewritten on every re-assert, so a row-id cursor loses rows mid-page during exactly the churn an outage produces. `alert_key` never moves. Cost: the drill-in pages in key order rather than most-recent-first.
- **Unfiltered `active_only` keeps its unpaged contract.** Only a *filtered* `active_only` request pages, so clients predating this change still get the whole firing set under the response cap. Alternative — paging all `active_only` calls — would have silently truncated old clients to one page.
- **Cache store rows, not responses.** Summary visibility depends on the caller's `miner:read` grant. Caching the finished response would leak device detail between callers sharing a snapshot; there is a test for this.
- **Detached fetch context.** The shared aggregate runs on a context detached from whichever caller won the singleflight slot, so an operator navigating away mid-poll cannot cancel the fetch for everyone else waiting on it. Cost: a fixed timeout is needed to bound the goroutine.
- **Bound the indexed labels before adding the indexes (`000136` before `000137`–`000139`).** The new indexes store `alert_name`, `rule_group` and `device_id` as raw webhook-supplied text; an oversized value would push an index entry past Postgres's btree tuple limit and fail the insert carrying it — rolling back the whole ingest batch. The trigger **skips** such rows rather than truncating, because the rollup groups on these columns and a truncated value would merge two distinct rules. `notification_history` still records the event either way. Bounds are 190 (Grafana's rule-title cap) and 255 (the `device_identifier` column).
- **Reject `chr(31)` in the same guard.** The fingerprintless fallback key joins its parts on that separator, so a label containing one could hash onto another rule's identity. Alternative — a length-prefixed or JSON encoding — would have changed the key shape for existing rows.
- **Add `rule_group` to the fingerprintless identity.** `000094` keyed on name and device alone, so two same-named rules firing on one device in different groups overwrote each other's active state. Existing fingerprintless rows hash differently now and are dropped in the same lock, since they are derived state rebuilt on the next assert.
- **Device-less rules are identified by the server's `device_count`, not a client allowlist.** Replaces two hardcoded sets of rule-group and template names that had to be kept in sync with every new rule scope.
- **Indexes built `CONCURRENTLY`.** `notification_active` is on the alert-ingest write path, so a blocking build would stall webhook persistence. Cost: each index migration must be the sole statement in its file and cannot run in a transaction.

## Testing & validation

- **Store integration tests:** rollup grouping and ordering, newest-instance severity/summary, the keyset drill-in across a mid-read re-assert (the failure mode a `history_id` cursor would hit), `rule_group` presence matching including the `""` case, and the trigger's length and separator guards.
- **Handler tests:** cursor shape per branch, the unpaged unfiltered `active_only` contract, rejection of a malformed history cursor, concurrent callers sharing one aggregate, and two callers with different `miner:read` grants being redacted independently off one cached snapshot.
- **Renderer tests:** Slack and webhook rollups, including a device-less group, one miner with two alerts, and the rule-level summary surviving the rollup.
- **Client tests:** card rows vs. callouts, the drill-in filtering on the stored (not display-renamed) alert name, and the device-less table dropping its identity columns.
- **Security review:** the repo's Codex security review run locally to a clean pass. It caught two issues now fixed here — an unbounded `device_id` reaching the new index, and the separator ambiguity above.
- **Not covered:** no load test at true outage scale; the index benefit is measured from a synthetic 100k-firing-row set (266ms/101k buffers before, 15ms/1.8k after). No E2E scenario exercises the drill-in modal.


server/internal/handlers/alerts/active_groups_cache.go
```diff
@@ -0,0 +1,80 @@
...
+	lru "github.com/hashicorp/golang-lru/v2/expirable"
+	"golang.org/x/sync/singleflight"
+
+	"github.com/block/proto-fleet/server/internal/domain/fleeterror"
+	"github.com/block/proto-fleet/server/internal/domain/notificationhistory"
+)
+
+const (
+	// Under the dashboard's 15s poll, so a lone viewer never waits an extra interval for a change.
+	activeGroupsTTL = 10 * time.Second
```
ankitgoswami: do we really need this cache if the query only takes 15ms for 100k rows? Also, with these constants, it will only help when there are multiple concurrent viewers in the same 10 second window.

server/sqlc/queries/notification_history.sql
```diff
@@ -146,3 +147,89 @@ WHERE na.organization_id = sqlc.arg('organization_id')
...
+    latest.severity::text AS severity,
+    latest.template::text AS template,
+    latest.summary::text AS summary,
+    g.alert_count,
+    g.device_count,
+    g.first_started_at
+FROM grouped g
+-- severity/template/summary off the newest instance, not aggregated, so a group never blends two instances'
+-- values; seeking per capped-page row beats sorting every instance of every group to take one.
+CROSS JOIN LATERAL (
```
ankitgoswami: Do we need severity, template, and summary in the rollup response? This lookup exists only to decorate each group with its newest instance, and it also requires the 000137 index plus extra proto/domain/client fields and summary-redaction logic. Maybe we keep the rollup to (alert_name, rule_group), counts, and first_started_at, and leave instance details to the drill-in? That would make this a single grouped query and remove a big chunk of the schema and API surface.

server/internal/handlers/alerts/handler.go
```diff
@@ -386,29 +389,32 @@ func (h *Handler) ListAlerts(ctx context.Context, req *connect.Request[alertsv1.
...
-		// paging. Over-fetch by one so the response can flag (rather than silently swallow) an alert storm past the cap.
+	// Each branch names its own cursor as it dispatches, so the shape a caller gets back can't drift from the
+	// branch that produced it. Only a filtered active_only pages: unfiltered keeps the contract it had before.
+	var (
+		rows      []notificationhistory.StoredNotification
+		pageLimit = historyPageLimit(req.Msg.GetPageSize())
+		cursorOf  func(notificationhistory.StoredNotification) string
+	)
+	switch {
+	case req.Msg.GetActiveOnly() && (req.Msg.GetAlertName() != "" || req.Msg.RuleGroup != nil):
```
ankitgoswami: Could we make the filtered active mode require alert_name and always treat rule_group as an exact value, where "" means the unlabeled group? Right now, omitting rule_group searches every group with that alert name, while supplying only rule_group searches for an empty alert name. Removing those partial modes would make RuleGroup a plain string and simplify the SQL.

ankitgoswami (approved):

PR #918, opened 2026-08-13
feat(alerts): move active alerts into a header pill & slack cleanup

illegalprime:
Reviewable diff: +546/-228 across 25 files (excludes generated, test, and story files).

## Summary

Active alerts move out of the dashboard card and into a pill in the top-right header widgets, so a firing fleet is visible from every route instead of only the dashboard. The pill appears only while something is firing (or while the poll is broken), and opens a wide popover listing the per-rule rollup with the existing affected-miners drill-in. Alert rules that fire on no miner — curtailment sources, ingest stalls — now state what fired inline rather than offering a drill-in with nothing to list. On the notification side, Slack messages now name which fleet instance sent them, since several instances can post to one channel.

## How it works

The poll moves from the dashboard card up into the app shell. `AppLayout` mounts `useActiveAlertsPillData`, which gates on the `alert:read` permission and the runtime alerts probe before `useActiveAlertGroups` starts polling, and skips the poll entirely on focused detail routes where the header is hidden. Its `hasVisiblePill` flag feeds the existing header-widget layout ladder alongside the curtailment, schedule, and update pills, so the phone-stacked header grows a rung when alerts are up.

Server-side, `ListActiveAlertGroups` still returns one row per rule with counts and no device identity. The rollup query is now a CTE plus a `LEFT JOIN LATERAL` that picks the newest device-less instance's summary for the groups that have one, bounded in SQL (500 chars for the summary, 64 for the template) since neither column is length-checked on insert. The template rides along because the handler decides from it whether the summary may name a miner: it runs through `visibleSummary` with `miner:read` false — the same gate the history API applies — so a device template's summary is withheld on an endpoint that only holds `alert:read`. A truncated template matches no known template and so fails closed.

Because a withheld summary leaves a row that can't describe itself, a device-less group stays drillable when its summary was withheld or when more than one instance is firing, the latter under a pluralized title from a small closed map (only built-in device-less rules roll up more than one instance; user rules fire per miner and are counted as miners).

Two client-side reliability fixes follow from the hook moving into the shell, where no route change can remount it to recover:

- The `alerts/enabled` probe had no deadline. A connection held open without answering left the shared in-flight promise pending forever and every later probe joined it, hiding the alerts surface for the whole session. It now has a 10s abort deadline and retries with backoff to 60s, and distinguishes "no answer" (`null`) from an answered "disabled".
- A permission denial used to stop the active-groups poll for good. It now backs off to a 5-minute retry so a restored grant is picked up without a reload, and overlapping polls drop the older reply via a request counter.

Slack rendering: the `header` block is replaced by a bold mrkdwn section carrying the title, which lets the instance name be a link. `FLEET_PUBLIC_URL`'s host names the sender and doubles as the link target, so no new config is needed; a value with no host falls back to the bare product name. The title's dot now reflects the batch's worst severity instead of always being red. MACs render as inline code — the only mrkdwn context where Slack leaves `:shortcode:` alone, so a colon-separated MAC no longer interpolates `:ab:` as an emoji.

## Diagrams

Poll and render path:

```mermaid
flowchart TD
    A["AppLayout (shell)"] -->|"enabled: header visible"| B["useActiveAlertsPillData"]
    B --> C["useHasPermission alert:read"]
    B --> D["useAlertsEnabled probe"]
    C --> E["useActiveAlertGroups poll"]
    D --> E
    E -->|"ListActiveAlertGroups"| F["alerts handler"]
    F -->|"authorize alert:read"| G["ListActiveNotificationGroups"]
    G -->|"CTE + LATERAL"| H["notification_active"]
    F -->|"visibleSummary(summary, template, false)"| I["ActiveAlertGroup rows"]
    I --> J["hasVisiblePill"]
    J -->|"true"| K["ActiveAlertsPill in header"]
    K --> L["ActiveAlertsPopover (wide)"]
    L -->|"device_count > 0"| M["AlertInstancesModal drill-in"]
    L -->|"device_count = 0, summary shown"| N["inline summary, no drill-in"]
```

Summary redaction decision:

```mermaid
stateDiagram-v2
    [*] --> Grouped
    Grouped --> HasDevices: "device_count > 0"
    Grouped --> DeviceLess: "device_count = 0"
    HasDevices --> NoSummary: "SQL CASE drops lateral result"
    DeviceLess --> TemplateGate: "summary + template (bounded 500/64)"
    TemplateGate --> Withheld: "device template, no miner:read"
    TemplateGate --> Shown: "non-device template"
    Withheld --> Drillable: "row must still be explorable"
    NoSummary --> Drillable
    Shown --> [*]
```

## Areas of the code involved

| --- | --- | --- |
| `proto/alerts/v1/alerts.proto` | `ActiveAlertGroup.summary` added as field 6 | Wire-compatible additive field; check the comment's scope claim (device-less only) |
| `server/generated/**`, `client/**/generated/**` | Regenerated Go/TS/sqlc | generated — skip |
| `server/sqlc/queries/notification_history.sql` | Rollup rewritten as CTE + `LEFT JOIN LATERAL`; summary/template bounded, ordering repeated after the join | The main query change: index usage, the `device_id = ''` prefix descent, and the re-applied `ORDER BY` |
| `server/internal/handlers/alerts/handler.go` | Populates `Summary` via `visibleSummary(..., false)` | The authorization boundary — an `alert:read`-only endpoint must not leak miner-naming text |
| `server/internal/domain/alerts/render.go` | Title becomes a linked mrkdwn section naming the instance; severity-based dot; MACs as inline code; backtick stripped in `escapeMrkdwn`; section cap moved into `mrkdwnSection` | Mrkdwn injection surface — check that field-escaping still holds for user-controlled text |
| `server/internal/domain/{notificationhistory,stores/sqlstores}` | Thread `Summary`/`Template` through the rollup row | Plumbing |
| `server/cmd/fleetd/config.go` | `FLEET_PUBLIC_URL` help text notes its second role | No behavior change |
| `client/src/protoFleet/components/PageHeader/*` | New `ActiveAlertsPill`, `ActiveAlertsPopover`, `useActiveAlertsPillData`; `headerWidgetLayout` gains an alerts slot | The new surface and its visibility gate |
| `client/src/protoFleet/components/AppLayout/AppLayout.tsx` | Mounts the poll, skips it when the header is hidden | Where the poll's lifetime is now decided |
| `client/src/protoFleet/features/alerts/api/useAlertsEnabled.ts` | Abort deadline, retry with backoff, `null` for "no answer" | Session-long hang fix |
| `client/src/protoFleet/features/alerts/api/useActiveAlertGroups.ts` | Denial backs off instead of stopping; stale-response guard; `loading` dropped | Poll lifecycle under permission changes |
| `client/src/protoFleet/features/alerts/components/ActiveAlertsCard.tsx` | Deleted, along with its dashboard mount | Confirm nothing else referenced it |
| `client/src/shared/components/Popover/*` | New `wide` size (2× medium, viewport-capped) | Additive shared-component change; no existing size affected |

- **Summary picked in a lateral, gated in the handler** — rather than adding a per-instance endpoint or widening the drill-in. The lateral is bounded to groups that need it, and the gate reuses the history API's `visibleSummary` instead of trusting `device_count = 0` as proof the text names no miner.
- **Bounds in SQL, not in Go** — `LEFT(...)` at 500/64 keeps a header-polled one-liner from shipping a whole `TEXT` column; migration 000136 bounds only the indexed columns, so this is the first length limit these two see. Truncating the template intentionally fails the visibility decision closed.
- **Denied poll backs off rather than stops** — the hook now outlives any route change, so a permanent stop would strand a restored grant until reload. Costs one request per 5 minutes for a site-scoped grant that can't reach the org RPC.
- **New `wide` popover size over a one-off width on the pill** — a scoped additive variant in the shared component rather than a global token change or an inline override.
- **Instance name from `FLEET_PUBLIC_URL`'s host** — reuses the link target already required for notifications instead of introducing an instance-name setting.
- **`escapeMrkdwn` strips backticks instead of escaping `*`/`_`/`~`** — a code span opened in one field would close at the next backtick anywhere in the section; the other chars are left so user text can still format itself and interpolate emoji.

## Testing & validation

- `just lint` — clean (buf lint, `eslint --max-warnings 0`, golangci-lint 0 issues).
- `just gen` — no drift; generated Go/TS/sqlc in the diff match their sources.
- Client: `vitest run` over `PageHeader`, `AppLayout`, `features/alerts`, `features/dashboard` — 38 files, 266 tests passing. New coverage for the pill, popover, header-widget count, `useActiveAlertGroups` (supersession, denial backoff), `useAlertsEnabled` (timeout, retry, `null` vs disabled), and the alerts API shape.
- Server: `go test -p 1 -count=1 ./internal/domain/alerts/... ./internal/domain/stores/sqlstores/... ./internal/handlers/alerts/... ./cmd/fleetd/...` — all passing, including the sqlstores suite against real Postgres. New assertions cover the rollup summary/template columns, the handler's redaction gate, and the Slack title/MAC rendering.
- Local Codex security review (`codex-security-review-local.sh`) — overall risk NONE.

Not covered:

- Playwright E2E was not run; worth a manual pass on the header pill and drill-in since this changes a shell-level surface.
- The plugin contract suite was not run — the proto change is `alerts.proto`, which has no miner-driver surface.
- Slack output is verified by unit assertions on the Block Kit payload, not against a live workspace.


ankitgoswami (approved):
lgtm

PR #921, opened 2026-08-14
refactor(updates): make one-click upgrade operation state host-authoritative

mcharles-square:
Reviewable diff: +1067/-493 across 12 files (excludes generated, test, and story files).

## Summary

One-click upgrades now use one durable operation identity from the browser through Fleet to the host updater, so ambiguous responses and reloads can be reconciled without target/version guesses. A host-owned operation incarnation plus explicit outcome revision prevents a stale tab from dismissing either a rewritten result or a later operation that deliberately reuses a historical UUID. The browser keeps only an authentication-scoped correlation lock while the exact operation is pending or active; the host remains authoritative for progress, outcomes, and acknowledgement.

This is a semantic simplification rather than a line-count reduction: three operation interpretations collapse to one host authority and four browser inference rules disappear, while explicit replay and compare-and-set safeguards increase the affected production footprint by 6.1%.

## How it works

1. The browser generates a canonical UUIDv4, records that ID and target in session storage for the current authentication session, then submits the mutation. The record keeps competing actions locked across reloads until the exact operation becomes terminal or the request is conclusively rejected.
2. Fleet validates the ID and first returns an exact retained ID/target replay, even if the installed version or release offer has since changed. Only a successful status read proving the ID is not current may enter fresh eligibility; an unavailable probe remains retryable instead of becoming a definitive offer rejection. Genuinely new IDs still require the currently eligible channel target before Fleet forwards them to the local updater.
3. The updater persists the operation with a host-owned start time that is strictly later than the retained prior operation, even if the host clock is frozen or moves backward. Its first terminal result receives outcome revision 1; startup or HA recovery increments the revision whenever it materially rewrites the visible outcome and clears any earlier acknowledgement.
4. The browser renders host state directly. Success is acknowledged using the exact ID, start time, and revision before Fleet reloads; a failed or stale acknowledgement leaves the result visible. Failures use the same compare-and-set contract.
5. Fleet makes activity writes idempotent by organization and the host-generated operation incarnation (ID plus start time), adding outcome revision for acknowledgements. Exact retries converge without allowing historical client UUID reuse to hide later audits.

## Diagrams

```mermaid
flowchart LR
  B["Browser upgrade hook"] -->|"Persist auth-scoped ID + target"| C["Correlation lock"]
  B -->|"UUID + target"| F["Fleet update service"]
  F -->|"Exact replay or validated fresh target"| U["Host updater API"]
  U --> S["Durable operation state"]
  S -->|"Operation + incarnation + outcome revision"| U
  U --> F
  F --> B
  B -->|"Acknowledge ID + start time + revision"| F
  F -->|"Compare-and-set"| U
```

```mermaid
stateDiagram-v2
  [*] --> Active: "trigger admitted (revision 0)"
  Active --> Terminal1: "succeed or fail (revision 1)"
  Terminal1 --> Acknowledged1: "acknowledge incarnation + revision 1"
  Terminal1 --> Terminal2: "startup/recovery rewrite (revision 2)"
  Acknowledged1 --> Terminal2: "new recovery outcome clears acknowledgement"
  Terminal2 --> Acknowledged2: "acknowledge incarnation + revision 2"
```

## Areas of the code involved

| --- | --- | --- |
| `proto/instance/v1/updates.proto` | Added caller operation ID, outcome revision, and expected operation start time/revision | Defines the end-to-end identity contract |
| Generated Go/TypeScript protobufs | Regenerated from the proto source | Generated — skip |
| `server/internal/domain/updates` and handler | Replays exact retained operations before fresh eligibility and scopes audit idempotency to a host operation incarnation | Fleet remains the authorization, eligibility, and audit boundary |
| `server/internal/updaterapi` | Serializes incarnation-aware acknowledgement and verifies response identity | Protects the local transport boundary from mismatched responses |
| `server/internal/updater` | Assigns strictly increasing operation incarnations, advances terminal revisions, and compare-and-sets acknowledgements | Review recovery rewrites and durable-state semantics; activation ordering is unchanged |
| `client/.../useUpgradeOperation` | Uses host-authoritative polling plus one auth-scoped active-operation correlation record | Preserves reload locking without restoring browser phase, outcome, or remediation inference |
| `client/.../Updates` and modal | Acknowledges successful outcomes before reload and retains them on failure | Prevents terminal successes from resurfacing indefinitely |
| `docs/plans/...operation-authority-tdd.md` | Records the identity, reload-lock, audit, and outcome-revision contracts | Makes the simplification boundary explicit |

- Caller-generated UUIDs replace server-generated IDs so a lost response has an exact reconciliation key.
- One authentication-scoped correlation record survives while the operation is pending or active; the browser does not persist phase, outcome, or dismissal authority.
- `AlreadyExists` remains on bounded reconciliation because it confirms a conflicting host mutation or recovery lock; only conclusive non-admission errors clear correlation immediately.
- Exact current-operation replay precedes fresh offer validation because it returns an already-admitted mutation; new IDs still revalidate channel and target.
- A failed replay-status probe returns sanitized `Unavailable` because the host may already own that ID; caller cancellation and deadlines remain distinguishable.
- A strictly increasing host start time distinguishes historical UUID reuse in acknowledgement and audit identities without adding a permanent UUID ledger to the updater.
- Explicit monotonic outcome revisions replace using `completed_at` as both a timestamp and implicit outcome identity.
- Intermediate RC protocol/state compatibility is intentionally omitted because this feature has not shipped in a stable release.
- Activation, directory swap, fsync, HA takeover, self-refresh, and trust ordering remain unchanged.

## Testing & validation

- `just lint` passed across proto, client, server, and plugins.
- Affected Go packages passed: updater, updater API, update domain, and update handlers.
- Upgrade hook, Updates page, and operation modal suites passed: 105 tests.
- TypeScript `tsc --noEmit` passed.
- Added coverage for reloads during admission and active operations, storage failure, auth-session and overlapping-mutation races, acknowledge-before-reload, exact-ID replay after offer/version changes, replay-probe failures and caller cancellation, conflicting-operation reconciliation, HA-completion mismatch, audit incarnation keys, constant/backward-clock UUID reuse, incarnation-keyed upgrade logs, monotonic remediation cutoffs, client incarnation/outcome deduplication, auth-session modal ownership, incarnation/revision CAS, recovery rewrites, and transport response mismatches.
- External kill-window qualification and a real root/systemd/HA deployment were not rerun; this change deliberately does not alter those execution paths.


server/internal/updater/manager.go
```diff
@@ -1150,6 +1200,12 @@ func (m *Manager) trigger(targetVersion, operationID string, idempotent, complet
 		return updaterapi.Operation{}, fmt.Errorf("reserve updater operation log capacity: %w", err)
 	}
 	now := m.cfg.Now().UTC()
+	// StartedAt is part of the acknowledgement compare-and-set identity. Keep
+	// each freshly admitted incarnation strictly newer than the one retained in
+	// state even when the host clock is coarse, frozen, or moves backward.
+	if m.operation != nil && !now.After(m.operation.StartedAt) {
+		now = m.operation.StartedAt.Add(time.Nanosecond).UTC()
```
rl-block, line 1207: Looks like this gives a reused UUID a distinct host incarnation, but `run` still derives the log name from the UUID alone and opens it with `O_EXCL`. In an A -> B -> A sequence, the retained first log can make A2 fail before download. Could we perhaps include `StartedAt` in the log identity and pruning protection, then have the reuse test assert progress beyond log creation?

server/internal/updater/manager.go
```diff
@@ -785,19 +789,25 @@ func (m *Manager) RecoverApplication() error {
...
 	if err := m.runHACommand(context.Background(), m.cfg.ActivationTimeout, deployment, io.Discard, "app-start", version, "any"); err != nil {
 		recoveryErr := fmt.Errorf("restart interrupted HA application: %w", err)
+		now := m.cfg.Now().UTC()
+		advanceOutcomeRevision(m.operation)
 		m.operation.Phase = updaterapi.PhaseFailed
 		m.operation.Message = "HA application recovery failed"
 		m.operation.Error = recoveryErr.Error()
-		m.operation.UpdatedAt = m.cfg.Now().UTC()
+		m.operation.UpdatedAt = now
+		m.operation.CompletedAt = &now
```
rl-block: Do we want to keep this remediation cutoff monotonic relative to any existing startup proof? If `cfg.Now()` moves backward, `Status` can see the proof mtime after `CompletedAt`, derive `Acknowledged=true`, and hide this new recovery revision. A backward-clock test with an existing proof might help cover the case.

client/src/protoFleet/features/updates/api/useUpgradeOperation.ts
```diff
@@ -108,393 +167,310 @@ export const isUpgradeTerminal = (phase: UpgradePhase) =>
...
     }
-    setConnectionLost(true);
-    allowManualFallbackAfterTimeout();
-    return true;
-  }, [allowManualFallbackAfterTimeout, updateReconciling]);
-
-  const resolveRecoveredOperationMiss = useCallback(() => {
-    if (!reconcilingRef.current || !trackedOperationRef.current?.id) {
-      return false;
+    const revision = `${next.id}:${next.outcomeRevision}`;
```
rl-block: Should these ownership and deduplication keys use the full operation incarnation (`id` + `startedAt`) and outcome revision? A later operation may reuse the UUID and also start at revision 1, so the current ID-only submitted set and `id:revision` key can skip its status refresh; the modal auto-open key has the same shape. A same-ID/different-`startedAt` test might make this behavior explicit.

rl-block (approved):

server/internal/updater/manager.go
mcharles-square: 🤖 Updated by Codex: Fixed in fcfaa93c7. Upgrade-log creation and every pruning protection now derive from the full host incarnation (operation ID plus UTC StartedAt), so A→B→A historical UUID reuse receives a distinct fixed-format log. The regression verifies A2 reaches checksum verification rather than failing at O_EXCL.

server/internal/updater/manager.go
mcharles-square: 🤖 Updated by Codex: Fixed in fcfaa93c7. Terminal rewrites now use a monotonic remediation cutoff: the maximum of now, StartedAt, prior UpdatedAt/CompletedAt, and any regular startup-proof mtime. Because remediation requires proof time to be strictly later, equality keeps the new revision visible even when the wall clock moves backward. Recovery and startup regressions cover both success and failure paths.

client/src/protoFleet/features/updates/api/useUpgradeOperation.ts
mcharles-square: 🤖 Updated by Codex: Fixed in fcfaa93c7. Browser ownership and deduplication now use the full incarnation (ID plus StartedAt), while terminal/modal identity also includes OutcomeRevision. Ownership is established only by a directly correlated trigger response, and modal keys are scoped to the authenticated session. Reused-incarnation, rewritten-outcome, stale-pending, and auth-switch regressions were added.

server/internal/updater/manager.go
```diff
@@ -2119,6 +2182,7 @@ func (m *Manager) loadState() error {
 		return fmt.Errorf("successful operation retains an unexpected activation marker")
 	}
 	wasTerminal := op.Phase.Terminal()
+	terminalOutcomeBeforeRecovery := op
```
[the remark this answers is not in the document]
mcharles-square: 🤖 This is mechanically valid for an in-place upgrade from an intermediate v0.2.10 RC that persisted unversioned terminal state, but that path is outside this PR’s supported rollout contract. v0.2.9, the previous stable, predates the host updater entirely, and no v0.2.10 stable exists. A stable-to-next-stable rollout therefore starts with no legacy updater state, while every terminal write created by this implementation advances to revision 1 before persistence or self-exec. The TDD explicitly treats intermediate RC state as disposable, so normalizing zero to one would add the compatibility shim this simplification intentionally excludes. We recommend leaving this unchanged unless RC-to-RC compatibility becomes a supported requirement.

Issue #909, opened 2026-08-11
Add building, rack, and group targets to curtailment


negarn:
🤖 Updated by Negar's AI agent.

## Summary

Add Buildings, Racks, and Groups to the shared **Apply to** targeting used by:

- Energy → New curtailment
- Settings → Create/Edit response profile

Each curtailment or response profile has exactly one terminal miner scope. The
backend resolves that logical selector against current topology; the browser
does not expand topology selections into miner snapshots.

## Source of truth

The [curtailment building, rack, and group targets plan](https://github.com/block/proto-fleet/blob/main/docs/plans/2026-08-11-curtailment-building-rack-and-group-targets-plan.md)
defines the detailed behavior, security boundaries, rollout, and verification.
[PR #926](https://github.com/block/proto-fleet/pull/926) updates the plan and
contract foundation to the single-terminal-scope model. The plan is
authoritative if this tracker and the plan differ.

Exact target-builder interactions and presentation will follow the pending
design without changing these semantics. This issue tracks the implementation
PRs; merging the plan or contract foundation does not complete the feature.

## Required behavior

- Represent exactly one terminal miner-scope type: Whole organization, Sites,
  Buildings, Racks, Groups, or explicit Miners.
- Allow multiple IDs within the selected type, with deterministic
  deduplication. Reject mixed scope types and Whole organization combined with
  a narrower selector.
- Reuse Schedules picker and filtering behavior for both curtailment modal
  variants. Navigation ancestors constrain child catalogs but are not submitted
  as additional targets. Changing the terminal type replaces the prior miner
  scope.
- Keep Schedules' existing multi-type union semantics; only picker behavior is
  shared with curtailment.
- Keep Infrastructure as an independent facility-fan selection that never
  contributes miners.
- Use the same canonical terminal scope for validation, confirmation copy,
  Preview, Start, profile save/reload/edit/test, automation, active events, and
  history. Unsupported or missing typed scopes must never fall back to Whole
  organization.
- Preserve logical topology selectors in response profiles so later manual and
  automated executions resolve current topology.
- Apply **Target all paired miners** to FULL_FLEET with any terminal scope type.
  Unpaired miners may belong to site/building/rack/group selectors and can be
  admitted when they become paired.
- Preserve the execution semantics from the plan:
  - Non-`FULL_FLEET` events resolve once and freeze concrete targets.
  - Unflagged `FULL_FLEET` events follow topology selectors; explicit miner IDs
    remain Start-time snapshots.
[...]

PR #926, opened 2026-08-17
Define typed curtailment scope contract

negarn:
Reviewable diff: +761/-484 across 19 files (excludes generated, test, and story files).

## Summary

This PR establishes the typed, single-terminal-selector contract needed to add building, rack, and group targeting to curtailment and response profiles. It centralizes current site/miner scope handling across the client and server, persists schema-versioned logical selectors, and fails closed for missing, mixed-type, or not-yet-resolvable topology scopes. The topology resolver and shared Schedules-style drill-down UI remain follow-up work tracked by #909.

## How it works

Preview, Start, and response-profile requests now carry exactly one terminal scope type: whole organization, sites, buildings, racks, groups, or explicit miners. Multiple values of the same type are allowed; selecting different types as a union is rejected. Infrastructure/facility-fan selection remains independent of the miner scope.

The client routes its existing site/miner builders, profile adapters, hydration, and display mapping through one scope module. At the RPC boundary, translators normalize repeated entries, enforce raw input limits and schema-version rules, and pass one typed domain scope through authorization, persistence, and event/profile mapping. Building, rack, and group contracts can round-trip, but execution deliberately returns `Unimplemented` and requires organization-wide authorization until topology membership and complete site coverage can be resolved safely.

```mermaid
flowchart LR
    UI["Curtailment or response-profile flow"] --> CS["Canonical client scope module"]
    CS --> RPC["Preview, Start, or profile RPC"]
    RPC --> TR["Translate and validate one terminal type"]
    TR --> DS["Typed domain scope"]
    DS --> SP["Persist and round-trip logical selector"]
    DS --> EX{"Selector executable now?"}
    EX -->|"Whole org, sites, miners"| RUN["Existing resolution and execution"]
    EX -->|"Buildings, racks, groups"| CLOSED["Fail closed pending topology resolver"]
```

## Areas of the code involved

| --- | --- | --- |
| `proto/curtailment/v1` | Added building, rack, and group scope cases, schema versions, and cardinality bounds; removed the unused generic device-set case | Defines the public contract and rejects ambiguous request shapes |
| Generated protobuf and sqlc output | Regenerated from the source contracts and queries | Generated — skip |
| `client/src/protoFleet/api` and energy adapters | Added canonical scope constructors/parser and routed request, profile, preview, event, and display adapters through it | Prevents the site/miner mapping from diverging across entry points |
| `server/internal/handlers/curtailment` | Normalizes repeated scopes, applies bounds/version checks, maps typed scopes, and derives conservative authorization requirements | Main RPC validation and security boundary |
| `server/internal/domain/curtailment` | Enforces exactly one selector type, serializes versioned logical scopes, and removes implicit whole-org widening | Main fail-closed domain invariant and persistence model |
| `server/sqlc/queries` and SQL store | Carries typed scope JSON/version through event decision snapshots | Keeps persisted execution state reviewable and round-trippable |
| `docs/plans` | Updates the accepted implementation direction to Schedules-style parent-constrained drill-down with no cross-type union | Records the remaining resolver and UI phases under #909 |

- A request may contain multiple IDs only within one terminal selector type; parent selections are future drill-down filters, not additional submitted targets.
- Logical selectors are persisted instead of expanding them at the API boundary, preserving the planned frozen versus topology-following execution semantics.
- Topology selectors fail closed until resource validation, membership expansion, and full authorization coverage can happen together in the backend resolver.
- Infrastructure remains a separate fan-control envelope and cannot make an otherwise missing miner scope valid.
- Raw cardinality is checked before deduplication and again in the domain so alternate callers cannot bypass the contract.

## Testing & validation

- `just gen`
- Client lint
- ProtoFleet production build
- 401 focused client tests across 11 files
- Server lint
- `go build ./...`
- Curtailment domain and handler test suites
- SQL store compile-only test (`go test ./internal/domain/stores/sqlstores -run '^$'`); database-backed store integration tests were not run locally
- `git diff --check`

Part of #909


server/internal/domain/curtailment/response_profile.go
```diff
@@ -564,36 +565,72 @@ func ScopeFromJSON(scopeJSON []byte) (Scope, bool, error) {
...
+		"whole_org",
+		"site_id",
+		"site_ids",
+		"building_ids",
+		"rack_ids",
+		"group_ids",
+		"device_identifiers",
+	} {
+		if _, ok := fields[name]; ok {
+			selectorFieldCount++
```
rl-block, line 613: Perhaps we could treat selector keys whose JSON value is `null` as absent here?

The pre-PR `ScopeTypeMixed` encoder always wrote both keys, so existing multi-site rows contain `"device_identifiers": null`; counting that key makes `ScopeFromJSON` reject those persisted events and profiles, which stops reconciliation and can fail profile listing.

Maybe we could ignore null-valued keys while still rejecting multiple non-null selectors, with a regression test for the legacy JSON shape.

client/src/protoFleet/api/useCurtailmentResponseProfiles.ts
```diff
@@ -308,32 +320,32 @@ function getApiResponseProfileScopeValues(
...
   const minerCount = values.deviceIdentifiers.length;
   const minerSummary = minerCount === 1 ? "1 miner" : `${minerCount} miners`;
   const siteSummary =
     siteIds.length === 1
       ? getResponseProfileSiteNameForId(values, siteIds[0]) || `Site ${siteIds[0]}`
       : `${siteIds.length} sites`;
-  if (siteSelection === "site" && siteIds.length > 0 && minerCount > 0) {
-    return `${siteSummary} + ${minerSummary}`;
+  if (minerCount > 0) {
+    return minerSummary;
```
rl-block: These summary branches should probably call one shared helper with `CurtailmentSettingsPage.tsx`?

Both copies changed the same ordering in this PR, while the plan calls for the canonical scope module to own summaries. A shared helper with a caller-provided fallback label could keep the two profile views from drifting.

rl-block (approved):

PR #936, opened 2026-08-18
Publish the HA service CA through the virtual IP

ankitgoswami:
Reviewable diff: +70/-2 across 4 files (excludes generated, test, and story files).

## Summary

Operators can retrieve the HA cluster's public service CA through the working VIP after installation. The authenticated `ha-a` session prints the trust fingerprint, and the final `ha-c` session prints a constrained download and comparison workflow without exposing any private credential.

**Stack:** [#935](https://github.com/block/proto-fleet/pull/935) → [#936](https://github.com/block/proto-fleet/pull/936) (this PR). The diff is relative to #935, which establishes the one-command-per-host install and automated peer handoff. This PR adds only client trust bootstrap; automatic trust-store mutation remains out of scope.

## How it works

1. During cluster preparation, `ha-a` computes the SHA-256 fingerprint from the generated service CA certificate and prints it in the authenticated SSH session.
2. Installation keeps private keys and credentials at mode `0600`, while public certificates and public keys are installed root-owned at mode `0644` so the unprivileged web process can read them.
3. The Fleet web container mounts only `service-ca.crt` at `/proto-fleet-ha-service-ca.crt`. No directory of secrets and no private key enters the static content root.
4. After `ha-c` proves that the VIP serves active Fleet health, it prints commands that bypass proxy configuration, download the certificate, canonicalize a single PEM certificate, and display its fingerprint.
5. The operator compares that value with the fingerprint from `ha-a` before importing the canonical certificate. The installer does not change any client trust store.

```mermaid
sequenceDiagram
  participant A as "Authenticated ha-a session"
  participant V as "Fleet VIP"
  participant O as "Operator machine"
  A->>O: "Trusted CA SHA-256 fingerprint"
  O->>V: "Download public service CA"
  V-->>O: "service-ca.crt only"
  O->>O: "Canonicalize and calculate fingerprint"
  O->>O: "Compare before import"
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/ha/deployment/guided_install.go` | Prints the generated public CA fingerprint on `ha-a` | Establishes the authenticated trust anchor |
| `server/internal/ha/deployment/install.go` | Installs public material readably and prints the post-VIP download workflow | Separates public certificate access from protected secrets and avoids proxy-mediated LAN retrieval |
| `deployment-files/ha/fleet-compose.yaml` | Mounts one public CA file into nginx static content | Defines the exact public exposure boundary |
| `deployment-files/ha/README.md` | Documents fingerprint comparison and client trust | Keeps the operator flow visible without automating trust-store changes |

- Serve the CA as a static file through the existing VIP rather than adding an API or remote diagnostic endpoint.
- Use the `ha-a` SSH output as the trust anchor. The initial HTTPS download allows the private CA only for transport, so the operator must compare fingerprints before import.
- Canonicalize one parsed certificate before fingerprinting and importing it. This prevents appended certificate material from becoming trusted.
- Keep the public certificate world-readable while every private key and password remains owner-only.
- Do not automate browser or operating-system trust. Client trust policy remains an operator decision.

## Testing & validation

- `go test ./internal/ha/deployment -count=1`
- Hermit-backed `golangci-lint` for `./internal/ha/deployment/...`
- `git diff --check`

Focused coverage verifies the fingerprint format, public-versus-private install modes, exact nginx mount, witness-only instructions, proxy bypass, and canonical output. A real HTTPS download and browser trust import remain part of three-host manual qualification.

### DIFF
--- a/deployment-files/ha/fleet-compose.yaml
+++ b/deployment-files/ha/fleet-compose.yaml
@@ -28,3 +28,4 @@ services:
     volumes: !override
       - "${HA_SECRETS_DIR}/fleet-client.crt:/etc/nginx/ssl/cert.pem:ro"
       - "${HA_SECRETS_DIR}/fleet-client.key:/etc/nginx/ssl/key.pem:ro"
+      - "${HA_SECRETS_DIR}/service-ca.crt:/usr/share/nginx/html/proto-fleet-ha-service-ca.crt:ro"
--- a/server/internal/ha/deployment/guided_install.go
+++ b/server/internal/ha/deployment/guided_install.go
@@ -4,7 +4,10 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -216,6 +219,18 @@ func prepareAndInstallCluster(ctx context.Context, source string, release cluste
 	if err := prepareInstallBundles(exportDir, metadata); err != nil {
 		return fmt.Errorf("bundle generation failed; partial exports remain at %s: %w", exportDir, err)
 	}
+	haABundle := filepath.Join(exportDir, hostBundleName("ha-a"))
+	bundle, err := readHostBundle(haABundle)
+	if err != nil {
+		return err
+	}
+	fingerprint, err := serviceCAFingerprint(bundle.Secrets["service-ca.crt"])
+	if err != nil {
+		return err
+	}
+	if err := writeInstallerOutput(deps.output, "Public service CA SHA-256 fingerprint: %s\n", fingerprint); err != nil {
+		return err
+	}
 	peers := []struct {
 		role    string
 		address string
@@ -237,7 +252,6 @@ func prepareAndInstallCluster(ctx context.Context, source string, release cluste
 		}
 	}
 
-	haABundle := filepath.Join(exportDir, hostBundleName("ha-a"))
 	if err := installPreparedHost(ctx, source, haABundle, release, true, deps); err != nil {
 		return err
 	}
@@ -440,6 +454,20 @@ func decodeHostBundle(contents []byte) (preparedHostBundle, error) {
 		return preparedHostBundle{}, errors.New("host bundle rejected: etcd root password is only valid for ha-a")
 	}
 	return bundle, nil
+}
+
+func serviceCAFingerprint(contents []byte) (string, error) {
+	block, _ := pem.Decode(contents)
+	if block == nil || block.Type != "CERTIFICATE" {
+		return "", errors.New("host bundle contains an invalid public service CA")
+	}
+	digest := sha256.Sum256(block.Bytes)
+	encoded := strings.ToUpper(hex.EncodeToString(digest[:]))
+	pairs := make([]string, 0, len(encoded)/2)
+	for index := 0; index < len(encoded); index += 2 {
+		pairs = append(pairs, encoded[index:index+2])
+	}
+	return strings.Join(pairs, ":"), nil
 }
 
 func validateBundleMetadata(metadata bundleMetadata) error {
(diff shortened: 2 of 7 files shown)


deployment-files/ha/fleet-compose.yaml
```diff
@@ -28,3 +28,4 @@ services:
     volumes: !override
       - "${HA_SECRETS_DIR}/fleet-client.crt:/etc/nginx/ssl/cert.pem:ro"
       - "${HA_SECRETS_DIR}/fleet-client.key:/etc/nginx/ssl/key.pem:ro"
+      - "${HA_SECRETS_DIR}/service-ca.crt:/usr/share/nginx/html/proto-fleet-ha-service-ca.crt:ro"
```
rl-block, line 31: Should we pair this mount with an update-time permission migration? Existing HA installs keep `service-ca.crt` at root-owned mode 0600, and the rolling-update path does not change it, so nginx cannot serve the new URL. Perhaps we could migrate only this public certificate to 0644 before starting `fleet-client` and cover the pre-936 upgrade case.

server/internal/ha/deployment/guided_install_test.go
```diff
@@ -34,6 +35,12 @@ func TestGuidedInstallPreparesClusterAndInstallsHAA(t *testing.T) {
 		require.Equal(t, "operator", localUser)
 		require.FileExists(t, bundlePath)
 		requireMode(t, bundlePath, 0o600)
+		if expectedFingerprint == "" {
+			bundle, err := readHostBundle(bundlePath)
+			require.NoError(t, err)
+			expectedFingerprint, err = serviceCAFingerprint(bundle.Secrets["service-ca.crt"])
+			require.NoError(t, err)
```
rl-block, line 42: We could probably use a fixed certificate and a fingerprint precomputed with OpenSSL here? Calling `serviceCAFingerprint` to build the expected value means a digest or formatting bug would make both sides wrong in the same way.

server/internal/ha/deployment/guided_install.go
```diff
@@ -216,6 +219,18 @@ func prepareAndInstallCluster(ctx context.Context, source string, release cluste
...
 		return fmt.Errorf("bundle generation failed; partial exports remain at %s: %w", exportDir, err)
 	}
+	haABundle := filepath.Join(exportDir, hostBundleName("ha-a"))
+	bundle, err := readHostBundle(haABundle)
+	if err != nil {
+		return err
+	}
+	fingerprint, err := serviceCAFingerprint(bundle.Secrets["service-ca.crt"])
+	if err != nil {
+		return err
```
rl-block: Maybe these two errors could include `exportDir`, matching the peer-check and transfer errors below. At this point credential-bearing bundles already exist, so the operator otherwise has no path to clean them up.

rl-block (approved):

deployment-files/ha/fleet-compose.yaml
ankitgoswami: we don't need to worry about existing HA installs, there's only one in Markham that no one is using yet and I'll just do a fresh install.

server/internal/ha/deployment/guided_install_test.go
ankitgoswami: 🤖

> Calling `serviceCAFingerprint` to build the expected value means a digest or formatting bug would make both sides wrong in the same way.

Fixed in c107dad61. A focused test now compares a fixed certificate against a literal SHA-256 fingerprint independently generated with OpenSSL; the existing flow test remains as wiring coverage.

server/internal/ha/deployment/guided_install.go
ankitgoswami: 🤖

> Maybe these two errors could include `exportDir`, matching the peer-check and transfer errors below.

Fixed in c107dad61. Both post-generation errors now identify the retained bundle export directory so the operator can locate and remove the credential-bearing artifacts.

PR #944, opened 2026-08-19
feat(auth): add break-glass SUPER_ADMIN password reset

mcharles-square:
Reviewable diff: +996/-137 across 27 files (excludes generated, test, and story files).

## Summary

Adds a host-facing break-glass command that resets the password for the deployment's sole active organization-level `SUPER_ADMIN` without exposing Docker Compose details. The reset runs as one fail-closed database transaction: it changes the password, requires a password change at next login, revokes existing sessions, and records an auditable `cli_reset_password` activity. The same wrapper supports standalone and HA installations while keeping topology-specific Compose arguments internal.

The command refuses to mutate data when there are zero or multiple matching super admins, preserving the existing authorization model instead of introducing username-targeted privilege bypasses.

## How it works

1. An operator runs `./reset-super-admin-password.sh` from the deployment bundle and either receives a generated temporary password or supplies one over stdin.
2. The wrapper accepts only `--password-stdin` and selects exactly one installed topology. Standalone uses the persisted deployment environment and Compose project; HA requires an active-install marker tied to the current deployment, then delegates to `fleet-ha reset-password` with the generated HA environment and overlays.
3. Both paths pin the local Docker socket, reject caller database targeting, and start a dependency-free one-off `fleet-api` container.
4. `fleetd` loads only the database configuration needed by the administrative command while retaining the existing bare `fleetd` server behavior and root-level YAML compatibility.
5. Inside one database transaction, the auth service locks and verifies the sole live organization-level `SUPER_ADMIN`, writes the new bcrypt hash, requires a password change, revokes all sessions, and inserts the system activity.
6. Any topology ambiguity, cardinality, update, session-revocation, or audit failure returns non-zero without a partial reset.

## Diagrams

```mermaid
flowchart LR
    O["Host operator"] --> W["reset-super-admin-password.sh"]
    W --> D{"One active installed topology"}
    D --> S0["Persisted standalone profile"]
    D --> H["Marked HA profile via fleet-ha"]
    S0 --> C["Dependency-free fleet-api container"]
    H --> C
    C --> F["fleetd admin reset-password"]
    F --> T["Postgres transaction"]
    T --> U["Lock and update sole live SUPER_ADMIN"]
    T --> S["Revoke sessions"]
    T --> A["Insert cli_reset_password activity"]
    U --> R["Commit or roll back together"]
    S --> R
    A --> R
```

## Areas of the code involved

| --- | --- | --- |
| `deployment-files/` and release workflows | Added the reset wrapper, persisted standalone project identity, and installed it in release artifacts | Review the operator contract, argument allowlist, local-Docker pinning, and fail-closed topology selection |
| `server/cmd/fleetd/` | Added the `admin reset-password` command, password input modes, minimal admin config, and default-server compatibility | Review CLI safety, secret handling, and config compatibility |
| `server/cmd/fleet-ha/` and `server/internal/ha/deployment/` | Added HA reset orchestration and active-install marker lifecycle across install, update, and uninstall | Review HA target identity, legacy-update migration, and stale-state handling |
| `server/internal/domain/auth/` | Added the transactional reset policy and shared password validation | This is the primary security and atomicity boundary |
| `server/sqlc/queries/user.sql` and store interfaces | Added locked lookup, password update, and session-revocation operations | Review cardinality enforcement and transaction-scoped persistence |
| `server/internal/domain/stores/sqlstores/generated/` | Regenerated sqlc output | generated — skip |
| `server/migrations/000141_cli_reset_password_activity_label.*.sql` | Added the searchable activity label with reversible migration behavior | Review audit visibility and downgrade compatibility |
| `client/src/protoFleet/features/activity/utils/` | Added activity label, description, and lock icon | Makes the break-glass event recognizable in the activity UI |
| `docs/plans/2026-08-19-break-glass-super-admin-password-reset-tdd.md` | Updated the accepted design and rollout plan | Captures the intended operational and security constraints |

- Require exactly one active organization-level `SUPER_ADMIN`; this favors a narrow recovery path over a general username-targeted credential override.
- Treat the audit insert as mandatory within the same transaction; an unavailable audit path prevents recovery instead of allowing an unrecorded password change.
- Add the command to `fleetd` while preserving bare invocation as the server default; this reuses the shipped image without adding another binary.
- Persist the validated standalone Compose project name in `.env`; recovery targets the original stack even when installation used a process-level override.
- Allow only `--password-stdin` through the wrapper and reject caller project, database, and Docker targets before Docker access.
- Route HA recovery through `fleet-ha` only when a root-owned marker identifies the current deployment; non-purge uninstall removes the marker, updates backfill it, stale retained data is ignored, and mixed topology state fails closed.
- Enforce the same 8-character minimum and bcrypt-compatible 72-byte maximum in onboarding, authenticated changes, and recovery; stdin-supplied credentials are never echoed.
- Omit a direct SQL fallback; this keeps password hashing, session revocation, and audit semantics on one maintained path.

## Testing & validation

- `just gen` completed without generated-file changes.
- `just lint` in `server/` passed with zero issues.
- Focused `fleetd`, `fleet-ha`, and HA deployment Go suites passed, including legacy bare-flag and environment parsing through the production CLI.
- Wrapper tests passed for standalone and HA routing, argument rejection before Docker, local socket and project pinning, no-dependency execution, stdin forwarding, active-marker matching, stale HA state, and ambiguous topology.
- The full deployment-runner upgrade/safety suite passed, including project-name persistence and fail-closed rejection of transient database targets without logging their values.
- Affected auth/onboarding suites and real-Postgres break-glass integration tests passed, including rollback when session revocation fails.
- Targeted client Vitest (27 tests) and ESLint passed.
- Pre-commit formatting/import checks and pre-push server lint passed.

The database integration coverage exercises password hashing, forced-change state, session revocation, audit insertion, concurrency/cardinality failures, and migration up/down behavior. A manual reset followed by an interactive browser login was not run.


server/cmd/fleetd/main.go
```diff
@@ -143,24 +142,40 @@ const shutdownTimeout = 10 * time.Second
...
-
-	_ = kong.Parse(
-		config,
+	cli := &fleetdCLI{}
+	parser := kong.Must(
+		cli,
 		kong.Name("fleetd"),
-		kong.Configuration(kongyaml.Loader, "/etc/fleetd/config.yaml"),
+		kong.Configuration(fleetdYAMLLoader, "/etc/fleetd/config.yaml"),
+		kong.BindTo(context.Background(), (*context.Context)(nil)),
```
rl-block, line 151: Worth having the admin reset derive a bounded context before database access instead of inheriting `context.Background()` here? A blocked connection or `FOR UPDATE` can probably hang recovery indefinitely. Perhaps we could add an operation timeout and a blocked-query test.

server/cmd/fleetd/commands.go
```diff
@@ -0,0 +1,140 @@
...
+		sqlstores.NewSQLTransactor(conn),
+		sessionService,
+		activityService,
+	)
+	result, err := resetService.ResetSuperAdminPassword(ctx, password)
+	if err != nil {
+		return err
+	}
+
+	return cmd.writeResult(runtime.Stdout, result.Username)
```
rl-block: Should this success-message write be best-effort after the transaction commits? If stdout or the attach stream fails here, fleetd exits nonzero and the host wrappers withhold the generated password even though the reset completed. It might be worth retaining the credential securely until delivery succeeds and covering a failing writer.

deployment-files/reset-super-admin-password.sh
```diff
@@ -0,0 +1,136 @@
...
+
+cd "$DEPLOYMENT_DIR"
+docker_command=(
+    docker compose
+    --project-name "$FLEET_COMPOSE_PROJECT_NAME"
+    --project-directory "$DEPLOYMENT_DIR"
+    --env-file "$ENV_FILE"
+    -f "$COMPOSE_FILE"
+    run --rm --no-deps -T fleet-api
+    /app/fleetd admin reset-password --password-stdin
```
rl-block: Worth capturing the effective Docker endpoint or context once and use it for both the daemon check and this Compose call? The mutable current context can change between the two, so verification may cover daemon A while recovery runs on daemon B. A context-switch test might help pin this down.

server/sqlc/queries/user.sql
```diff
@@ -89,3 +89,37 @@ SET
...
+    AND r.organization_id = uor.organization_id
+JOIN "user" AS u ON u.id = uor.user_id
+WHERE uor.scope_type = 'org'
+    AND uor.scope_id IS NULL
+    AND uor.deleted_at IS NULL
+    AND r.deleted_at IS NULL
+    AND r.builtin_key = 'SUPER_ADMIN'
+    AND u.deleted_at IS NULL
+ORDER BY u.id
+FOR UPDATE OF uor, r, u;
```
rl-block: Perhaps this query could also join and lock the live `user_organization` membership? Session authentication later requires `uo.deleted_at IS NULL`, so this can reset an account that still cannot sign in. Maybe we could add the join and a soft-deleted-membership integration case before regenerating sqlc.

rl-block (approved):

server/cmd/fleetd/commands.go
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Claude Code — the post-commit write is now best-effort (`reportResetResult`): a stdout failure logs a warning to stderr and exits 0, so the wrappers keep delivering the already-active temporary password. Covered by a failing-writer unit test. (fb95d24f7)

server/cmd/fleetd/main.go
mcharles-square: 🤖 Updated by Claude Code — the admin path now bounds its database phase with a 60s context and pings with `InitialConnectionTimeout` (2s default) before use, so an unreachable database or held lock fails recovery fast instead of hanging. Added `TestBreakGlassResetHonorsContextDeadlineWhileLockHeld`, which holds a FOR UPDATE lock and verifies the reset returns `context.DeadlineExceeded` and rolls back. (fb95d24f7)

server/cmd/fleetd/commands.go
mcharles-square: 🤖 Updated by Claude Code — yes: the success write is now best-effort after commit (warning to stderr, exit 0), so the host wrappers never withhold the generated password for a committed reset; the credential itself is generated and retained host-side, so exit 0 guarantees delivery. Failing-writer path is unit-tested. (fb95d24f7)

deployment-files/reset-super-admin-password.sh
mcharles-square: 🤖 Updated by Claude Code — the wrapper now pins `DOCKER_CONTEXT` (the context name, so any TLS material the context carries is preserved) for the whole run when neither `DOCKER_HOST` nor `DOCKER_CONTEXT` is set, so the daemon-identity check and the Compose call can no longer observe different daemons after a concurrent `docker context use`. Regression assertion added to the wrapper test suite. (fb95d24f7)

server/sqlc/queries/user.sql
mcharles-square: 🤖 Updated by Claude Code — good catch: sign-in role resolution requires a live membership (`uo.deleted_at IS NULL`), so the old query could reset an account that still couldn't log in. `LockActiveSuperAdminUsers` now joins and locks the live `user_organization` row (sqlc regenerated), and `TestBreakGlassResetRejectsSoftDeletedOrganizationMembership` verifies the reset fails explicitly with the invariant error and leaves the password unchanged. (fb95d24f7)

server/cmd/fleet-ha/main.go
```diff
@@ -89,6 +92,46 @@ type composeCmd struct {
...
+
+	password, err := generatePassword()
+	if err != nil {
+		return fmt.Errorf("generate temporary password: %w", err)
+	}
+	if err := reset(ctx, strings.NewReader(password+"\n")); err != nil {
+		return err
+	}
+	if _, err := fmt.Fprintf(stdout, "Temporary password: %s\n", password); err != nil {
+		return fmt.Errorf("write temporary password: %w", err)
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Claude Code — partially agreed. The observation is right: the error after a committed reset didn't say the password had already changed, and it now does ("the reset already committed but the temporary password could not be delivered: …; rerun recovery to issue a fresh temporary password"), with a failing-writer test. (62e4e52c2)

Pushing back on the parallel to the fleetd fix, though: this write should NOT become best-effort. In fleetd the failing write was informational — the credential holder (wrapper/fleet-ha) still had the password and needed exit 0 to release it. Here the failing write IS the credential delivery: exiting 0 would report a successful recovery while the operator never received the now-active password. The nonzero exit stays.

server/internal/domain/stores/interfaces/mocks/mock_user_store.go
```diff
@@ -431,3 +432,72 @@ func (mr *MockUserManagementStoreMockRecorder) UpdateUserPasswordAndClearPasswor
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "UpdateUserPasswordAndClearPasswordChangeFlag", reflect.TypeOf((*MockUserManagementStore)(nil).UpdateUserPasswordAndClearPasswordChangeFlag), ctx, userID, passwordHash)
 }
+
+// MockBreakGlassUserStore is a mock of BreakGlassUserStore interface.
+type MockBreakGlassUserStore struct {
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Updated by Claude Code — pushing back on this one: mockgen's `-source` mode generates mocks for every interface in the source file and ignores the trailing interface list (those positional args only apply to reflect/package mode). Verified empirically: running `go generate ./internal/domain/stores/interfaces/` on this head produces a zero diff, and the Generated Code Check job is green on this PR, so the committed output is reproducible as-is. No change needed.

PR #946, opened 2026-08-19
feat(fleetnode): support HA control stream failover

ankitgoswami:
Reviewable diff: +224/-44 across 8 files (excludes generated, test, and story files).

## Summary

Fleet Nodes can now remain usable through Proto Fleet application-host failover while keeping the existing VIP, certificate, port 443, and `/api-proxy` URL. The authenticated ControlStream reconnects quickly after demotion without waiting for commands from the old session, while every other Fleet Node RPC stays on the Connect protocol. This extends the existing active/passive HA boundary to Fleet Node command delivery without adding a supervisor, persisted coordination state, operator command, or qualification automation.

## How it works

1. The Fleet Node sends ordinary authenticated RPCs through the existing Connect client. Its authenticated ControlStream alone uses native gRPC over the same HTTP client and public URL.
2. The client keeps HTTP/2 connections observable with a PING after 30 idle seconds and a 10-second response timeout. The settings apply to both HTTPS and h2c transports.
3. NGINX enables HTTP/2 and selects an exact ControlStream location. That location strips `/api-proxy`, forwards the request to the existing h2c Fleet server, disables upstream retry, and removes request-body limits that would terminate a long-lived stream. The generic proxy continues to serve all other RPCs.
4. A passive host rejects stream admission with `Unavailable` plus `FLEET_ERROR_CODE_NOT_ACTIVE`. An accepted stream returns the same structured error if its active lifetime ends. Client cancellation and normal newest-stream replacement still close without being labeled as demotion.
5. The Fleet Node recognizes the structured error and retries in an independent 1-1.5 second jitter window. Other transient failures retain the existing 1-30 second exponential backoff.
6. `RunCmd` owns process-wide command permits and worker accounting. Stream loss cancels the old command contexts, closes the old acknowledgement sender, and immediately starts reconnecting. Old workers retain their permits until exit, so a replacement stream returns `BUSY` instead of exceeding process capacity. Daemon shutdown waits at most 10 seconds for those workers before existing plugin cleanup continues.

## Diagrams

```mermaid
flowchart LR
    FN["Fleet Node daemon"] -->|"ControlStream: gRPC over HTTP/2"| VIP["Existing Fleet VIP :443/api-proxy"]
    FN -->|"Other RPCs: Connect"| VIP
    VIP --> NG["NGINX on selected application host"]
    NG -->|"Exact ControlStream route: grpc_pass"| GW["Fleet Node gateway on :4000"]
    NG -->|"Generic routes: proxy_pass"| GW
    GW --> ADM["Active admission lifetime"]
    ADM -->|"Active"| REG["Control session registry"]
    ADM -->|"Passive or demoted"| NA["Unavailable plus NOT_ACTIVE detail"]
    NA --> FN
```

```mermaid
sequenceDiagram
    participant FN as Fleet Node
    participant VIP as VIP and NGINX
    participant A as Active Fleet host
    participant W as Old command worker
    participant B as Replacement Fleet host

    FN->>VIP: Open authenticated gRPC ControlStream
    VIP->>A: Forward exact RPC path
    A-->>FN: Accepted
    A->>W: Start command
    A-->>FN: NOT_ACTIVE after demotion
    FN->>FN: Cancel old session and close old sender
    FN->>VIP: Reconnect after 1-1.5 second jitter
    VIP->>B: Forward to selected host
    B-->>FN: Accepted
    B-->>FN: BUSY while old permits remain occupied
    W-->>FN: Late acknowledgement rejected by closed old sender
    B-->>FN: Command accepted after one old permit is released
```

## Areas of the code involved

| --- | --- | --- |
| `client/nginx.runner-protofleet.conf`, `deployment-files/client/nginx.*.conf` | Enabled HTTP/2 and added the exact, long-lived gRPC ControlStream route | Verify route specificity, timeout/body limits, retry behavior, and unchanged generic proxying |
| `server/internal/fleetnode/bootstrap` | Split authenticated protocol selection and configured HTTP/2 PING health checks | Verify only ControlStream uses gRPC and bearer authentication is preserved on both protocols |
| `server/internal/handlers/fleetnode/gateway` | Mapped accepted-stream demotion to the existing structured not-active error | Verify demotion remains distinct from client cancellation and stream replacement |
| `server/cmd/fleetnode/control.go` | Added structured retry classification and process-wide command capacity | Verify retry windows, permit ownership, sender isolation, and immediate reconnect ordering |
| `server/cmd/fleetnode/run.go` | Added bounded daemon-level worker drainage | Verify the 10-second shutdown boundary occurs before existing plugin cleanup |
| Fleet Node and gateway tests | Added protocol, demotion, retry, reconnect, capacity, acknowledgement, and shutdown coverage | Focus on cross-session concurrency behavior and race-detector coverage |

- Use gRPC only for authenticated ControlStream instead of moving the full Fleet Node API; this keeps existing unary and streaming RPC behavior unchanged.
- Reuse `Unavailable` plus `FLEET_ERROR_CODE_NOT_ACTIVE` instead of adding a new protobuf or endpoint; clients can distinguish HA demotion without a contract migration.
- Keep permits on `RunCmd` instead of introducing a supervisor; reconnect stays immediate while work from prior sessions remains bounded at the process level.
- Close each session's acknowledgement sender instead of redirecting workers to the replacement stream; late results cannot be attributed to a new session.
- Disable NGINX retry for the stream and let the Fleet Node reconnect loop own host reselection; this avoids replaying a bidirectional stream inside the proxy.
- Server and NGINX must deploy before the updated Fleet Node. The exact proxy route remains compatible with the previous Connect ControlStream during rollout.

## Testing & validation

- Targeted Fleet Node, bootstrap transport, and gateway demotion tests pass under the Go race detector.
- All three NGINX configurations pass `nginx -t` in the repository's pinned `nginx:1.29.0-alpine` image.
- A disposable NGINX probe confirmed that the exact `grpc_pass` route preserves the prior Connect ControlStream protocol and bearer header during rollout.
- The existing HA deployment profile checks pass unchanged.
- Hermit-backed diff-scoped server lint, protobuf lint, and `git diff --check` pass.
- The full server lint baseline still reports the unrelated current-main `goconst` issue in `server/internal/domain/alerts/deliver.go`; this PR adds no new lint findings.
- The real two-host Fleet Node failover remains the requested post-merge check. No HA qualification section, RFC, qualification document, or qualification automation changed in this PR.

- **Owner/window:** The release operator runs the Fleet Node failover check after the server/NGINX rollout and again after the Fleet Node rollout.
- **Logs:** Search for `control stream disconnected; will reconnect`, structured `NOT_ACTIVE`, repeated `BUSY`, and `control command handlers still running; continuing daemon shutdown`.
- **Operational signals:** Keep `fleet-ha status` healthy and watch the existing `fleet_ha_failover_ready` Grafana signal while failing the active application host.
- **Healthy result:** A harmless command succeeds before failover, the node reconnects through the same VIP, and a second command succeeds within the existing 180-second RTO.
- **Failure/mitigation:** Repeated not-active responses, a missing command acknowledgement, or recovery beyond 180 seconds blocks rollout. Roll back the Fleet Node package first; the server and proxy remain compatible with the previous Connect ControlStream.
- **Recordkeeping:** Keep the result in the task or release notes. Open a corrective PR only if the check exposes a defect.

---

### DIFF
--- a/deployment-files/client/nginx.http.conf
+++ b/deployment-files/client/nginx.http.conf
@@ -1,5 +1,6 @@
 server {
     listen 80;
+    http2 on;
 
     # Runtime client config, regenerated from env at container start. Never cache
     # so operator changes take effect on restart.
@@ -14,6 +15,20 @@ server {
         root /usr/share/nginx/html;
         index index.html;
         try_files $uri $uri/ /index.html;
+    }
+
+    # ControlStream is the only Fleet Node RPC that uses native gRPC. Strip the
+    # public /api-proxy prefix and keep this long-lived HTTP/2 stream pinned to
+    # the selected upstream; the node reconnect loop handles failures.
+    location = /api-proxy/fleetnodegateway.v1.FleetNodeGatewayService/ControlStream {
+        rewrite ^/api-proxy/(.*)$ /$1 break;
+        client_body_timeout 24h;
+        client_max_body_size 0;
+        grpc_pass grpc://localhost:4000;
+        grpc_set_header Host $host;
+        grpc_read_timeout 24h;
+        grpc_send_timeout 24h;
+        grpc_next_upstream off;
     }
 
     # Pair is a unary RPC that can wait for multiple fleet-node batches before
--- a/deployment-files/client/nginx.http.conf
+++ b/deployment-files/client/nginx.http.conf
@@ -26,7 +26,7 @@ server {
         client_max_body_size 0;
         grpc_pass grpc://localhost:4000;
         grpc_set_header Host $host;
-        grpc_read_timeout 24h;
+        grpc_read_timeout 90s;
         grpc_send_timeout 24h;
         grpc_next_upstream off;
     }
--- a/deployment-files/client/nginx.https.conf
+++ b/deployment-files/client/nginx.https.conf
@@ -8,6 +8,7 @@ server {
 # HTTPS server
 server {
     listen 443 ssl;
+    http2 on;
     server_name _;
 
     # SSL certificate configuration
@@ -35,6 +36,20 @@ server {
         root /usr/share/nginx/html;
         index index.html;
         try_files $uri $uri/ /index.html;
+    }
+
+    # ControlStream is the only Fleet Node RPC that uses native gRPC. Strip the
+    # public /api-proxy prefix and keep this long-lived HTTP/2 stream pinned to
+    # the selected upstream; the node reconnect loop handles failures.
+    location = /api-proxy/fleetnodegateway.v1.FleetNodeGatewayService/ControlStream {
+        rewrite ^/api-proxy/(.*)$ /$1 break;
+        client_body_timeout 24h;
+        client_max_body_size 0;
+        grpc_pass grpc://localhost:4000;
+        grpc_set_header Host $host;
+        grpc_read_timeout 24h;
+        grpc_send_timeout 24h;
+        grpc_next_upstream off;
     }
 
     # Pair is a unary RPC that can wait for multiple fleet-node batches before
--- a/deployment-files/client/nginx.https.conf
+++ b/deployment-files/client/nginx.https.conf
@@ -47,7 +47,7 @@ server {
         client_max_body_size 0;
         grpc_pass grpc://localhost:4000;
         grpc_set_header Host $host;
-        grpc_read_timeout 24h;
+        grpc_read_timeout 90s;
         grpc_send_timeout 24h;
         grpc_next_upstream off;
     }
--- a/server/cmd/fleetnode/control.go
+++ b/server/cmd/fleetnode/control.go
@@ -72,6 +72,8 @@ type acker interface {
 	Send(req *pb.ControlStreamRequest) error
 }
 
+var errControlSenderClosed = errors.New("control session sender closed")
+
 // connect-go bidi streams are not safe for concurrent Send. The receive
 // loop's busy-ack and the worker's completion ack now share a stream;
 // serialize through this wrapper.
@@ -83,12 +85,12 @@ type lockedAcker struct {
 
 func (l *lockedAcker) Send(req *pb.ControlStreamRequest) error {
 	if l.closed.Load() {
-		return io.ErrClosedPipe
+		return errControlSenderClosed
 	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	if l.closed.Load() {
-		return io.ErrClosedPipe
+		return errControlSenderClosed
 	}
 	return l.inner.Send(req)
 }
@@ -612,7 +614,7 @@ func (r *RunCmd) sendAckWithPayload(stream acker, commandID string, code pb.AckC
 		ErrorMessage: errMsg,
 		Code:         code,
 		Payload:      payload,
-	}}}); err != nil {
+	}}}); err != nil && !errors.Is(err, errControlSenderClosed) {
 		logger.Warn("send ack failed", "command_id", commandID, "err", err)
 	}
 }
--- a/server/cmd/fleetnode/control.go
+++ b/server/cmd/fleetnode/control.go
@@ -74,6 +74,11 @@ type acker interface {
 
 var errControlSenderClosed = errors.New("control session sender closed")
 
+var (
+	errControlSessionRotated = errors.New("control session credentials rotated")
+	errControlSessionExpired = errors.New("control session credentials expired")
+)
+
 // connect-go bidi streams are not safe for concurrent Send. The receive
 // loop's busy-ack and the worker's completion ack now share a stream;
 // serialize through this wrapper.
@@ -125,7 +130,7 @@ func (r *RunCmd) runControlLoop(ctx context.Context, client gatewayClient, st *b
 		default:
 		}
 		started := time.Now()
-		err := r.runControlSession(ctx, loopLogger, client)
+		err := r.runControlSession(ctx, loopLogger, client, st)
 		if err == nil {
 			return nil
 		}
@@ -185,9 +190,11 @@ func isNotActiveControlError(err error) bool {
 	return false
 }
 
-func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, client gatewayClient) error {
+func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, client gatewayClient, st *bootstrap.State) error {
 	r.initControlConcurrency()
-	stream := client.ControlStream(ctx)
+	credentialCtx, endSession := r.beginControlSession(ctx, st)
+	defer endSession()
+	stream := client.ControlStream(credentialCtx)
 	// stream.Receive parks in http2.pipe on a sync.Cond ctx can't unblock;
 	// the watcher below closes the stream so Ctrl+C returns. Defers run
 	// LIFO: close(done) fires first so the watcher exits quietly on normal
@@ -197,7 +204,7 @@ func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, cli
 	defer close(done)
 	go func() {
 		select {
-		case <-ctx.Done():
+		case <-credentialCtx.Done():
 			_ = stream.CloseRequest()
 			_ = stream.CloseResponse()
 		case <-done:
@@ -205,10 +212,22 @@ func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, cli
 	}()
 
 	if err := stream.Send(&pb.ControlStreamRequest{Kind: &pb.ControlStreamRequest_Hello{Hello: &pb.ControlHello{}}}); err != nil {
+		if ctx.Err() != nil {
+			return nil
+		}
+		if cause := context.Cause(credentialCtx); cause != nil {
+			return fmt.Errorf("control session ended while sending hello: %w", cause)
+		}
 		return fmt.Errorf("send hello: %w", err)
 	}
 	first, err := stream.Receive()
 	if err != nil {
+		if ctx.Err() != nil {
+			return nil
+		}
+		if cause := context.Cause(credentialCtx); cause != nil {
+			return fmt.Errorf("control session ended while awaiting acceptance: %w", cause)
+		}
 		return fmt.Errorf("await accepted: %w", err)
 	}
 	if first.GetAccepted() == nil {
@@ -219,7 +238,7 @@ func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, cli
 	// sessionCtx so a dropped stream cancels the in-flight scan immediately;
 	// without it the agent would burn up to commandTimeout finishing an old
 	// scan before opening a new session.
-	sessionCtx, cancelSession := context.WithCancel(ctx)
+	sessionCtx, cancelSession := context.WithCancel(credentialCtx)
 
 	// Serialize all sends on the bidi: the worker's completion ack and the
 	// receive loop's busy ack would otherwise race on stream.Send.
@@ -230,9 +249,9 @@ func (r *RunCmd) runControlSession(ctx context.Context, logger *slog.Logger, cli
 	}()
 
 	// Two process-wide lanes owned by RunCmd:
-	//   - discovery (and the pairing effort's future pair) is a heavy, report-bearing
-	//     scan; it is single-flight per node via an exclusive slot, so a second
-	//     concurrent discovery is rejected BUSY rather than doubling the scan load.
+	//   - discovery and pairing are heavy, report-bearing scans that share an
+	//     exclusive slot, so a second concurrent scan is rejected BUSY rather than
+	//     doubling the load.
 	//   - quick per-miner commands use a broader pool, so they run concurrently and a
 	//     long discovery never head-of-line-blocks them.
 	// Both are non-blocking acquires: parking the receive loop would hide stream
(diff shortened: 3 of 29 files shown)


server/internal/handlers/fleetnode/gateway/handler.go
```diff
@@ -727,7 +728,7 @@ func (h *Handler) ControlStream(ctx context.Context, stream *connect.BidiStream[
 	for {
 		select {
 		case <-ctx.Done():
-			return nil
+			return controlStreamContextError(ctx)
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Classify demotion before Hello as not-active.

Addressed in `d8ef0c5b4`: handshake failures now return structured `NOT_ACTIVE` whenever the admitted active lifetime has ended, including while waiting for Hello and when sending Accepted fails. Existing non-demotion error classifications remain unchanged, with focused coverage for both handshake phases.

server/cmd/fleetnode/control.go
```diff
@@ -74,16 +76,27 @@ type acker interface {
...
+	if l.closed.Load() {
+		return io.ErrClosedPipe
+	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
+	if l.closed.Load() {
+		return io.ErrClosedPipe
+	}
 	return l.inner.Send(req)
 }
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Old in-flight workers will commonly attempt to ACK after the session is torn down, which will now produce a warning log per ACK.

Addressed in `d8ef0c5b4`: intentional old-session closure now returns a private sentinel that suppresses only expected late-ACK warnings. The closed sender still drops the ACK, while genuine transport errors—including `io.ErrClosedPipe`—continue to warn.

ankitgoswami:
🤖

> Reconnects can execute overlapping destructive commands.

Not addressing in this bridge: per-target exclusion, stable idempotency keys, and exactly-once miner actions are explicitly outside the approved scope. This PR preserves process-wide capacity across reconnects and cancels old command contexts, but it intentionally does not add per-device ownership or deduplication state.

> Demotion during stream handshake bypasses fast failover handling.

Addressed in `d8ef0c5b4`: handshake failures now map an ended admitted active lifetime to structured `NOT_ACTIVE`, including while waiting for Hello and while sending Accepted. Focused race tests cover both phases.

server/internal/handlers/fleetnode/gateway/handler.go
```diff
@@ -727,7 +726,7 @@ func (h *Handler) ControlStream(ctx context.Context, stream *connect.BidiStream[
...
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> The receive/send error exits should apply the same active-lifetime classification as this context branch.

Addressed in `f2225ab6a`: active-lifetime classification is now centralized across the bounded `ControlStream` exit class, including handshake errors, invalid first messages, accepted/command send failures, receive failures, EOF, context cancellation, and registry termination. Concurrent demotion returns structured `NOT_ACTIVE`; the same exits retain their existing result while the admitted active lifetime is live.

ankitgoswami:
🤖

> [HIGH] Control stream remains authorized after session rotation or expiry

Addressed in `a7c0d7298`. Each ControlStream now has an application-level deadline tied to the session expiry captured when it opens. A successful token refresh cancels only the active stream, and the existing reconnect loop opens the replacement with the rotated bearer. Focused tests verify the replacement presents the new token and a failed refresh cannot extend the old stream past expiry.

> [HIGH] Reconnect can overlap mutating commands on the same miner

Not addressing in this PR: per-device ownership, stable logical command IDs, and exactly-once miner actions are explicitly outside the Fleet Node HA bridge scope. This PR keeps the agreed process-wide capacity and old-sender isolation behavior; device-level execution ownership needs a separate design rather than an anonymous-permit change here.

> [MEDIUM] Persistent passive-state responses disable reconnect backoff

Not addressing in this PR: repeated structured `not-active` responses intentionally remain in the 1-1.5-second initial jitter window to meet the active/passive failover requirement. Generic transient failures still retain the existing 1-30-second exponential backoff.

ankitgoswami:
🤖

> ControlStream credential expiry is enforced only by the client

Addressed in `49b093e59`. The server now propagates the persisted session expiry into Fleet Node authentication and applies it as the authenticated streaming-handler deadline. Successful credential replacement disconnects the prior registered stream only after the new session is persisted, and successful Fleet Node revocation uses the same server-side eviction path. Failed replacement or revocation preserves the current stream. Focused race tests cover expiry, replacement, revocation, and failure ordering.

> Session rotation can execute queued device commands more than once

Not addressing in this PR. Bounded draining would contradict the approved immediate-reconnect behavior, while durable outcomes, stable logical operation IDs, and Fleet Node deduplication are the explicitly excluded exactly-once design. The bridge retains process-wide capacity and old-sender isolation, but its physical device operations remain at-least-once across a lost acknowledgement.

server/internal/domain/fleetnode/control/stream.go
```diff
@@ -38,6 +38,18 @@ func (r *Registry) Register(fleetNodeID int64) *Stream {
...
 }
 
+// Disconnect tears down the active ControlStream for fleetNodeID, if any.
+func (r *Registry) Disconnect(fleetNodeID int64) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	conn, ok := r.conns[fleetNodeID]
+	if !ok {
+		return
+	}
```
rl-block, line 87: Should we account for streams that have authenticated but not yet sent Hello? `Disconnect` returns when no connection is registered, so rotation or revocation can complete during that window and the stale stream can register afterward until expiry. Maybe bind registration to an invalidation generation or pending lease, with a delayed-Hello rotation/revocation test.

deployment-files/client/nginx.http.conf
```diff
@@ -16,6 +17,20 @@ server {
...
+    # the selected upstream; the node reconnect loop handles failures.
+    location = /api-proxy/fleetnodegateway.v1.FleetNodeGatewayService/ControlStream {
+        rewrite ^/api-proxy/(.*)$ /$1 break;
+        client_body_timeout 24h;
+        client_max_body_size 0;
+        grpc_pass grpc://localhost:4000;
+        grpc_set_header Host $host;
+        grpc_read_timeout 24h;
+        grpc_send_timeout 24h;
+        grpc_next_upstream off;
```
rl-block, line 31: It might be worth checking in a proxy-level regression test for this route. The Go tests connect directly to h2c, so removing the exact match, rewrite, or `grpc_pass` could leave the suite green while production ControlStream fails. Maybe exercise native gRPC through `/api-proxy` and one Connect unary RPC.

deployment-files/client/nginx.https.conf
```diff
@@ -37,6 +38,20 @@ server {
...
+    # ControlStream is the only Fleet Node RPC that uses native gRPC. Strip the
+    # public /api-proxy prefix and keep this long-lived HTTP/2 stream pinned to
+    # the selected upstream; the node reconnect loop handles failures.
+    location = /api-proxy/fleetnodegateway.v1.FleetNodeGatewayService/ControlStream {
+        rewrite ^/api-proxy/(.*)$ /$1 break;
+        client_body_timeout 24h;
+        client_max_body_size 0;
+        grpc_pass grpc://localhost:4000;
+        grpc_set_header Host $host;
+        grpc_read_timeout 24h;
```
rl-block: Want to bound the NGINX-to-fleetd hop below the 180-second failover objective? The client HTTP/2 PING terminates at NGINX, so a stalled `localhost:4000` upstream can leave the downstream stream open for this 24-hour timeout. Perhaps add upstream or application-level liveness and tighten `grpc_read_timeout`.

server/cmd/fleetnode/control_test.go
```diff
@@ -1206,6 +1472,7 @@ func TestControlLoop_DroppedStreamCancelsInFlightScan(t *testing.T) {
 	require.Eventually(t, func() bool { return fake.helloCount() >= 2 }, 4*time.Second, 50*time.Millisecond)
 	cancel()
 	<-done
+	cmd.waitForControlWorkers(discardLogger(t))
```
rl-block, line 1516: Could this test assert that the discovery worker observes session cancellation before the parent is canceled? As written, reconnect plus the later `cancel()` can still make it pass if the dropped stream never releases the discovery slot. Maybe signal cancellation or slot release first, then add context-ignoring `BUSY` coverage.

rl-block (approved):

server/internal/domain/fleetnode/control/stream.go
ankitgoswami: 🤖

> Maybe bind registration to an invalidation generation or pending lease, with a delayed-Hello rotation/revocation test.

Addressed in `cc6ecdc8b`: authenticated streams now carry a one-way session fingerprint, and the control registry atomically fences registration against successful rotation or revocation. If registration wins first, invalidation disconnects it; if invalidation wins first, delayed `Hello` registration is rejected. Focused tests cover both rotation and revocation ordering.

deployment-files/client/nginx.http.conf
ankitgoswami: 🤖

> Removing the exact match, rewrite, or `grpc_pass` could leave the suite green while production ControlStream fails.

Addressed differently in `cc6ecdc8b`: the existing deployment-config suite now checks every shipped NGINX template for HTTP/2, the exact ControlStream location, prefix rewrite, sole `grpc_pass`, disabled retry, 90-second read timeout, and the generic Connect `proxy_pass`. This gives CI regression coverage for the production routing contract without adding a separate live-proxy harness.

deployment-files/client/nginx.https.conf
ankitgoswami: 🤖

> Perhaps add upstream or application-level liveness and tighten `grpc_read_timeout`.

Addressed in `cc6ecdc8b`: fleetd now sends the existing valid `Accepted` envelope every 30 seconds after stream admission, and all three NGINX configurations use a 90-second `grpc_read_timeout`. The Fleet Node ignores repeated `Accepted` envelopes and continues processing commands, so a silent fleetd-to-NGINX hop is closed below the 180-second objective.

server/cmd/fleetnode/control_test.go
ankitgoswami: 🤖

> Could this test assert that the discovery worker observes session cancellation before the parent is canceled?

Addressed in `cc6ecdc8b`: the blocking discoverer now signals when its session-scoped context is canceled, and the dropped-stream test requires that signal before canceling the daemon parent. The existing context-ignoring worker test continues to prove occupied process-wide permits return `BUSY` on the replacement stream.

server/internal/fleetnode/bootstrap/authclient.go
```diff
@@ -18,11 +19,34 @@ func NewAuthenticatedGatewayClient(serverURL string, tokenSource func() string)
...
 		serverURL,
-		connect.WithInterceptors(bearerInterceptor(tokenSource)),
-	), nil
+		authOption,
+	)
+	grpcClient := fleetnodegatewayv1connect.NewFleetNodeGatewayServiceClient(
+		httpClient,
+		serverURL,
+		authOption,
+		connect.WithGRPC(),
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> When a Fleet Node binary is upgraded before an HTTPS Fleet installation ... Preserve a Connect fallback or otherwise negotiate the protocol during mixed-version upgrades.

Not addressing: this PR's explicit rollout contract requires server/NGINX to deploy before the updated Fleet Node. The updated proxy remains compatible with the previous Connect ControlStream, so that supported order has mixed-version coverage. Supporting the reverse order would add a second client transport/fallback path for an excluded rollout sequence.

server/internal/handlers/fleetnode/gateway/handler.go
```diff
@@ -94,10 +100,11 @@ func (h *Handler) BeginAuthHandshake(ctx context.Context, req *connect.Request[p
 }
 
 func (h *Handler) CompleteAuthHandshake(ctx context.Context, req *connect.Request[pb.CompleteAuthHandshakeRequest]) (*connect.Response[pb.CompleteAuthHandshakeResponse], error) {
-	token, expiresAt, err := h.auth.CompleteHandshake(ctx, req.Msg.GetChallenge(), req.Msg.GetSignature())
+	token, expiresAt, fleetNodeID, err := h.auth.CompleteHandshake(ctx, req.Msg.GetChallenge(), req.Msg.GetSignature())
 	if err != nil {
 		return nil, err
 	}
+	h.registry.ReplaceSession(fleetNodeID, auth.SessionFingerprint(token))
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> When two handshakes for the same node overlap, the persisted token and in-memory fence can be applied in opposite orders.

Addressed in `7232eb074`: session persistence and fence replacement now share one handler-level critical section, preventing concurrent handshakes from applying those updates in opposite orders. A concurrent regression test verifies that the newest persisted completion owns the accepted session fence.

deployment-files/client/nginx.http.conf
```diff
@@ -1,5 +1,6 @@
 server {
     listen 80;
+    http2 on;
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Enabling HTTP/2 on this non-TLS listener makes NGINX expect h2c prior knowledge rather than ordinary HTTP/1.1.

Not addressing: this does not reproduce with the exact packaged config and pinned `nginx:1.29.0-alpine` image. An explicit `curl --http1.1` request to the cleartext listener returns `HTTP/1.1 200 OK`; `http2 on;` enables h2c while the listener continues accepting HTTP/1.1. The runner uses the same directive and behavior, so a second listener is unnecessary.

PR #949, opened 2026-08-20
alerts: rename channels & maintenance windows, updates

illegalprime:
Reviewable diff: +133/-104 across 19 files (excludes generated, test, and story files).

## Summary

This PR updates the alerting interface to call notification channels “Destinations” and maintenance windows “Quiet Periods,” including minute-precision period timestamps. Slack recovery messages now identify each alert as `Resolved: <alert name>` without a separate heading, severity, or non-identifying miner summary, while preserving the summary for device-less alerts whose source would otherwise be ambiguous. Resolved events continue to close active alert state but no longer appear as separate activity-log rows, and a partial index keeps that filtered feed efficient after large recovery batches.

## How it works

1. The alerts settings UI and user-facing validation errors use the new terminology while retaining the existing channel and maintenance-window API fields.
2. Alertmanager webhook processing continues to persist both firing and resolved events. The existing database trigger uses the resolved row to close the corresponding active alert.
3. Activity history reads filter out rows whose status is `resolved`, so the lifecycle state remains correct without showing recovery entries in the feed. A concurrent partial index matches the organization, status, and descending keyset-pagination predicates.
4. Slack rendering partitions firing and resolved alerts, groups them by alert name, and prefixes resolved group names inline. Miner-backed resolutions omit severity and summary; device-less resolutions retain identifying summaries such as a source or host name.

## Diagrams

```mermaid
flowchart LR
    operator["Operator"]
    ui["Alerts settings UI"]
    api["Existing alerts API contracts"]
    grafana["Grafana Alertmanager"]
    handler["Alert webhook handler"]
    history["Notification history"]
    active["Active alert state"]
    feed["Activity history query"]
    deliverer["Alert deliverer"]
    renderer["Slack renderer"]
    destination["Slack destination"]

    operator --> ui
    ui --> api
    grafana --> handler
    handler --> history
    history --> active
    history --> feed
    handler --> deliverer
    deliverer --> renderer
    renderer --> destination
```

```mermaid
sequenceDiagram
    participant G as Grafana Alertmanager
    participant H as Webhook handler
    participant N as Notification history
    participant A as Active alert state
    participant S as Slack destination
    participant U as Activity UI
    G->>H: Send resolved alert
    H->>N: Insert resolved row
    N->>A: Close active alert
    H->>S: Send resolved notification
    U->>N: Request activity history
    N-->>U: Return firing rows only
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/protoFleet/features/alerts/` | Renamed visible channel/window copy and formatted quiet-period timestamps without seconds | Review terminology coverage and locale-aware period display |
| `client/e2eTests/protoFleet/` | Updated alert page helpers and RBAC selectors for renamed controls | Keeps E2E automation aligned with the UI copy |
| `server/internal/domain/alerts/` | Changed Slack recovery rendering and user-facing service/routing errors; added focused regressions | Review the device-backed versus device-less resolution detail policy |
| `server/internal/handlers/alerts/` | Updated request-validation terminology | Ensures API errors shown in the UI use the same vocabulary |
| `server/sqlc/queries/notification_history.sql` | Excluded resolved rows from activity-history reads while retaining persisted rows | Review the separation between lifecycle persistence and feed presentation |
| `server/migrations/000141_notification_history_unresolved_index.*.sql` | Added a concurrent partial index for unresolved activity-history rows | Review the predicate, key order, and non-blocking rollout/rollback behavior |
| `server/generated/sqlc/` | Regenerated sqlc bindings — generated, skip | Confirms generated output matches the query source |

- Keep channel and maintenance-window names in internal/API contracts, changing only user-facing language instead of introducing a schema and wire-format rename.
- Filter resolved rows when reading activity history instead of suppressing their inserts, preserving the trigger that closes active alerts.
- Add a concurrent partial index for the filtered keyset query instead of relying on the broader organization/id index to skip potentially large resolved batches.
- Omit summary and severity for miner-backed Slack resolutions, but retain summaries for device-less rules because they can carry the only source identity.
- Use locale-aware date formatting with explicit year-through-minute fields instead of slicing formatted strings or forcing a single timezone/locale.

## Testing & validation

- `just gen` (sqlc output regenerated; only expected generated bindings changed)
- `just lint`
- `go test ./internal/domain/alerts ./internal/handlers/alerts`
- `DB_PASSWORD=fleet go test ./internal/handlers/alertmanagerwebhook`
- `DB_PASSWORD=fleet go test ./internal/domain/stores/sqlstores -run TestNotificationHistoryStore_ListExcludesResolvedRows -count=1`
- `EXPLAIN (ANALYZE, BUFFERS)` with 100 firing rows followed by 100,000 resolved rows: matching partial index-only scan, 0.023 ms execution in the synthetic plan check
- `npm test -- --run src/protoFleet/features/alerts/api/useChannelSelection.test.ts`
- Pre-commit hooks: protected-branch guard, Go imports, and client formatting
- Pre-push hooks: client typecheck and server lint
- Not run locally: ProtoFleet Playwright E2E; the touched helpers require the full docker-compose application environment.

### DIFF
--- a/client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
+++ b/client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
@@ -228,7 +228,7 @@ const AddMaintenanceWindowModal = ({
       return;
     }
     if (ruleMode === "all" && channelMode === "all" && !confirmedAllAlertingMute) {
-      setErrorMsg("Confirm that no alerts will be delivered during this window");
+      setErrorMsg("Confirm that no alerts will be delivered during this quiet period");
       return;
     }
     if (ruleMode === "selected" && liveRuleIds.size === 0) {
--- a/client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
+++ b/client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
@@ -224,7 +224,7 @@ const AddMaintenanceWindowModal = ({
       return;
     }
     if (new Date(ends).getTime() - new Date(starts).getTime() > MAX_MAINTENANCE_WINDOW_DURATION_MS) {
-      setErrorMsg("Maintenance windows cannot exceed 30 days");
+      setErrorMsg("Quiet periods cannot exceed 30 days");
       return;
     }
     if (ruleMode === "all" && channelMode === "all" && !confirmedAllAlertingMute) {
@@ -236,7 +236,7 @@ const AddMaintenanceWindowModal = ({
       return;
     }
     if (channelMode === "selected" && liveChannelIds.size === 0) {
-      setErrorMsg("Pick at least one channel, or use All channels");
+      setErrorMsg("Pick at least one destination, or use All destinations");
       return;
     }
 
@@ -252,15 +252,15 @@ const AddMaintenanceWindowModal = ({
     try {
       if (isEditing && editingMaintenanceWindow) {
         await updateMaintenanceWindow({ id: editingMaintenanceWindow.id, ...payload });
-        pushToast({ message: "Maintenance window updated", status: STATUSES.success });
+        pushToast({ message: "Quiet period updated", status: STATUSES.success });
       } else {
         await createMaintenanceWindow(payload);
-        pushToast({ message: "Maintenance window saved", status: STATUSES.success });
+        pushToast({ message: "Quiet period saved", status: STATUSES.success });
       }
       onDismiss();
     } catch (error) {
       pushToast({
-        message: getErrorMessage(error, "Failed to save maintenance window"),
+        message: getErrorMessage(error, "Failed to save quiet period"),
         status: STATUSES.error,
       });
       setSaving(false);
@@ -285,11 +285,11 @@ const AddMaintenanceWindowModal = ({
     <Modal
       open={open}
       onDismiss={onDismiss}
-      title={isEditing ? "Edit maintenance window" : "Add maintenance window"}
+      title={isEditing ? "Edit quiet period" : "Add quiet period"}
       description="Mute alert delivery during planned work. Alerts still show up in history."
       buttons={[
         {
-          text: saving ? "Saving…" : "Save maintenance window",
+          text: saving ? "Saving…" : "Save quiet period",
           onClick: () => {
             void handleSave();
           },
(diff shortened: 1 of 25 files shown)


server/sqlc/queries/notification_history.sql
```diff
@@ -106,6 +108,7 @@ LEFT JOIN device d
     AND d.deleted_at IS NULL
 LEFT JOIN discovered_device dd ON dd.id = d.discovered_device_id
 WHERE nh.organization_id = sqlc.arg('organization_id')
+  AND nh.status <> 'resolved'
   AND (sqlc.narg('before_id')::bigint IS NULL OR nh.id < sqlc.narg('before_id'))
```
rl-block, line 112: Worth using a partial index on `(organization_id, id DESC) WHERE status <> 'resolved'` here for this filter? After a fleet-wide recovery, the current index may scan many resolved rows before filling a page. Perhaps we could add a concurrent index migration and verify the plan with `EXPLAIN (ANALYZE, BUFFERS)`.

client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
```diff
@@ -224,7 +224,7 @@ const AddMaintenanceWindowModal = ({
       return;
     }
     if (new Date(ends).getTime() - new Date(starts).getTime() > MAX_MAINTENANCE_WINDOW_DURATION_MS) {
-      setErrorMsg("Maintenance windows cannot exceed 30 days");
+      setErrorMsg("Quiet periods cannot exceed 30 days");
       return;
     }
     if (ruleMode === "all" && channelMode === "all" && !confirmedAllAlertingMute) {
```
rl-block, line 230: In the validation branch immediately below, should "during this window" become "during this quiet period" to match the terminology used throughout the modal? That would avoid showing the previous term in this error path.

mcharles-square (approved):

rl-block (approved):

server/sqlc/queries/notification_history.sql
illegalprime: 🤖 Added migration 000141 with a concurrent partial index on (organization_id, id DESC) using the exact status <> resolved predicate. The fresh-database store test applies the migration successfully. A matching EXPLAIN (ANALYZE, BUFFERS) check with 100 firing rows followed by 100,000 resolved rows selected an index-only scan and completed in 0.023 ms.

client/src/protoFleet/features/alerts/components/AddMaintenanceWindowModal.tsx
illegalprime: 🤖 Updated this validation error to: “Confirm that no alerts will be delivered during this quiet period.”

PR #953, opened 2026-08-20
Persist curtailment authorization envelopes

negarn:
Reviewable diff: +430/-30 across 12 files (excludes generated, test, and story files).

## Summary

This PR gives curtailment response profiles and events a durable authorization envelope alongside their executable scope. The envelope records the site coverage authorized when the record is written so later issue #909 work can authorize lifecycle and recovery operations without reinterpreting topology selectors or widening them into site targets.

This is one implementation slice of #909. Permission enforcement from persisted envelopes, execution principals and admin-control requirements, automation bindings, closed-loop admission, and dispatch-time reauthorization remain out of scope and continue in the tracking issue.

## How it works

For explicit miner scopes, the handler carries the device-to-site snapshot used for permission checks into the domain request. Profile and event writes then enter a database transaction, lock the referenced sites, miners, and facility fans, compare their current placement with the authorized snapshot, and fail with `FailedPrecondition` if it changed.

The store parses the canonical terminal selector and derives a versioned envelope with separate selected-resource sites, current-member sites, facility-fan sites, and unbounded flags. It persists that envelope in a dedicated JSONB column while leaving `scope_json` as the executable selector. Event and profile reads hydrate both fields independently.

The migration deliberately refuses to run when pre-contract response-profile or event rows exist. This project has no legacy records to preserve, so deployment fails closed instead of inventing authorization coverage through a backfill.

## Diagrams

```mermaid
flowchart TD
  A["Curtailment or profile request"] --> B["Handler authorization snapshot"]
  B --> C["Canonical domain selector"]
  C --> D["SQL store transaction"]
  D --> E["Lock sites, miners, and facility fans"]
  E --> F["Compare authorized and current placement"]
  F --> G["Derive versioned authorization envelope"]
  G --> H["Persist selector and envelope separately"]
```

```mermaid
sequenceDiagram
  participant H as Handler
  participant D as Domain service
  participant S as SQL store
  participant DB as PostgreSQL
  H->>H: Resolve resource sites and authorize
  H->>D: Pass selector and authorized snapshots
  D->>S: Save normalized profile or event
  S->>DB: Begin transaction and lock resources
  S->>S: Compare snapshots and derive envelope
  S->>DB: Insert or update selector plus envelope
  DB-->>S: Return persisted row
  S-->>H: Return model with hydrated envelope
```

## Areas of the code involved

| --- | --- | --- |
| `server/migrations/000142_*` | Adds required profile and event authorization-envelope JSONB columns with version/shape checks and a zero-row deployment gate. | Defines the fail-closed persistence contract and rollback behavior. |
| `server/internal/domain/stores/sqlstores/curtailment_authorization_envelope.go` | Strictly parses scopes, locks explicit miners, resolves topology coverage, and builds envelope version 1. | Main authorization-boundary and race-handling logic. |
| `server/internal/domain/stores/sqlstores/curtailment.go` | Derives envelopes inside profile/event write transactions and hydrates them on reads. | Ensures selector and authorization state are committed atomically. |
| `server/internal/handlers/curtailment/` and domain request models | Carries the handler-authorized explicit-miner site snapshot through Start and profile saves. | Prevents placement changes between authorization and persistence from silently widening coverage. |
| `server/sqlc/queries/` | Adds envelope persistence/projection and a deterministic `FOR UPDATE` miner-site query. | Review the SQL locking and column ordering that support the transaction boundary. |
| `server/generated/sqlc/` | Regenerated sqlc models, queries, interfaces, and retry wrappers. | Generated — skip. |
| Server tests and fixtures | Covers envelope shape, strict scope parsing, topology/member separation, unbounded miners, and authorization races; updates required insert fixtures. | Pins the fail-closed behavior and new schema requirements. |

- Store authorization coverage in a dedicated column rather than encoding it into `scope_json`, so authorization data can never become an executable selector.
- Derive envelopes inside the write transaction and compare handler snapshots under row locks rather than trusting pre-transaction authorization results.
- Keep selected-resource, current-member, and facility-fan coverage separate, with explicit unbounded flags, instead of collapsing them into one ambiguous site list.
- Reuse the canonical strict scope parser so unknown selectors, unsupported versions, and malformed IDs fail before persistence.
- Reject deployments with existing profile/event rows instead of adding a legacy backfill whose authorization coverage could not be proven.

## Testing & validation

- `bin/just gen` — passed with no generated diff after rebasing onto current `main`.
- `cd server && ../bin/just lint` — passed with zero issues.
- `cd server && ../bin/go test ./internal/domain/curtailment/... ./internal/handlers/curtailment -count=1` — passed.
- `cd server && ../bin/go test ./internal/domain/stores/sqlstores -run '^TestBuildAuthorizationEnvelopeJSON' -count=1` — passed.
- `cd server && ../bin/go test ./internal/domain/stores/sqlstores -run '^$' -count=1` — store package compilation passed.
- Database-backed integration tests were not executable locally because the configured `fleet` Postgres credentials failed SASL authentication; CI remains responsible for those cases.

Part of #909


server/internal/domain/stores/sqlstores/curtailment_authorization_envelope.go
```diff
@@ -0,0 +1,183 @@
...
+			return nil, err
+		}
+		envelope.CurrentMemberSiteIDs = sites
+		envelope.MinerScopeUnbounded = unbounded
+	case models.ScopeTypeMixed:
+		switch {
+		case len(scope.SiteIDs) > 0:
+			envelope.SelectedResourceSiteIDs = uniqueSortedInt64s(scope.SiteIDs)
+		case len(scope.BuildingIDs) > 0, len(scope.RackIDs) > 0, len(scope.GroupIDs) > 0:
+			coverage, err := resolveCurtailmentTopologyScope(ctx, q, interfaces.ListCandidatesParams{
```
flesher: The device-list path locks rows FOR UPDATE and diffs against a snapshot; the topology branch calls resolveCurtailmentTopologyScope with neither. A device reassigned between this coverage read and the target insert produces an envelope whose sites disagree with the persisted targets, with no retry. Worth either locking the coverage read or documenting why topology accepts the race (the org-wide handler stopgap in #4 currently masks it).

server/internal/domain/stores/sqlstores/curtailment_authorization_envelope.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	rowsByIdentifier := make(map[string]sqlc.LockCurtailmentResponseProfileDeviceSitesByOrgRow, len(rows))
+	for _, row := range rows {
+		rowsByIdentifier[row.DeviceIdentifier] = row
+	}
+	sites := make([]int64, 0, len(rows))
+	unbounded := false
+	for _, identifier := range identifiers {
+		row, exists := rowsByIdentifier[identifier]
+		if expectedDeviceSites != nil {
```
flesher, line 269: The staleness check only runs when expectedDeviceSites != nil. Automation starts go through startRequestFromAutomationProfile, which never sets AuthorizedDeviceSites, so those events skip it and can record a device’s new site after it’s been reassigned. The same drift on a manual start gets rejected. Intended?

flesher (approved):

PR #954, opened 2026-08-21
perf(dbtest): clone a migrated template database per test

mcharles-square:
Reviewable diff: +529/-22 across 3 files (excludes generated, test, and story files).

## Summary

DB-backed server tests get ~3x faster and stop threatening CI: the full server suite goes from **3m25s to 1m09s** and `internal/domain/stores/sqlstores` from **159s to 33s**. The trigger was a real failure — that package hit Go's default 10-minute per-package timeout (`FAIL … 600.051s`) on a downstream sync, because the harness replayed all 142 migrations into a fresh database for each of 259 tests, spending ~501s of the 600s budget inside `migrate.Up()`. This replaces the per-test migration replay with a migrated template database that PostgreSQL clones per test, so setup cost stops scaling with the number of migrations.

Upstream `main` was already at `ok … 388.626s` for that package — 65% of the budget, growing with every migration added. This is a fix for a cliff we were already walking toward, not just for the branch that fell off it.

## How it works

`dbtest.GetTestDB` previously did, per test: `CREATE DATABASE` → replay every migration (~2s) → run the test → `DROP DATABASE`. Cost was O(tests × migrations).

Now the migration set is applied **once** per test run, into a template database, and each test gets a filesystem-level clone of it:

1. **First DB-backed test in a binary** computes a fingerprint — `sha256` over the sorted embedded migration filenames and contents, truncated to 12 hex chars — and looks for `fleet_test_tmpl_<fingerprint>`.
2. **If it is not ready, it builds it** while holding a PostgreSQL advisory lock, so the many test binaries `go test ./...` runs concurrently cooperate rather than race: one builds, the others wait and reuse. Verified on a cold database: a full parallel suite produces exactly one template.
3. **Publishing the template** means `ALTER DATABASE … ALLOW_CONNECTIONS false` plus evicting its sessions. This is load-bearing twice over: `CREATE DATABASE … TEMPLATE` fails while any session is attached to the source, and "connections disabled" is also the readiness marker — a template that exists but still accepts connections is treated as half-built and rebuilt, so an interrupted run cannot publish a partially migrated schema.
4. **Each test** then runs `CREATE DATABASE … TEMPLATE <template>` (~20–40ms instead of ~2s) and still calls `ConnectAndMigrate`, which is a no-op version check on an already-current clone. That is what keeps the change safe: if no template can be prepared, or a clone is somehow behind, the harness falls back to migrating the database itself.

The one genuinely surprising constraint is TimescaleDB. `CREATE EXTENSION timescaledb` causes the launcher to attach a background worker to the database, and that counts as a connected session — so the first implementation of this change was *slower* than the code it replaced (1801s vs 159s), with every clone waiting ~5s and then failing:

```
ERROR:  source database "fleet_test_tmpl_867fe765862f" is being accessed by other users
```

`ALLOW_CONNECTIONS false` prevents the launcher reattaching but does not evict the worker already inside, so the template's sessions are terminated after sealing, and clone attempts retry on SQLSTATE 55006 as a backstop.

## Diagrams

```mermaid
flowchart TD
    A["Test calls dbtest.GetTestDB"] --> B{"sync.Once: template resolved?"}
    B -- "no" --> C["pg_advisory_lock (pinned session)"]
    C --> D{"fleet_test_tmpl_&lt;fingerprint&gt; ready?<br/>(exists AND datallowconn = false)"}
    D -- "yes" --> H["release lock"]
    D -- "no" --> E["drop stale, CREATE DATABASE, run all migrations"]
    E --> F["ALTER DATABASE ALLOW_CONNECTIONS false"]
    F --> G["evict TimescaleDB background worker, wait for 0 sessions"]
    G --> I["drop templates from other fingerprints"]
    I --> H
    B -- "yes" --> J
    H --> J["CREATE DATABASE test_db TEMPLATE template"]
    J --> K["ConnectAndMigrate: no-op version check"]
    K --> L["run test"]
    L --> M["t.Cleanup: DROP DATABASE test_db"]
```

```mermaid
stateDiagram-v2
    [*] --> Absent
    Absent --> Building: "CREATE DATABASE (holder of advisory lock)"
    Building --> Migrated: "migrations succeed"
    Building --> Absent: "migration fails -> dropped, callers fall back"
    Migrated --> Sealed: "ALLOW_CONNECTIONS false + sessions evicted"
    Sealed --> [*]: "cloneable; reused by later processes and runs"
    note right of Migrated
        Reachable but never published.
        Still allows connections, so a
        later run treats it as half-built
        and rebuilds it.
    end note
    Sealed --> Absent: "migration set changes -> new fingerprint, this one dropped as stale"
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/testutil/dbtest/template.go` | New. Fingerprinting, advisory-locked template preparation, sealing plus session eviction, stale-template GC | The whole mechanism. Concentrate on the readiness definition and the lock scope |
| `server/internal/testutil/dbtest/dbtest.go` | `createTestDatabase` / `tryCreateTestDatabase` take a template and emit `CREATE DATABASE … TEMPLATE`; new `isRetryableCreateError` adds SQLSTATE 55006; retries evict template sessions first | Only entry point tests use; the existing transient-restart retry behaviour is preserved, not replaced |
| `server/internal/testutil/dbtest/template_test.go` | New unit tests: retry classification, fingerprint stability, 63-char identifier limit, migration-listing order | Tests — skip for architecture |
| `docs/solutions/database-issues/dbtest-template-database-timescaledb-worker-2026-08-21.md` | New solution doc | Captures the TimescaleDB-worker gotcha and the advisory-lock-on-pool footgun |

No proto, migration, client, or plugin changes; nothing in the production code path — this is test-harness only.

| Decision | Chosen over | Why |
| --- | --- | --- |
| Template database cloned per test | Bumping `-timeout` in `server/justfile` | A timeout bump hides growth that is O(migrations); this removes it. (A bump alone was the tempting one-line fix) |
| Fingerprint the migration set into the template name | A fixed template name plus a version check | Any migration edit — including editing an unshipped one on a branch — yields a new template, so a stale schema cannot be silently reused |
| `datallowconn = false` as the readiness marker | A marker table inside the template | Checking a marker table requires connecting to the template, and a connection is exactly what breaks concurrent clones |
| Advisory lock on a pinned `sql.Conn` | Lock on the `*sql.DB` pool, or a Go-level mutex | Advisory locks are session-scoped (a pooled unlock can land on a different session), and a Go mutex would not coordinate across the separate processes `go test ./...` spawns |
| Keep `ConnectAndMigrate` on each clone | Plain connect for speed | Costs a couple of queries and makes the change fail-safe: stale or unavailable template degrades to the old behaviour |
| Default `STRATEGY = WAL_LOG` | `FILE_COPY` | Only ~2x faster per clone at this template size (16MB) and forces two checkpoints, which hurts under hundreds of concurrent clones |
| Best-effort GC of other fingerprints' templates | Leave them, or drop on exit | Keeps long-lived local databases tidy without a teardown hook that a crashed run would skip |

## Testing & validation

Measured with `-count=1` on both sides, same machine, same TimescaleDB container:

| Scope | Before | After |
| --- | --- | --- |
| `internal/domain/stores/sqlstores` | 159.4s | **33.4s** (32.1s warm) |
| Full server suite (`go test ./... -skip ./generated/...`) | 3m25s | **1m09s** |
| Full suite from a cold database | — | 1m18s |
| Per-test database setup | ~2s (142 migrations) | ~20–40ms (clone) |
| Template build (once per fingerprint) | — | ~0.6–1.0s |

Also verified:

- Full server suite green before and after; `golangci-lint run ./internal/testutil/dbtest/...` reports 0 issues.
- **Cross-process locking:** full parallel suite from a cold database creates exactly **one** template.
- **Invalidation:** adding a temp `000143` migration built a new template and dropped the old one; removing it rebuilt the original. Tests passed in both states.
- **No leakage:** after full runs, the only remaining `fleet_test%` database is the template itself.
- **Unit tests** for retry classification, fingerprint stability and hex/length shape, the 63-char PostgreSQL identifier bound, and migration-listing order.

Not covered: no automated test forces the "TimescaleDB worker reattaches mid-run" path — it is covered by the 55006 retry plus eviction, exercised manually but not asserted. Timings are from one developer machine; CI runners are slower, so the absolute numbers will differ while the ratio should hold.

### DIFF
--- a/server/internal/testutil/dbtest/dbtest.go
+++ b/server/internal/testutil/dbtest/dbtest.go
@@ -254,7 +254,21 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 	t.Helper()
 
 	var lastErr error
-	for attempt := 1; attempt <= migrationMaxRetries; attempt++ {
+
+	// Two independent budgets. Retries hope a transient failure passes, so they
+	// are few. Recoveries *change the situation* — rebuild a swept template, give
+	// up on cloning, wait out somebody's build — and each one deserves a fresh
+	// attempt: charging them to the retry counter meant a recovery on the final
+	// attempt fell out of the loop and failed the test with the stale error,
+	// having never tried the configuration it just repaired.
+	attempts := 0
+	recoveries := 0
+
+	// One deadline for every rebuild wait belonging to this database, so repeated
+	// waits cannot accumulate into a stall that outlives the package timeout.
+	rebuildWaitDeadline := time.Now().Add(templateRebuildWaitTimeout)
+
+	for {
 		// Wait for the server before issuing admin DDL.
 		waitForServerReady(t.Context(), t, adminConfig)
 
@@ -263,50 +277,102 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 			return template
 		}
 
-		// The template vanished mid-run: another checkout on a different
-		// migration set swept it as stale. Rebuild it and try again rather than
-		// failing a test for someone else's cleanup.
-		if isMissingTemplateError(lastErr.Error(), template) {
-			t.Logf("template database %s disappeared, rebuilding: %v", template, lastErr)
-			invalidateTemplateDatabase(template)
-			template = templateDatabase(t, adminConfig)
-			continue
+		if recoveries < templateRecoveryMaxAttempts {
+			recovered, updatedTemplate := recoverTemplateForClone(
+				t, adminConfig, template, lastErr, rebuildWaitDeadline)
+			if recovered {
+				recoveries++
+				template = updatedTemplate
+				continue
+			}
 		}
 
-		// We are not allowed to copy this template — it belongs to another role
-		// and we are not a superuser. No amount of retrying changes that, so stop
-		// cloning for the rest of the process and migrate directly.
-		if isTemplatePermissionError(lastErr.Error(), template) {
-			t.Logf("not permitted to copy template database %s (%v); "+
-				"migrating test databases directly for the rest of this process", template, lastErr)
-			disableTemplateCloning()
-			template = ""
-			continue
-		}
-
-		if !isRetryableCreateError(lastErr.Error()) || attempt == migrationMaxRetries {
+		attempts++
+		if !isRetryableCreateError(lastErr.Error()) || attempts >= migrationMaxRetries {
 			break
 		}
 
-		t.Logf("create test database failed transiently (attempt %d/%d), retrying: %v", attempt, migrationMaxRetries, lastErr)
+		t.Logf("create test database failed transiently (attempt %d/%d), retrying: %v",
+			attempts, migrationMaxRetries, lastErr)
 
-		// A clone is blocked by whatever is attached to the template. If that is
-		// a rebuild (the template is not sealed), wait for it to publish: killing
-		// it would abort legitimate work, and the retry budget here is far shorter
-		// than a migration run. Otherwise the session is a stray — a TimescaleDB
-		// worker that slipped in — and gets evicted.
-		if templateRebuildInProgress(t.Context(), adminConfig, template) {
-			t.Logf("template %s is being rebuilt by another process, waiting for it", template)
-			waitForTemplateRebuild(t.Context(), adminConfig, template)
-			continue
-		}
-
+		// Whatever is attached to the template is not a live build (that is a
+		// recovery case handled above), so it is a stray — a TimescaleDB worker
+		// that slipped in — and gets evicted.
 		evictTemplateSessions(t.Context(), adminConfig, template)
-		time.Sleep(time.Duration(attempt) * migrationRetryBaseDelay)
+		time.Sleep(time.Duration(attempts) * migrationRetryBaseDelay)
 	}
 
 	assert.NoError(t, lastErr, "error creating test database")
 	return template
+}
+
+// recoverTemplateForClone handles the clone failures that a changed template can
+// fix. It reports whether it did something, along with the template the next
+// attempt should use ("" meaning migrate directly).
+func recoverTemplateForClone(
+	t *testing.T,
+	adminConfig *db.Config,
+	template string,
+	createErr error,
+	rebuildWaitDeadline time.Time,
+) (bool, string) {
+	t.Helper()
+
+	msg := createErr.Error()
+
+	// The template vanished mid-run: another checkout on a different migration
+	// set swept it as stale. Rebuild it rather than failing a test for someone
+	// else's cleanup.
+	if isMissingTemplateError(msg, template) {
+		t.Logf("template database %s disappeared, rebuilding: %v", template, createErr)
+		invalidateTemplateDatabase(template)
+		return true, templateDatabase(t, adminConfig)
+	}
+
+	// We are not allowed to copy this template — it belongs to another role and
+	// we are not a superuser. No amount of retrying changes that, so stop cloning
+	// for the rest of the process and migrate directly.
+	if isTemplatePermissionError(msg, template) {
+		t.Logf("not permitted to copy template database %s (%v); "+
+			"migrating test databases directly for the rest of this process", template, createErr)
+		disableTemplateCloning()
+		return true, ""
+	}
+
+	if !isRetryableCreateError(msg) || template == "" {
+		return false, template
+	}
+
+	switch inspectTemplateBuildState(t.Context(), adminConfig, template) {
+	case templateStateBuilding:
+		// Somebody holds the preparation lock and is migrating right now. Wait for
+		// them to publish: killing the build would abort legitimate work, and the
+		// retry budget is far shorter than a migration run.
+		if remaining := time.Until(rebuildWaitDeadline); remaining > 0 {
+			t.Logf("template %s is being rebuilt by another process, waiting up to %s", template, remaining)
+			waitForTemplateRebuild(t.Context(), adminConfig, template, rebuildWaitDeadline)
+			return true, template
+		}
+
+		// Waited as long as we are prepared to. Stop waiting and rebuild.
+		t.Logf("template %s still unpublished after %s, rebuilding it", template, templateRebuildWaitTimeout)
+		invalidateTemplateDatabase(template)
+		return true, templateDatabase(t, adminConfig)
+
+	case templateStateAbandoned:
+		// Unsealed with nobody building it: a build died and left a partially
+		// migrated database that nothing will ever finish or clean up. Its stray
+		// session is not evictable either (eviction spares unsealed templates), so
+		// rebuilding is the only way forward — dropping it disconnects the session.
+		t.Logf("template %s was left unpublished by an interrupted build, rebuilding it", template)
+		invalidateTemplateDatabase(template)
+		return true, templateDatabase(t, adminConfig)
+
+	case templateStatePublished:
+		return false, template
+	}
+
+	return false, template
 }
 
 // dropTestDatabase drops the test database at cleanup, waiting out and
--- a/server/internal/testutil/dbtest/dbtest.go
+++ b/server/internal/testutil/dbtest/dbtest.go
@@ -276,8 +276,18 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 		}
 
 		t.Logf("create test database failed transiently (attempt %d/%d), retrying: %v", attempt, migrationMaxRetries, lastErr)
-		// A clone can only be blocked by a session that reached the template, so
-		// clear the way before retrying.
+
+		// A clone is blocked by whatever is attached to the template. If that is
+		// a rebuild (the template is not sealed), wait for it to publish: killing
+		// it would abort legitimate work, and the retry budget here is far shorter
+		// than a migration run. Otherwise the session is a stray — a TimescaleDB
+		// worker that slipped in — and gets evicted.
+		if templateRebuildInProgress(t.Context(), adminConfig, template) {
+			t.Logf("template %s is being rebuilt by another process, waiting for it", template)
+			waitForTemplateRebuild(t.Context(), adminConfig, template)
+			continue
+		}
+
 		evictTemplateSessions(t.Context(), adminConfig, template)
 		time.Sleep(time.Duration(attempt) * migrationRetryBaseDelay)
 	}
--- a/server/internal/testutil/dbtest/dbtest.go
+++ b/server/internal/testutil/dbtest/dbtest.go
@@ -85,6 +85,8 @@ const (
 	pgInternalError                   = "XX000"
 	pgObjectInUse                     = "55006"
 	pgUndefinedDatabase               = "3D000"
+	pgInsufficientPrivilege           = "42501"
+	templateCopyDenied                = "permission denied to copy database"
 	timescaleTupleConcurrentlyDeleted = "tuple concurrently deleted"
 	templateSourceInUse               = "is being accessed by other users"
 
@@ -271,6 +273,17 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 			continue
 		}
 
+		// We are not allowed to copy this template — it belongs to another role
+		// and we are not a superuser. No amount of retrying changes that, so stop
+		// cloning for the rest of the process and migrate directly.
+		if isTemplatePermissionError(lastErr.Error(), template) {
+			t.Logf("not permitted to copy template database %s (%v); "+
+				"migrating test databases directly for the rest of this process", template, lastErr)
+			disableTemplateCloning()
+			template = ""
+			continue
+		}
+
 		if !isRetryableCreateError(lastErr.Error()) || attempt == migrationMaxRetries {
 			break
 		}
@@ -390,6 +403,17 @@ func isRetryableCreateError(msg string) bool {
 // database we are creating cannot itself be the undefined one.
 func isMissingTemplateError(msg string, template string) bool {
 	return template != "" && strings.Contains(msg, pgUndefinedDatabase)
+}
+
+// isTemplatePermissionError reports whether a clone failed because the
+// connecting role may not copy the template (SQLSTATE 42501). PostgreSQL only
+// permits copying a database you own unless you are a superuser, so this is what
+// a foreign template looks like.
+func isTemplatePermissionError(msg string, template string) bool {
+	if template == "" {
+		return false
+	}
+	return strings.Contains(msg, pgInsufficientPrivilege) || strings.Contains(msg, templateCopyDenied)
 }
 
 // generateTestDBName creates a unique database name that includes part of the test name for readability.
--- a/server/internal/testutil/dbtest/dbtest.go
+++ b/server/internal/testutil/dbtest/dbtest.go
@@ -1,7 +1,8 @@
 // Package dbtest provides the per-test database harness: it creates a unique
-// database per test, migrates it, and drops it at cleanup, retrying each step
-// across transient server restarts (crash-recovery under heavy concurrent
-// migration load). It lives in a leaf package — importing only
+// database per test by cloning a migrated template database (see template.go),
+// and drops it at cleanup, retrying each step across transient server restarts
+// (crash-recovery under heavy concurrent migration load). It lives in a leaf
+// package — importing only
 // infrastructure/db — so tests in packages that testutil itself imports
 // (e.g. domain/command) can use the same hardened harness without an import
 // cycle. Most tests should keep using testutil.GetTestDB, which delegates here.
@@ -47,16 +48,23 @@ func GetTestDB(t *testing.T) *sql.DB {
 	// admin DDL before reaching the migration retry below.
 	adminConfig := config
 	adminConfig.Name = "postgres"
-	createTestDatabase(t, &adminConfig, dbName)
 
-	// Connect and run migrations with retry. TimescaleDB continuous aggregate
-	// DDL acquires instance-level catalog locks that can deadlock when parallel
-	// tests migrate concurrently; a transient server restart is also tolerated.
-	// On failure we drop and recreate the database for a clean slate (avoids
-	// golang-migrate dirty flag issues).
+	// Clone the migrated template so this test skips the migration replay
+	// entirely. Falls back to an empty database (migrated below) when no
+	// template could be prepared.
+	template := templateDatabase(t, &adminConfig)
+	createTestDatabase(t, &adminConfig, dbName, template)
+
+	// Connect and run migrations with retry. Cloned databases are already at the
+	// latest version, so this is a no-op version check that also self-heals a
+	// stale clone; without a template it runs the full migration set. TimescaleDB
+	// continuous aggregate DDL acquires instance-level catalog locks that can
+	// deadlock when parallel tests migrate concurrently; a transient server
+	// restart is also tolerated. On failure we drop and recreate the database for
+	// a clean slate (avoids golang-migrate dirty flag issues).
 	testDBConfig := config
 	testDBConfig.Name = dbName
-	conn, err := connectAndMigrateWithRetry(t, &testDBConfig, &adminConfig, dbName)
+	conn, err := connectAndMigrateWithRetry(t, &testDBConfig, &adminConfig, dbName, template)
 	assert.NoError(t, err)
 
 	// Clean up the database when the test is done
@@ -76,7 +84,9 @@ const (
 	migrationRetryBaseDelay = 200 * time.Millisecond
 
 	pgInternalError                   = "XX000"
+	pgObjectInUse                     = "55006"
 	timescaleTupleConcurrentlyDeleted = "tuple concurrently deleted"
+	templateSourceInUse               = "is being accessed by other users"
 
 	// serverReadyTimeout bounds how long we wait for the database server to
 	// start accepting connections again after a transient restart before a
@@ -120,6 +130,7 @@ func connectAndMigrateWithRetry(
 	testDBConfig *db.Config,
 	adminConfig *db.Config,
 	dbName string,
+	template string,
 ) (*sql.DB, error) {
 	t.Helper()
 
@@ -138,7 +149,7 @@ func connectAndMigrateWithRetry(
 		t.Logf("retryable migration error (attempt %d/%d), retrying: %v", attempt, migrationMaxRetries, lastErr)
 		// Recreate the database for a clean slate (clears any dirty migration
 		// state), waiting out / retrying a transient server restart.
-		createTestDatabase(t, adminConfig, dbName)
+		createTestDatabase(t, adminConfig, dbName, template)
 
 		delay := time.Duration(attempt) * migrationRetryBaseDelay
 		time.Sleep(delay)
@@ -226,12 +237,13 @@ func pingAdminDatabase(ctx context.Context, adminConfig *db.Config) error {
 }
 
 // createTestDatabase drops any existing database with the given name and
-// creates a fresh one. It waits for the server to accept connections and
-// retries the admin DDL across a transient server restart (the same
-// 57P03 / "bad connection" / startup window connectAndMigrateWithRetry
-// tolerates), so a test that starts while the server is recovering survives the
-// blip instead of failing on the very first DROP/CREATE.
-func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string) {
+// creates a fresh one, cloning template when non-empty. It waits for the server
+// to accept connections and retries the admin DDL across a transient server
+// restart (the same 57P03 / "bad connection" / startup window
+// connectAndMigrateWithRetry tolerates), so a test that starts while the server
+// is recovering survives the blip instead of failing on the very first
+// DROP/CREATE.
+func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, template string) {
 	t.Helper()
 
 	var lastErr error
@@ -239,15 +251,18 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string) {
 		// Wait for the server before issuing admin DDL.
 		waitForServerReady(t.Context(), t, adminConfig)
 
-		lastErr = tryCreateTestDatabase(t.Context(), adminConfig, dbName)
+		lastErr = tryCreateTestDatabase(t.Context(), adminConfig, dbName, template)
 		if lastErr == nil {
 			return
 		}
-		if !isTransientServerError(lastErr.Error()) || attempt == migrationMaxRetries {
+		if !isRetryableCreateError(lastErr.Error()) || attempt == migrationMaxRetries {
 			break
 		}
 
 		t.Logf("create test database failed transiently (attempt %d/%d), retrying: %v", attempt, migrationMaxRetries, lastErr)
+		// A clone can only be blocked by a session that reached the template, so
+		// clear the way before retrying.
+		evictTemplateSessions(t.Context(), adminConfig, template)
 		time.Sleep(time.Duration(attempt) * migrationRetryBaseDelay)
 	}
 
@@ -309,8 +324,9 @@ func tryDropTestDatabase(ctx context.Context, adminConfig *db.Config, dbName str
 // tryCreateTestDatabase drops any existing database with the given name (after
 // terminating lingering connections) and creates a fresh one, returning any
 // error rather than failing the test so the caller can retry transient
-// failures.
-func tryCreateTestDatabase(ctx context.Context, adminConfig *db.Config, dbName string) error {
+// failures. When template is non-empty the new database is a clone of it, which
+// skips the migration replay.
+func tryCreateTestDatabase(ctx context.Context, adminConfig *db.Config, dbName string, template string) error {
 	if err := tryDropTestDatabase(ctx, adminConfig, dbName); err != nil {
 		return err
 	}
@@ -321,10 +337,24 @@ func tryCreateTestDatabase(ctx context.Context, adminConfig *db.Config, dbName s
 	}
 	defer conn.Close()
 
-	if _, err := conn.ExecContext(ctx, fmt.Sprintf("CREATE DATABASE %s", dbName)); err != nil {
+	create := fmt.Sprintf("CREATE DATABASE %s", dbName)
+	if template != "" {
+		create += fmt.Sprintf(" TEMPLATE %s", template)
+	}
+	if _, err := conn.ExecContext(ctx, create); err != nil {
 		return fmt.Errorf("create test database: %w", err)
 	}
 	return nil
+}
+
+// isRetryableCreateError extends the transient-server check with the contention
+// PostgreSQL reports when a database is cloned while another session touches the
+// source (SQLSTATE 55006). Parallel tests all clone the same template, so this
+// is expected under load rather than a real failure.
+func isRetryableCreateError(msg string) bool {
+	return isTransientServerError(msg) ||
+		strings.Contains(msg, pgObjectInUse) ||
+		strings.Contains(msg, templateSourceInUse)
 }
 
 // generateTestDBName creates a unique database name that includes part of the test name for readability.
--- a/server/internal/testutil/dbtest/dbtest.go
+++ b/server/internal/testutil/dbtest/dbtest.go
@@ -52,8 +52,7 @@ func GetTestDB(t *testing.T) *sql.DB {
 	// Clone the migrated template so this test skips the migration replay
 	// entirely. Falls back to an empty database (migrated below) when no
 	// template could be prepared.
-	template := templateDatabase(t, &adminConfig)
-	createTestDatabase(t, &adminConfig, dbName, template)
+	template := createTestDatabase(t, &adminConfig, dbName, templateDatabase(t, &adminConfig))
 
 	// Connect and run migrations with retry. Cloned databases are already at the
 	// latest version, so this is a no-op version check that also self-heals a
@@ -85,6 +84,7 @@ const (
 
 	pgInternalError                   = "XX000"
 	pgObjectInUse                     = "55006"
+	pgUndefinedDatabase               = "3D000"
 	timescaleTupleConcurrentlyDeleted = "tuple concurrently deleted"
 	templateSourceInUse               = "is being accessed by other users"
 
@@ -149,7 +149,7 @@ func connectAndMigrateWithRetry(
 		t.Logf("retryable migration error (attempt %d/%d), retrying: %v", attempt, migrationMaxRetries, lastErr)
 		// Recreate the database for a clean slate (clears any dirty migration
 		// state), waiting out / retrying a transient server restart.
-		createTestDatabase(t, adminConfig, dbName, template)
+		template = createTestDatabase(t, adminConfig, dbName, template)
 
 		delay := time.Duration(attempt) * migrationRetryBaseDelay
 		time.Sleep(delay)
@@ -243,7 +243,12 @@ func pingAdminDatabase(ctx context.Context, adminConfig *db.Config) error {
 // connectAndMigrateWithRetry tolerates), so a test that starts while the server
 // is recovering survives the blip instead of failing on the very first
 // DROP/CREATE.
-func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, template string) {
+//
+// It returns the template actually used, which can differ from the one passed
+// in: a template swept by a concurrent run on a different migration set is
+// rebuilt (or given up on, yielding "") so the caller migrates the database
+// itself.
+func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, template string) string {
 	t.Helper()
 
 	var lastErr error
@@ -253,8 +258,19 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 
 		lastErr = tryCreateTestDatabase(t.Context(), adminConfig, dbName, template)
 		if lastErr == nil {
-			return
+			return template
 		}
+
+		// The template vanished mid-run: another checkout on a different
+		// migration set swept it as stale. Rebuild it and try again rather than
+		// failing a test for someone else's cleanup.
+		if isMissingTemplateError(lastErr.Error(), template) {
+			t.Logf("template database %s disappeared, rebuilding: %v", template, lastErr)
+			invalidateTemplateDatabase(template)
+			template = templateDatabase(t, adminConfig)
+			continue
+		}
+
 		if !isRetryableCreateError(lastErr.Error()) || attempt == migrationMaxRetries {
 			break
 		}
@@ -267,6 +283,7 @@ func createTestDatabase(t *testing.T, adminConfig *db.Config, dbName string, tem
 	}
 
 	assert.NoError(t, lastErr, "error creating test database")
+	return template
 }
 
 // dropTestDatabase drops the test database at cleanup, waiting out and
@@ -307,15 +324,16 @@ func tryDropTestDatabase(ctx context.Context, adminConfig *db.Config, dbName str
 	defer conn.Close()
 
 	// Best effort: lingering connections just make the DROP fail, which the
-	// caller retries.
-	_, _ = conn.ExecContext(ctx, fmt.Sprintf(`
+	// caller retries. The name is bound as a parameter here; the DDL below has
+	// to interpolate it, so it goes through quoteIdentifier.
+	_, _ = conn.ExecContext(ctx, `
 		SELECT pg_terminate_backend(pg_stat_activity.pid)
 		FROM pg_stat_activity
-		WHERE pg_stat_activity.datname = '%s'
+		WHERE pg_stat_activity.datname = $1
 		AND pid <> pg_backend_pid()
-	`, dbName))
+	`, dbName)
 
-	if _, err := conn.ExecContext(ctx, fmt.Sprintf("DROP DATABASE IF EXISTS %s", dbName)); err != nil {
+	if _, err := conn.ExecContext(ctx, "DROP DATABASE IF EXISTS "+quoteIdentifier(dbName)); err != nil {
 		return fmt.Errorf("drop test database: %w", err)
 	}
 	return nil
@@ -337,9 +355,9 @@ func tryCreateTestDatabase(ctx context.Context, adminConfig *db.Config, dbName s
 	}
 	defer conn.Close()
 
-	create := fmt.Sprintf("CREATE DATABASE %s", dbName)
+	create := "CREATE DATABASE " + quoteIdentifier(dbName)
 	if template != "" {
-		create += fmt.Sprintf(" TEMPLATE %s", template)
+		create += " TEMPLATE " + quoteIdentifier(template)
 	}
 	if _, err := conn.ExecContext(ctx, create); err != nil {
 		return fmt.Errorf("create test database: %w", err)
@@ -355,6 +373,13 @@ func isRetryableCreateError(msg string) bool {
 	return isTransientServerError(msg) ||
 		strings.Contains(msg, pgObjectInUse) ||
 		strings.Contains(msg, templateSourceInUse)
+}
+
+// isMissingTemplateError reports whether a clone failed because its source
+// template no longer exists (SQLSTATE 3D000). Only meaningful when cloning: the
+// database we are creating cannot itself be the undefined one.
+func isMissingTemplateError(msg string, template string) bool {
+	return template != "" && strings.Contains(msg, pgUndefinedDatabase)
 }
 
 // generateTestDBName creates a unique database name that includes part of the test name for readability.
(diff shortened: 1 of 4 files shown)


mcharles-square:
## Security review response

Thanks — all three findings were accurate readings of the code. Two were real bugs and are fixed in dcc67561d; I disagree with the HIGH severity on the first, but fixed it anyway since the correct version is no harder to write.

### 1. Cleanup could drop databases outside the template namespace — **valid bug, severity overstated**

The wildcard claim is correct and I confirmed it:

```sql
select 'fleetXtestYtmplZproduction' LIKE 'fleet_test_tmpl_%';  -- t
select 'fleet' LIKE 'fleet_test_tmpl_%';                       -- f
```

Every `_` in `fleet_test_tmpl_%` is a single-character wildcard, so the filter was wider than intended. Fixed.

On severity: HIGH overstates it. This is test-harness-only code that never runs in production, and reaching it requires someone to have already created a database on the same server whose name matches `fleet_test_tmpl_` position-for-position with non-underscore characters. The real `fleet` database does not match, nor do per-test databases (`fleet_test_<name>_<hex>`). The "second-order SQL injection" also requires an attacker who can already `CREATE DATABASE` with a quote in the name — i.e. someone who can already drop it directly. I'd call it LOW-to-MEDIUM correctness/hygiene, not HIGH.

Fixed as recommended, all three layers:
- anchored regex in the query: `datname ~ '^fleet_test_tmpl_[0-9a-f]{12}$'`
- re-validated in Go before the name reaches DDL (defence in depth)
- new `quoteIdentifier` on every interpolated identifier, since database names cannot be bound as DDL parameters
- bonus: the *pre-existing* `pg_terminate_backend ... datname = '%s'` literal now binds `$1` instead of interpolating

`TestTemplateNamePatternOnlyMatchesOwnTemplates` pins the boundary (including `fleet`, `fleetXtestYtmplZproduction`, and a quote-injection name), `TestQuoteIdentifier` covers escaping, and `TestDropStaleTemplateDatabasesLeavesLookalikesAlone` plants a real `fleet_test_tmplxdeadbeef1234` database alongside a genuinely stale template and asserts only the latter is dropped.

### 2. Removing stale templates breaks concurrent test runs — **valid**

Correct, including the detail that the "still in use" comment was wrong: a sealed template has no sessions while idle, so the drop succeeds between clones. I verified the resulting failure is a hard one, since it is not in the transient-retry set:

```
ERROR:  3D000: database "definitely_missing_tmpl" does not exist
```

CI is unaffected (each job gets its own database container), but two worktrees on different migration sets against one local server hit it.

I went with recovery rather than removing cleanup: on SQLSTATE 3D000 the harness invalidates the cached template, rebuilds it, and retries. That keeps unbounded template accumulation from coming back on dev machines while making the race non-fatal, and it needs no cross-process lease protocol. A lease covering the full usage lifetime would be the stronger fix, but it is a lot of machinery to protect a ~0.5s rebuild.

`TestGetTestDBRecoversFromSweptTemplate` sweeps the template out from under a running harness. It is a genuine regression test — with `isMissingTemplateError` stubbed to `false` it fails with `error creating test database`, and passes with the fix:

```
template database fleet_test_tmpl_867fe765862f disappeared, rebuilding: create test database: ERROR: database "..." does not exist (SQLSTATE 3D000)
template database fleet_test_tmpl_867fe765862f ready in 487ms
```

### 3. One transient preparation failure disables cloning for the whole process — **valid**

Correct, and the consequence is the exact regression this PR exists to prevent. `sync.Once` is replaced with a mutex plus a `templatePrepared` flag so only *success* is cached; a later test retries, bounded at 3 attempts so a genuinely broken setup stops paying the cost. The initial connection now goes through the existing `waitForServerReady` helper first, so a startup/recovery blip does not even consume an attempt.

### Re-validation after the fixes

| Scope | Result |
| --- | --- |
| `sqlstores`, cold database | 35.7s (was 159s before this PR) |
| Full server suite | 1m11s (was 3m25s) |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| Leftover `fleet_test%` databases after a full run | template only |

The performance result is unchanged, so none of these fixes cost anything measurable. I also found and fixed a leak in my own new test while checking this: its fixture cleanup ran after a deferred `adminDB.Close()`, so the drops were silently no-ops.

mcharles-square:
## Security review response (round 2)

Both findings are valid and both are fixed in 04c68234c, following the recommendations as written.

### 1. Template migration bypasses the preparation timeout — **valid**

Correct: `db.ConnectAndMigrate` runs migrations on a background context, so the 10-minute preparation context never reached the migration, and a wedged migration would hold the advisory lock and the process mutex until Go's package timeout fired — denying callers the fallback that is supposed to make this change fail-safe.

Making migrations context-aware means changing the production `db` API, which I did not want to do from a test-performance PR. I took the second recommendation instead and enforced the bound inside PostgreSQL: `ALTER DATABASE <template> SET lock_timeout` before migrating, which the migrating session inherits. A migration stuck behind a TimescaleDB catalog lock now fails with `lock_not_available`, preparation returns an error, and the caller migrates its own database as intended.

Confirmed in the catalog:

```
fleet_test_tmpl_867fe765862f | {lock_timeout=60000ms} | dbtest-template created=2026-08-21T08:45:31Z
```

Two deliberate limits on this. `statement_timeout` is left unset — migrations are legitimately slow and capping their total runtime would fail valid work; lock waits are the hang mode that matters here. And `templatePrepareTimeout` drops from 10m to **4m**, per the "comfortably below the package timeout" point: at 10m it was equal to Go's default package timeout, so the fallback could never actually run.

An unexpected bonus: the first version of this fix passed `1m0s` to `lock_timeout`, which PostgreSQL rejects (SQLSTATE 22023). Rather than breaking the suite, the run degraded exactly as designed — `could not prepare template database (attempt 1/3: invalid value for parameter "lock_timeout"); migrating this test database directly` — and my own test caught it via a skip. Good evidence the fallback path from finding 3 works. Now formatted as milliseconds.

### 2. Concurrent checkouts can repeatedly delete each other's active templates — **valid, and a better framing than my last fix**

You're right that the 3D000 recovery I added made the race *non-fatal* but not *non-pathological*: A rebuilds and sweeps B, B rebuilds and sweeps A, and the two checkouts can ping-pong. Recovery alone was the wrong stopping point.

Implemented the recommended combination of ownership metadata plus an age threshold:

- when a template is sealed, it records `dbtest-template created=<RFC3339Nano>` as its database comment
- cleanup selects only names matching `^fleet_test_tmpl_[0-9a-f]{12}$` **and** carrying that marker **and** older than **24h**
- anything with no comment, a foreign comment, or an unparseable timestamp is left alone rather than guessed about

A concurrent checkout's template is minutes old, so it is never swept — the ping-pong is structurally impossible while still bounding accumulation on long-lived local databases. The 3D000 recovery stays as a backstop for the genuinely rare case (a run older than 24h, or a manual drop).

I verified clones inherit neither the comment nor per-database settings, so a test database can never be mistaken for a template and never picks up the template's `lock_timeout`:

```
clone_gucs = "no per-db GUCs inherited" | clone_comment = "no comment inherited"
```

`TestDropStaleTemplateDatabasesOnlySweepsOldOwnedTemplates` plants one database per category — old+owned, fresh+owned, right-shape-but-unowned, and a LIKE-only lookalike — and asserts only the first is dropped. `TestTemplateIsStale` covers the comment parsing, including the missing/foreign/unparseable cases.

### Re-validation

| Scope | Result |
| --- | --- |
| `sqlstores`, cold database | 34.4s (159s before this PR) |
| Full server suite | 1m12s (3m25s before), green on 2 consecutive runs |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| Leftover `fleet_test%` databases after a full run | template only |

One unrelated flake to note, since it appeared in one of my local runs: `TestCloseTelemetryDeviceAsyncClosesSynchronouslyWhenCloseWorkersAreExhausted` in `cmd/fleetnode` asserts wall-clock behaviour with 10ms/5ms timeouts and is load-sensitive. It passes 3/3 in isolation on this branch, 5/5 on `origin/main`, and green on two subsequent full-suite runs; it touches no database and no `dbtest` code. Pre-existing timing sensitivity, not something this PR introduces — happy to file it separately.

mcharles-square:
## Security review response (round 3)

Valid, and the arithmetic is the part I got wrong: 3 attempts × 4 minutes = 12 minutes against a 10-minute package timeout, so the "comfortably below" framing I used last round only held for a single attempt. Fixed in c22881e7a, taking both halves of the recommendation.

**One process-wide budget.** Attempts now share a single **3m** budget for the whole process, each capped at **90s**, with the effective deadline being whichever is smaller. The worst case is therefore 3m total no matter how many tests queue up, leaving 7m of the package timeout for the fallback path.

**Circuit opens immediately on a deadline failure.** Running out of time is not bad luck, so a deadline failure no longer counts as one of N attempts — it disables cloning for the rest of the process straight away. Only genuinely transient failures (connection refused, a blip mid-build) still get retried.

**The migration is now bounded too.** Rather than change the production `db` API to be context-aware, a watchdog terminates the migrating backend when the preparation context expires. Combined with the template's `lock_timeout`, that covers non-lock stalls as well, which is the gap you identified.

Verified end-to-end by temporarily shrinking the budget to 300ms:

```
template database fleet_test_tmpl_867fe765862f hit its 299.999917ms preparation deadline
  (migrate template database: deadline exceeded, migration backend terminated: ...);
  migrating test databases directly for the rest of this process
--- PASS: TestZoneKeysWildcard_CrossOrgIsolation (1.02s)
```

The migration is killed, the circuit opens, and the test still passes by migrating its own database — which is the behaviour the finding said was unreachable.

`TestPrepareAttemptTimeoutStaysWithinProcessBudget` pins the arithmetic, including assertions that the budget stays under the package timeout and that no single attempt can consume more than the whole budget.

### One more thing that experiment surfaced

The deadline test left a half-built template behind: failure cleanup reused the preparation context, which had just expired, so the drop silently did nothing. Not a correctness bug — an unsealed template is treated as incomplete and rebuilt, which I confirmed (`datallowconn = t` → rebuilt → `datallowconn = f`) — but it wasted disk until the next run. Cleanup now uses its own short-lived context.

### Re-validation

| Scope | Result |
| --- | --- |
| `sqlstores`, cold database | 34.0s (159s before this PR) |
| Full server suite | 1m14s (3m25s before), green |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| `internal/testutil/dbtest` unit + integration tests | 12 passing |
| Leftover `fleet_test%` databases after a full run | template only |

Summary of the four rounds, for anyone reviewing the stack: the perf change is unchanged in mechanism and result (~3x faster suite, `sqlstores` from 388s to ~73s in CI); the three follow-up commits fix an over-broad `LIKE` in cleanup, a concurrent-checkout rebuild storm, and unbounded preparation. Every fix is covered by a test, and the fallback path — migrate your own database if anything about the template goes wrong — is now exercised deliberately rather than assumed.

mcharles-square:
## Security review response (round 4)

Both valid, both fixed in 905d228f4. The first one is the best catch of the four rounds — it identified a real behavioural difference between cloning and replaying migrations that none of my testing would have surfaced, because it only appears once a template has aged.

### 1. Cached templates preserve stale time-dependent state — **valid, confirmed by measurement**

The premise is real. `migrations/000042_create_curtailment.up.sql` seeds:

```sql
INSERT INTO curtailment_reconciler_heartbeat (id, last_tick_at, last_tick_uuid)
    VALUES (1, CURRENT_TIMESTAMP, ...)
```

Replaying migrations per test made that row "now". Cloning freezes it at template build time. Measured on a 23-minute-old template:

```
heartbeat_age
-----------------
 00:23:07.122302
```

Fixed with the age bound rather than per-clone resets: templates are reused only within **15 minutes** of being built, after which the next process rebuilds. Enumerating time-dependent seed rows would work today, but it silently rots — a future migration adds a seed and nobody updates the reset list. The age bound needs no such maintenance. Cost is one ~0.5s rebuild per 15 minutes per process, which does not move the numbers below.

A template with no recognisable creation stamp is now rebuilt rather than trusted, and `TestTemplateAgeBoundsAreConsistent` asserts the reuse window stays well inside the 24h sweep threshold, so a template can never be swept while still considered current.

Verified by backdating a template's stamp 2 hours:

```
backdated stamp: created=2026-08-21T07:41:40Z
  -> template database fleet_test_tmpl_867fe765862f ready in 556ms   (rebuilt)
new stamp:       created=2026-08-21T09:41:45Z
heartbeat age in a clone of the fresh template: 00:00:07.98
```

**One part I could not reproduce:** overdue TimescaleDB jobs. In a clone, `timescaledb_information.jobs.next_start` is NULL for every policy job, so there is no backlog firing immediately. Happy to be shown otherwise, but I found no evidence of that half of the finding.

### 2. Recovery test deletes the template shared by parallel test packages — **valid, and my test's fault**

Correct, and it was antisocial regardless of the race: any concurrent package or checkout on that server has to rebuild or replay migrations because of my test.

Rewritten to use a private name that never existed (`fleet_test_tmpl_ffffffffffff`) and to drive the same recovery path through `createTestDatabase` directly, asserting the missing template is replaced and never resurrected. Nothing shared is touched. Added `TestInvalidateTemplateDatabaseOnlyForgetsTheNamedTemplate` to cover the state-machine half — recovery for one template must not discard a different valid cached one — which the old test could not check because it only ever had one template in play.

### Re-validation

| Scope | Result |
| --- | --- |
| `sqlstores`, cold database | 34.0s (159s before this PR) |
| Full server suite | 1m13s (3m25s before) |
| `internal/testutil/dbtest` | 15 tests passing |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| Leftover `fleet_test%` databases after a full run | template only |

### Unrelated flake, now quantified

`TestCloseTelemetryDeviceAsyncClosesSynchronouslyWhenCloseWorkersAreExhausted` (`cmd/fleetnode`) failed in one of my full-suite runs. Since it would be convenient to blame the branch, I checked: it also fails on **`origin/main`**, once in two full-suite runs, with this branch's changes stashed. It asserts wall-clock behaviour with 10ms/5ms timeouts, touches no database, and passes 3/3 in isolation here. Pre-existing load sensitivity — I'll file it separately rather than fold it into this PR.

mcharles-square:
## Security review response (round 5)

Valid, and it's a regression I introduced last round: bounding reuse by age created in-place rebuilds of a name other processes may already have cached, and my round-1 eviction — written when the only thing that could attach to a template was a stray TimescaleDB worker — was too blunt for that. Fixed in d14212e0f.

**Eviction is now scoped to sealed templates**, with the `datallowconn` check inside the same statement so there is no decide-then-act window:

```sql
SELECT pg_terminate_backend(activity.pid)
FROM pg_stat_activity activity
JOIN pg_database database ON database.datname = activity.datname
WHERE activity.datname = $1
  AND NOT database.datallowconn
  AND activity.pid <> pg_backend_pid()
```

An unsealed template is somebody's build in progress and the session on it is that build's migration, so it is left alone. I used `datallowconn` rather than the advisory lock as the signal because it is already the publication marker and is visible in the same query — checking lock ownership would need a second round trip and a wider race.

**A clone that meets a rebuild now waits for it** instead of retrying into it. This mattered more than the guard alone: the clone retry budget is ~2s total, while a rebuild is 1–5s, so without waiting the guard would have converted "aborts the rebuild" into "exhausts retries and fails". Waiting is bounded at 2m, and ends early when the template is published or dropped.

### Multi-process validation, as recommended

Two test binaries launched concurrently against a deliberately expired template (stamp backdated 3h), so one must rebuild while the other clones the same name:

```
proc1 (sqlstores):   0 failures, ok ... 33.961s
proc2 (curtailment): 0 failures, ok ... 0.840s

template coordination messages across both processes:
   1 template database fleet_test_tmpl_867fe765862f ready in 518.977458ms
```

Exactly **one** rebuild across both processes, no eviction fight, no fallback replays — the advisory lock serialises the build and the second process reuses the freshly published template.

For the checked-in test I drove the invariant deterministically in-process rather than spawning binaries from a test: `TestEvictTemplateSessionsSpareRebuildsButClearStrays` stands up a real unsealed template with a live session, asserts eviction spares it, then seals it and asserts the now-stray session *is* cleared — both halves matter, since sparing everything would resurrect the original TimescaleDB-worker bug. It fails with exactly the intended message if the guard is removed:

```
--- FAIL: TestEvictTemplateSessionsSpareRebuildsButClearStrays
    eviction must not terminate the migration session of an in-progress rebuild
```

### Re-validation

| Scope | Result |
| --- | --- |
| Full server suite | 1m12s (3m25s before this PR), green |
| `internal/testutil/dbtest` | 17 tests passing |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| Leftover `fleet_test%` databases after a full run | template only |

### Where this leaves the PR

The mechanism and payoff are unchanged from the first commit — `sqlstores` 388s → ~74–86s in CI, full local suite 3m25s → 1m12s. The five follow-ups are all hardening of the same harness, each with a test: over-broad `LIKE` in cleanup, concurrent-checkout sweep storms, unbounded preparation, stale `CURRENT_TIMESTAMP` seed data in aged templates, and now rebuild-versus-clone contention. The recurring theme is that sharing one template across processes and runs has more edges than replaying migrations did; the invariant that holds them together is that *sealed* is the only state a clone trusts, and everything else is rebuilt, waited for, or fallen back from.

mcharles-square:
## Security review response (round 6)

Fixed in f7f19bb1c. Two of the three are valid — the first is the best catch so far, because it identified that my *previous* round silently disarmed a safeguard the round before it. The second I could not reproduce and am pushing back on with measurements.

### 1. Migration watchdog cannot terminate the migration it watches — **valid, and a regression I introduced**

Exactly right. Round 3 added the watchdog; round 5 scoped `evictTemplateSessions` to sealed templates; the template is connectable for the whole of its migration, so from round 5 onward the watchdog's terminate query matched nothing. My round-3 evidence was collected *before* round 5 and I did not re-run it — that's on me.

The watchdog now uses its own unconditional terminate (`terminateTemplateBuildSessions`) rather than the clone-path helper. That is safe here for a reason worth stating: the caller holds the preparation advisory lock, so no other process can be building this template, and clones never connect to their source — the only session on it is our own migration. Preparation also waits for the watchdog to finish before dropping the template, so a late termination cannot land on an unrelated database that reuses the name.

Re-verified with the budget shrunk to 400ms, which is the test I should have re-run last round:

```
migrate template database: deadline exceeded, migration backend terminated:
  ... migration failed in line 0: -- Rename the notification permission keys ...
migrating test databases directly for the rest of this process
--- PASS: TestZoneKeysWildcard_CrossOrgIsolation (1.08s)
```

The migration is killed mid-run, the circuit opens, and the test still passes by migrating its own database.

I did not take the "capture the backend PID" route: `db.ConnectAndMigrate` owns its connection and does not expose one, so getting the PID means changing production code. Scoping by database name plus the lock invariant achieves the same thing.

### 2. Warm clones inherit overdue TimescaleDB job schedules — **not reproduced**

The schedule detail is right: `000006_create_continuous_aggregates.up.sql` does create a policy with a 30-second `schedule_interval`. But the impact does not follow, because clones carry **no** run state to be overdue. Comparing a clone against a database built by replaying migrations, via the same harness:

```
FRESHLY MIGRATED: jobs=22 with_next_start=0
CLONE:            jobs=22 with_next_start=0
```

`next_start` is NULL for all 22 jobs in both cases, and `count(*) filter (where next_start < now())` is 0. The scheduler computes the first run when it attaches to the clone, so there is no backlog to fire, and the job set is identical to what the old per-test migration path produced — so whatever those jobs do, this PR does not change it.

The concurrent-clone test you suggested is effectively the multi-process run from last round: two binaries cloning a **3-hour-old** template (far beyond the 30s job interval), 0 failures on both sides. If there's a specific query that would show the overdue state you have in mind, I'll happily run it — I could not construct one.

### 3. Template reuse is not scoped to the database owner — **valid PostgreSQL semantics, not reachable with this migration set**

The mechanism is real and I confirmed it directly:

```
$ psql -U other_dev -c 'CREATE DATABASE other_clone TEMPLATE fleet_test_tmpl_...'
ERROR:  42501: permission denied to copy database "fleet_test_tmpl_867fe765862f"
```

Implemented both recommendations: the connecting role is now hashed into the template fingerprint (so two roles never contend for one name, and a long role name cannot overflow the 63-char identifier limit), reuse additionally verifies `pg_get_userbyid(datdba) = current_user`, and a clone that hits 42501 stops cloning for the rest of the process instead of retrying a permission that cannot be granted.

On the impact claim, though: I tried to reproduce "every database-backed test fails until the foreign template expires" and could not, because a role that cannot copy another role's database also cannot run this migration set. `000115_system_monitoring.up.sql` does `CREATE EXTENSION pg_stat_statements`, which requires superuser:

```
as other_dev (CREATEDB, not superuser):
  ... migrating this test database directly
  ERROR: permission denied to create extension "pg_stat_statements" (SQLSTATE 42501)
  --- FAIL
```

Note the fallback did engage — it stopped using the template and migrated directly — and then failed on the migration itself, which is a pre-existing property of the migration set, not of this change. Two superuser roles, meanwhile, can copy each other's databases freely. So I've kept the fix as defence in depth (it makes an unreachable failure graceful and costs nothing), but I don't believe the stated impact is currently attainable.

### Re-validation

| Scope | Result |
| --- | --- |
| `sqlstores`, cold database | 33.4s (159s before this PR) |
| Full server suite | 1m12s (3m25s before), green |
| `internal/testutil/dbtest` | 19 tests passing |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| CI `sqlstores` on the previous commit | `ok … 85.357s` (388s on `main`) |

New tests: role scoping of the fingerprint (including a long-role-name identifier bound) and 42501 classification, including an assertion that it is *not* absorbed by the generic retry path.

### A note on this review loop

Six rounds in, the findings have moved from the original diff to the fixes themselves, and this round caught a fix breaking an earlier fix — which is a fair signal that the hardening has reached the point of diminishing returns. The mechanism and payoff have not changed since the first commit (`sqlstores` 388s → ~85s in CI, local suite 3m25s → 1m12s), and every behaviour now has a test that fails without its fix. I'd suggest human review of the invariant rather than another automated pass: **a clone only ever trusts a template that is sealed, owned by us, and recent; everything else is rebuilt, waited for, or falls back to migrating directly.** If that holds, the rest is detail.

mcharles-square:
## Security review response (round 7)

Both valid, fixed in 1b9d35849 — and writing the test for the first one uncovered a third problem the review had not spotted, which is the most useful outcome of this round.

### 1. Abandoned templates can trigger a long retry stall — **valid**

Right on both counts: `templateRebuildInProgress` inferred "somebody is building this" from connectability alone, and each attempt got its own two-minute wait.

- **Waiting now requires evidence of a live build**: the template must be unsealed *and* somebody must hold its preparation lock. An unsealed template with no builder is classified `templateStateAbandoned` and rebuilt immediately.
- **All waits for one test database share a single deadline** (90s total, not per attempt).
- Rebuilding is also what clears the stranded session, since eviction deliberately spares unsealed templates — so the abandoned case had no exit before this.

### 2. Final-attempt recovery exits without creating the database — **valid**

Correct, including the consequence: the recovery would succeed and then the loop would exit and report the stale error. Recoveries now draw on their own budget (`templateRecoveryMaxAttempts`), separate from the DDL retry counter, because they change the situation rather than hoping a transient failure passes — so each one earns a fresh attempt. The branchy loop body moved into `recoverTemplateForClone`, which returns whether it acted and which template to use next.

### 3. What the test found: the preparation lock was global

Writing the abandoned-template test surfaced a bug neither of us had named. The advisory lock used one fixed key for all templates, so "somebody holds the lock" was not evidence about *this* template — a process building template A made an abandoned template B look under construction, and clones of B would wait instead of rebuilding.

It showed up as my new tests failing only under the parallel full suite, which is exactly the interference the finding describes, one level up. Lock keys are now derived per template name (`sha256(namespace + name)`, masked to 63 bits). Besides fixing the evidence, this lets unrelated templates — different roles, or different migration sets across checkouts — build concurrently instead of serialising on one key.

Also, my abandoned-template test initially sabotaged the *live* template and was flaky under the parallel suite for precisely the reason you gave me two rounds ago. It now plants a private template. Same lesson, second offence; the tests that touch shared state are the ones to distrust.

### Verification

| Check | Result |
| --- | --- |
| Abandoned template with a stranded session | recovers in ~0.8s (previously: failed after 28.7s of 55006 retries) |
| Two concurrent binaries, 3h-expired template | 0 failures, exactly 1 rebuild, no waiting |
| `dbtest` + `sqlstores` + `curtailment` concurrently, 3 runs | green, green, green |
| Full server suite, 2 consecutive runs | green (1m12s, 1m02s) |
| `internal/testutil/dbtest` | 23 tests passing |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |

New tests: abandoned-vs-building state detection across all four combinations of sealed/unsealed and lock held/not held; per-template lock key distinctness, stability, and non-negativity (the `pg_locks` reconstruction requires a positive bigint); and the recovery-versus-retry budget invariant, including a worst-case bound that stays well inside the package timeout.

### On continuing

Seven rounds in, rounds 5–7 have each been about a fix from the previous round rather than the original change, and this round's real find came from *writing a test*, not from another review pass. The reviewed diff is now ~5x the size of the change that started it, all of it hardening one test helper.

I'm going to stop here and ask for human review. The payoff is unchanged and measured — CI `sqlstores` 388s → ~85s, local suite 3m25s → ~1m10s — and the invariant is small enough to audit directly:

> **A clone trusts a template only if it is sealed, owned by us, and recent. Anything else is rebuilt, waited for (only with proof of a live builder, on a shared deadline), or abandoned in favour of migrating the test's own database.**

Every clause of that sentence now has a test that fails without its implementation. If a reviewer disagrees with the invariant, the right move is to change it deliberately rather than to keep patching the edges; if they agree, the remaining risk is bounded by the fact that every failure path ends in "migrate directly", which is what the harness did before this PR.

mcharles-square:
## Security review response (round 8)

Both valid, fixed in 310bd52f6. The second one was a straightforward bug in my own tests, and the first made me reverse an earlier decision of mine.

### 1. Migration deadline cannot guarantee termination — **valid**

Correct: the deadline rested entirely on `pg_terminate_backend` succeeding, and that call discarded its own errors, so a failed termination left the migration blocked with the advisory lock held.

**I've reversed my round-3 position and set `statement_timeout` on the template**, derived from the per-attempt deadline (90s), alongside the existing `lock_timeout` (60s). I declined it then on the grounds that migrations are legitimately slow — which is true in production but wrong here: these run against an *empty* database and take milliseconds each, so a single statement outliving the entire preparation attempt is wedged, not slow. That gives a guarantee that does not depend on the watchdog at all:

```
fleet_test_tmpl_fdb5f60bf1bc | {lock_timeout=60000ms,statement_timeout=90000ms}
```

Clones inherit neither setting (verified: `statement_timeout = 0`, `lock_timeout = 0` in a clone), so test databases are unconstrained.

The watchdog is now belt-and-braces rather than load-bearing, and it no longer lies: termination retries until no sessions remain, reports failure, and a migration that somehow finishes *after* a failed termination is treated as untrustworthy and not published.

On the recommendation to add a context-aware migration API: that means changing `db.ConnectAndMigrate`, which production also uses. I'm not doing that from a test-harness PR — the server-side timeouts achieve the same bound without touching the production path. If we want a context-aware migration API, it's worth doing deliberately, with the startup path in scope, as its own change.

### 2. Fixed database fixture names collide across concurrent runs — **valid, my bug**

Right, and it's the third variant of the same mistake in this PR (first the recovery test dropping the live template, then the abandoned-template test sabotaging it, now constant fixture names). Each test now derives a random 12-hex suffix per invocation and drops only what it created.

Confirmed the failure mode was real, not theoretical — two concurrent runs of the package:

| Fixture naming | Result |
| --- | --- |
| Constant names (before) | **2 failures in each run** |
| Per-invocation random suffix (after) | both runs green |

The remaining constant `fleet_test_tmpl_*` strings in the test file are pure in-memory fixtures — pattern matching, error-message classification, cache state, lock-key hashing — and never reach `CREATE DATABASE`.

### Verification

| Check | Result |
| --- | --- |
| Two concurrent runs of `dbtest` | both green (2 failures each before) |
| `sqlstores`, cold database | 33.1s |
| Full server suite, 2 consecutive runs | green (1m13s, 1m02s) |
| `golangci-lint ./internal/testutil/dbtest/...` | 0 issues |
| Template settings applied / not inherited by clones | confirmed in the catalog |

### Handing over to human review

This is round 8. Rounds 5–8 have each been about a fix from the previous round, and the two genuine defects found since round 6 came from *writing tests* (the global lock key, the fixture collisions) rather than from review passes. I'm treating the automated loop as complete and won't be posting another round of changes in response to it.

The change that started this is unchanged in mechanism and payoff: CI `sqlstores` **388s → ~85s**, local suite **3m25s → ~1m05s**, with per-test database setup going from ~2s of migration replay to a ~20-40ms clone. Everything after the first commit is hardening of one test helper, and the whole of it reduces to one auditable rule:

> **A clone trusts a template only if it is sealed, owned by us, and recent. Anything else is rebuilt, waited for (only with proof of a live builder, on a shared deadline), or abandoned in favour of migrating the test's own database — and every failure path ends where the harness started, migrating directly.**

That last clause is the safety argument: the worst case of any bug in this code is the old behaviour, which is slow, not wrong. Every clause has a test that fails without its implementation.

server/internal/testutil/dbtest/dbtest.go
```diff
@@ -226,32 +239,140 @@ func pingAdminDatabase(ctx context.Context, adminConfig *db.Config) error {
...
+		if remaining := time.Until(rebuildWaitDeadline); remaining > 0 {
+			t.Logf("template %s is being rebuilt by another process, waiting up to %s", template, remaining)
+			waitForTemplateRebuild(t.Context(), adminConfig, template, rebuildWaitDeadline)
+			return true, template
+		}
+
+		// Waited as long as we are prepared to. Stop waiting and rebuild.
+		t.Logf("template %s still unpublished after %s, rebuilding it", template, templateRebuildWaitTimeout)
+		invalidateTemplateDatabase(template)
+		return true, templateDatabase(t, adminConfig)
```
rl-block, line 360: We could probably fall back to direct migration once `rebuildWaitDeadline` is exhausted? Calling `templateDatabase` here starts a fresh advisory-lock wait, so recovery can exceed the shared 90-second bound. A test holding the lock past the deadline might help pin this behavior.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,901 @@
...
+		return fmt.Errorf("pin admin connection: %w", err)
+	}
+	defer adminConn.Close()
+
+	lockKey := templateLockKey(name)
+	if _, err := adminConn.ExecContext(ctx, "SELECT pg_advisory_lock($1)", lockKey); err != nil {
+		return fmt.Errorf("acquire template advisory lock: %w", err)
+	}
+	defer func() {
+		_, _ = adminConn.ExecContext(ctx, "SELECT pg_advisory_unlock($1)", lockKey)
```
rl-block, line 278: Do we want to route the parameterizable catalog, advisory-lock, and session-management queries through sqlc as well? `AGENTS.md` says all server DB access goes through sqlc; perhaps only identifier-bearing DDL needs to remain direct.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,901 @@
...
+	// Bound the migrating session inside PostgreSQL, since we cannot cancel it
+	// from Go. These are the only guarantees that do not depend on the watchdog
+	// below succeeding: whatever happens, the server itself gives up eventually.
+	// Milliseconds, not Duration.String(): PostgreSQL rejects Go's "1m0s" form.
+	for setting, limit := range map[string]time.Duration{
+		"lock_timeout": templateLockTimeout,
+		// Derived from the per-attempt deadline. Migrations run against an empty
+		// database and take milliseconds each, so a single statement outliving
+		// this is wedged, not slow.
+		"statement_timeout": templatePrepareTimeout,
```
rl-block, line 335: These PostgreSQL timeouts could probably use the current attempt's remaining deadline. A late attempt may have less than 90 seconds left, and the watchdog can run before a backend exists. We could keep cancellation active until `ConnectAndMigrate` returns and add a shortened-budget test.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,901 @@
...
+
+// templateNameFor returns the template name this configuration may use.
+//
+// The connecting role is part of the identity, not just the migration set:
+// PostgreSQL only lets a role copy a database it owns (or any, for superusers),
+// so two checkouts on one cluster using different roles must not compete for one
+// template name. Without this the second role gets "permission denied to copy
+// database" on every clone.
+func templateNameFor(config *db.Config) string {
+	return templateDBPrefix + migrationSetFingerprint(config.Username)
```
rl-block, line 854: Should template identity use the effective role from `ExplicitDSN`? The DSN overrides `Config.Username`, so different DSN users can hash to the same name and hit ownership failures. Perhaps we could derive the connected role and cover two explicit-DSN users in a test.

rl-block (approved):

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,328 @@
...
+		return fmt.Sprintf("%x", time.Now().UnixNano())[:templateFingerprintLength]
+	}
+
+	for _, name := range names {
+		contents, err := fs.ReadFile(migrations.Migrations, name)
+		if err != nil {
+			return fmt.Sprintf("%x", time.Now().UnixNano())[:templateFingerprintLength]
+		}
+		_, _ = hash.Write([]byte(name))
+		_, _ = hash.Write(contents)
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Not adding the extension versions to the fingerprint, but you're right that the drift is real -- documenting where it's bounded instead.

Template identity keys on the migration set and the role, deliberately not on the database runtime, so making it depend on installed extension versions means a live catalog query before the name can even be computed. What already covers the scenario is `templateMaxAge`: reuse is capped at 15 minutes, so a template carrying the pre-upgrade extension catalog is retired within that interval regardless of the fingerprint. The image rebuild you describe self-heals on the next age-out.

I've extended the `templateMaxAge` comment to say so explicitly, since "what the fingerprint deliberately doesn't hash" is exactly the thing a reader needs to know the age bound is load-bearing for.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,401 @@
...
+	defer func() {
+		_, _ = adminConn.ExecContext(ctx, "SELECT pg_advisory_unlock($1)", templateAdvisoryLockKey)
+	}()
+
+	ready, err := templateDatabaseReady(ctx, adminConn, name)
+	if err != nil {
+		return err
+	}
+	if ready {
+		return nil
```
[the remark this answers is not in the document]
mcharles-square: 🤖 This is the case `templateMaxAge` exists for, and it names the same row you cite.

Reuse is bounded at 15 minutes precisely because cloning copies `CURRENT_TIMESTAMP` seed data -- the constant's comment calls out `curtailment_reconciler_heartbeat` from `000042` by name. The alternative you suggest, resetting time-dependent seeds per clone, means enumerating every volatile seed and keeping that list current as migrations land; the age bound gets bounded staleness without that maintenance burden.

Checked the current exposure: no test asserts curtailment heartbeat freshness today, so nothing is affected. A future test with a sub-15-minute freshness threshold would be the thing to watch, and it would want to seed its own heartbeat row rather than rely on the clone's.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,479 @@
...
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Fixed -- the sweep now runs on the reuse path too.

The detail that decided it is the one you point at: because the sweep only ever ran when a *new* fingerprint got built, a long-lived volume cycling among a fixed set of branches accumulated every one of their templates and nothing ever reconsidered them. Preparation happens once per process, so running the sweep unconditionally costs one catalog query.

`buildTemplateUnderLock` no longer needs to report whether it built anything, so it returns a plain `error` now.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,566 @@
...
+	defer close(migrationDone)
+
+	go func() {
+		select {
+		case <-migrationDone:
+		case <-ctx.Done():
+			// The parent context is spent, so give the eviction its own.
+			// nolint: usetesting
+			evictCtx, cancel := context.WithTimeout(context.Background(), templateDetachTimeout)
+			defer cancel()
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Good catch, and this was the most important comment on the PR -- fixed.

You're right about the mechanism, and it's slightly worse than described: when the deadline fires with no backend attached yet, `terminateTemplateBuildSessions` finds nothing to kill and returns **nil**, so the watchdog reported a *successful* termination having terminated nothing. If the server then came back, `migrate.Up()` ran with nothing watching and `migrateTemplateDatabase` blocked until it returned -- past both the attempt timeout and the process budget.

The watchdog now sweeps in a loop until `migrationDone` closes or its own kill budget expires, and if the budget expires with nothing ever attached it synthesizes an explicit error ("still running past its deadline, with no backend to terminate") rather than reporting the empty sweeps as success. `terminateTemplateBuildSessions` is back to being a single sweep -- whether another one is worth making depends on whether the migration has returned, which only the watchdog knows.

Not making `ConnectAndMigrate` context-aware here; that's a change to `infrastructure/db` with callers well beyond this harness.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,724 @@
...
+//   - it is owned by the role we are connecting as, since PostgreSQL refuses to
+//     copy another role's database unless we are a superuser.
+//
+// A template with no recognisable creation stamp (an older format, or one built
+// before stamping existed) is rebuilt rather than trusted.
+func templateIsReusable(ctx context.Context, adminConn *sql.Conn, name string) (bool, error) {
+	var allowConnections bool
+	var ownedByCurrentRole bool
+	var comment sql.NullString
+	err := adminConn.QueryRowContext(ctx, `
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Declining this one -- I don't think the rule applies here, and sqlc can't express these queries.

`server/sqlc.yaml` points sqlc at `migrations/` for its schema, so `pg_database`, `pg_locks` and `pg_stat_activity` aren't in the schema it generates from and it cannot type them at all. The rest is DDL that can't take bind parameters (`CREATE`/`DROP`/`ALTER DATABASE`), which is why `quoteIdentifier` exists. The harness also runs against the maintenance database and against databases at arbitrary migration versions, neither of which the generated layer describes.

AGENTS.md rule 7 reads "Server: prepared statements only. All DB access goes through sqlc" -- this is a test-only leaf package addressing the maintenance database, not server DB access. Worth noting the direction of travel: the pre-existing `dbtest.go` on `main` already ran this same catalog SQL through `database/sql`, with `fmt.Sprintf` interpolating the database name straight into the statement. This PR replaced that with `pq.QuoteIdentifier`, so it moved the code toward the rule's intent rather than away from it.

I've added a paragraph to the package comment recording all of the above, since this is the second time it's been asked.

server/internal/testutil/dbtest/template.go
```diff
@@ -0,0 +1,852 @@
...
+	// Bound the migrating session inside PostgreSQL, since we cannot cancel it
+	// from Go. These are the only guarantees that do not depend on the watchdog
+	// below succeeding: whatever happens, the server itself gives up eventually.
+	// Milliseconds, not Duration.String(): PostgreSQL rejects Go's "1m0s" form.
+	for setting, limit := range map[string]time.Duration{
+		"lock_timeout": templateLockTimeout,
+		// Derived from the per-attempt deadline. Migrations run against an empty
+		// database and take milliseconds each, so a single statement outliving
+		// this is wedged, not slow.
+		"statement_timeout": templatePrepareTimeout,
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Fixed -- these derive from the actual attempt deadline now.

New `remainingAttemptTimeout(ctx)` reads the attempt context's deadline and clamps to `templatePrepareTimeout`, and `statement_timeout` is set from it. It never returns zero, since PostgreSQL reads `statement_timeout = 0` as "no limit" and a spent budget would otherwise silently remove the bound being set.

`lock_timeout` stays at `templateLockTimeout` on purpose: it bounds a single lock wait rather than total runtime, and shrinking it with the remaining budget would make a late attempt fail on lock contention that an early one tolerates.

Also fixed the related budget bug this exposed: `templatePrepareBudget` was enforced against a wall-clock deadline set on the first attempt and never cleared, so it charged the budget for time the binary spent happily running tests off a working template. It's a cumulative `templateSpent` now. Test added -- `TestRemainingAttemptTimeoutTracksTheAttemptDeadline`.

server/internal/testutil/dbtest/dbtest.go
mcharles-square: 🤖 Done -- falls back to direct migration, and there's a test.

`recoverTemplateForClone` returns `(true, "")` on this path now, so the caller migrates its own database. The cached template is still invalidated, so the *next* test rebuilds rather than queueing behind the same stalled build.

Your read on the cause was exactly right: calling `templateDatabase` here started a fresh advisory-lock wait on top of a deadline that was already spent, which is the accumulation that deadline exists to prevent.

`TestRecoveryFallsBackWhenTheRebuildDeadlineIsSpent` pins it -- holds the preparation lock on a pinned connection over an unsealed fixture template so the state reads as a live build, then calls recovery with an already-expired deadline and asserts it returns `""` promptly. No 90-second wait needed, so it's fast and deterministic.

While in there: the same class of leak existed one level up. The server-ready wait renewed its own 30s on every loop iteration, and a recovery that re-prepared the template started an advisory-lock wait outside the budget entirely -- so a single `createTestDatabase` call could run ~9 minutes against Go's 10-minute package timeout while the guard test asserted ~95s. Every wait belonging to one database now shares the one deadline (`waitDeadline`, renamed from `rebuildWaitDeadline` since it covers more than rebuilds), and `templateDatabase` takes a caller deadline that caps its lock wait. `TestRecoveryBudgetIsSeparateFromRetryBudget` now asserts the two relationships that make its arithmetic true.

server/internal/testutil/dbtest/template.go
mcharles-square: 🤖 Same answer as the Codex thread on this -- declining, with the reasoning now recorded in the package comment.

Short version: sqlc's schema is `migrations/` (`server/sqlc.yaml`), so the catalog relations this harness reads -- `pg_database`, `pg_locks`, `pg_stat_activity` -- aren't in the schema it generates from and it can't type them. You're right that the split you describe is the correct principle; it just lands with everything on the direct side here, because the parameterizable queries are all catalog reads and the rest is identifier-bearing DDL.

Your instinct on the direction is right though: the version of this file on `main` interpolated database names with `fmt.Sprintf`, and this PR moved them to `pq.QuoteIdentifier`.

server/internal/testutil/dbtest/template.go
mcharles-square: 🤖 Fixed, all three parts.

**Timeouts from the remaining deadline** -- new `remainingAttemptTimeout(ctx)` derives `statement_timeout` from the attempt context, clamped to `templatePrepareTimeout` and never zero (PostgreSQL reads 0 as "no limit"). `lock_timeout` stays fixed: it bounds one lock wait rather than total runtime.

**Watchdog before a backend exists** -- this was the real bug, and worse than a timing window. With nothing attached, `terminateTemplateBuildSessions` returned nil, so the watchdog reported success having killed nothing and exited; if the server recovered, `migrate.Up()` then ran unwatched past both the attempt timeout and the process budget. It now sweeps in a loop until `ConnectAndMigrate` returns, and reports an explicit error if its kill budget expires with the migration still running.

**Shortened-budget test** -- `TestRemainingAttemptTimeoutTracksTheAttemptDeadline` covers no-deadline, over-cap, a 5s attempt, and an already-spent deadline.

This also turned up why a late attempt could exist at all: `templatePrepareBudget` was measured against a wall-clock deadline set on the first attempt and never cleared, so it counted time the binary spent running tests off a working template. Now a cumulative `templateSpent`.

server/internal/testutil/dbtest/template.go
mcharles-square: 🤖 Answered differently than asked, and I think in a better place -- explicit DSNs no longer build templates at all.

You're right that `DSN()` returns `ExplicitDSN` verbatim and ignores `Config.Username`, so hashing `Username` can put two different effective roles on one template name. But chasing the role turned out to be treating a symptom: because `DSN()` also ignores `Name`, *every* config this harness routes by name resolves to the DSN's single database -- including the staging database a template is migrated in. That staging database would stay empty, then get sealed and renamed into the shared template name for every other process on the cluster to clone. A wrong-owner clone failure is a slow suite for one run; publishing an empty template is a slow suite for everyone who touches that cluster for the next 15 minutes.

So `templateDatabase` now bails out when `UsesExplicitDSN()` is true and migrates test databases directly. No template is built or cloned in that configuration, which makes the ownership collision unreachable rather than merely handled.

The test you suggest is worth having, but as "an explicit DSN disables cloning" rather than two DSN users colliding -- happy to add it if you'd like it in this PR. Worth noting for scope: `DB_DSN` isn't set by the justfile, by CI, or by any `t.Setenv` in the repo; docker-compose only forwards it to the deployed container. So this needed a developer to have it exported in their own shell.

PR #957, opened 2026-08-21
Enforce persisted curtailment authorization envelopes

negarn:
Reviewable diff: +299/-448 across 8 files (excludes generated, test, and story files).

## Summary

Curtailment events and response profiles now authorize reads, mutations, and recovery operations from the durable authorization envelope captured when each record is written. This lets site-scoped operators retain access to the records they were authorized to create even if topology membership changes later, while malformed or incomplete persisted envelopes fail closed. This is the next backend slice of issue #909; topology execution and the remaining lifecycle work stay tracked there.

## How it works

Handlers first require that the caller has the relevant curtailment capability at some supported scope, load the organization-owned event or profile, and strictly parse its persisted authorization envelope. Miner and facility-fan coverage are converted into site resource contexts; unbounded coverage requires an organization-wide grant, and facility-fan sites additionally require `site:read`.

Single-record operations enforce every required context before returning or mutating the record. List operations apply the same checks with per-request permission caches, filter unauthorized records, and only then hydrate display-oriented target coverage. Response-profile deletion also includes the authorized envelope in its SQL optimistic predicate, so a concurrent envelope change cannot invalidate the permission decision before deletion.

## Diagrams

```mermaid
flowchart TD
    A["Event or profile RPC"] --> B["Require scoped curtailment capability"]
    B --> C["Load organization-owned record"]
    C --> D["Strictly parse persisted authorization envelope"]
    D --> E["Check curtailment permission for miner and fan coverage"]
    E --> F["Check site:read for facility-fan coverage"]
    F --> G["Filter or continue the lifecycle operation"]
    G --> H["Hydrate display-only target coverage"]
```

```mermaid
sequenceDiagram
    participant H as Handler
    participant S as Response profile service
    participant DB as PostgreSQL
    H->>S: Delete with authorized selector, fan, and envelope snapshots
    S->>DB: DELETE where every expected snapshot still matches
    DB-->>S: One row or zero rows
    S-->>H: Success or FailedPrecondition
```

## Areas of the code involved

| --- | --- | --- |
| `server/internal/domain/curtailment/` | Added strict authorization-envelope parsing and validation | Establishes the fail-closed trust boundary for persisted coverage |
| `server/internal/handlers/curtailment/handler.go` | Replaced live topology reconstruction with envelope-based event authorization and cached list filtering | Controls event visibility and lifecycle/recovery authority |
| `server/internal/handlers/curtailment/response_profiles.go` | Applies the same persisted checks to profile get/list/update/delete flows | Keeps profile behavior aligned with event authorization |
| `server/internal/domain/stores/` and `server/sqlc/queries/` | Threads the expected envelope through response-profile deletion | Makes authorization and deletion atomic with respect to concurrent changes |
| `server/generated/sqlc/` | Regenerated query bindings | Generated — skip |
| Curtailment domain, handler, and SQL-store tests | Covers strict parsing, site-scoped access, filtering, recovery, and stale-envelope deletion | Verifies the authorization boundary and regression scenarios |

- Persisted envelopes are the authority for record reads and recovery; current topology is used only for display hydration, avoiding authorization drift after moves or deletions.
- The parser requires the exact supported schema and fields; invalid stored data surfaces as an internal error instead of widening access.
- Facility fans require both the curtailment grant and `site:read` for every captured fan site.
- List endpoints cache organization-wide and per-site decisions for the request rather than repeating middleware checks per record.
- Automation entry points remain organization-scoped until immutable automation bindings have equivalent persisted authorization coverage.
- Response-profile deletion compares the envelope alongside the selector and fan snapshots, preferring a failed precondition over a stale authorization decision.

## Testing & validation

- `just gen`
- `cd server && just lint`
- `cd server && go test ./internal/domain/curtailment ./internal/handlers/curtailment -count=1`
- `cd server && go test ./internal/domain/stores/sqlstores -run '^(TestBuildAuthorizationEnvelopeJSON|TestLockTopologyScopeCoverage)' -count=1`
- `git diff --check origin/main...HEAD`
- The full database-backed SQL-store suite was not runnable locally because the configured local Postgres rejected the `fleet` password; focused in-memory SQL-store authorization tests passed, and CI remains responsible for the database integration run.

Refs #909

### DIFF
--- a/server/internal/domain/curtailment/authorization_envelope.go
+++ b/server/internal/domain/curtailment/authorization_envelope.go
@@ -1,6 +1,7 @@
 package curtailment
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 
@@ -37,6 +38,11 @@ func AuthorizationEnvelopeFromJSON(raw []byte) (models.AuthorizationEnvelope, er
 	for _, field := range authorizationEnvelopeFields {
 		if _, ok := fields[field]; !ok {
 			return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("is missing %q", field)
+		}
+	}
+	for _, field := range [...]string{"miner_scope_unbounded", "facility_fan_scope_unbounded"} {
+		if bytes.Equal(bytes.TrimSpace(fields[field]), []byte("null")) {
+			return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("%s must be a boolean", field)
 		}
 	}
 
--- a/server/internal/domain/curtailment/authorization_envelope.go
+++ b/server/internal/domain/curtailment/authorization_envelope.go
@@ -1,0 +1,85 @@
+package curtailment
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/block/proto-fleet/server/internal/domain/curtailment/models"
+	"github.com/block/proto-fleet/server/internal/domain/fleeterror"
+)
+
+var authorizationEnvelopeFields = [...]string{
+	"schema_version",
+	"selected_resource_site_ids",
+	"current_member_site_ids",
+	"miner_scope_unbounded",
+	"facility_fan_site_ids",
+	"facility_fan_scope_unbounded",
+}
+
+// AuthorizationEnvelopeFromJSON parses the persisted authorization snapshot.
+// It is intentionally stricter than the database shape checks so every caller
+// fails closed on missing, unknown, null, or invalid coverage.
+func AuthorizationEnvelopeFromJSON(raw []byte) (models.AuthorizationEnvelope, error) {
+	var fields map[string]json.RawMessage
+	if len(raw) == 0 {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("must be set")
+	}
+	if err := json.Unmarshal(raw, &fields); err != nil {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("must be a JSON object: %v", err)
+	}
+	if fields == nil {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("must be a JSON object")
+	}
+	if len(fields) != len(authorizationEnvelopeFields) {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("must contain exactly the supported fields")
+	}
+	for _, field := range authorizationEnvelopeFields {
+		if _, ok := fields[field]; !ok {
+			return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("is missing %q", field)
+		}
+	}
+
+	var envelope models.AuthorizationEnvelope
+	if err := json.Unmarshal(raw, &envelope); err != nil {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("contains an invalid field: %v", err)
+	}
+	if envelope.SchemaVersion != models.AuthorizationEnvelopeSchemaVersion {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope(
+			"has unsupported schema_version %d", envelope.SchemaVersion,
+		)
+	}
+	if envelope.SelectedResourceSiteIDs == nil ||
+		envelope.CurrentMemberSiteIDs == nil ||
+		envelope.FacilityFanSiteIDs == nil {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("site ID fields must be arrays")
+	}
+	if err := validateAuthorizationEnvelopeSiteIDs(envelope.SelectedResourceSiteIDs, "selected_resource_site_ids"); err != nil {
+		return models.AuthorizationEnvelope{}, err
+	}
+	if err := validateAuthorizationEnvelopeSiteIDs(envelope.CurrentMemberSiteIDs, "current_member_site_ids"); err != nil {
+		return models.AuthorizationEnvelope{}, err
+	}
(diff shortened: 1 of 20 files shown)


server/internal/domain/stores/sqlstores/response_profile_facility_fans_integration_test.go
```diff
@@ -143,13 +143,28 @@ func TestSQLCurtailmentStore_ResponseProfileFacilityFanSettings(t *testing.T) {
...
+		orgID,
+		created.ID,
+		replacement.SiteID,
+		replacement.ScopeJSON,
+		updated.AuthorizationEnvelopeJSON,
+		responseProfileFanSettings(replacement),
+	)
+	require.Error(t, err)
+	assert.True(t, fleeterror.IsFailedPreconditionError(err))
+	_, err = service.Get(ctx, orgID, created.ID)
```
rl-block, line 165: Maybe add the successful CAS case here too? These assertions prove stale envelopes fail, but an always-false or misbound predicate would still pass. Maybe delete with `replacement.AuthorizationEnvelopeJSON` and verify the profile is gone.

server/internal/handlers/curtailment/handler.go
```diff
@@ -702,15 +770,13 @@ func (h *Handler) requireForceReleasePermission(ctx context.Context, orgID int64
...
-			return err
-		}
+	if event == nil || event.OrgID != orgID {
+		return fleeterror.NewNotFoundError("curtailment event not found")
+	}
+	if err := requireAuthorizationEnvelopePermissions(ctx, permission, event.AuthorizationEnvelopeJSON); err != nil {
 		return err
 	}
-	return requireResourceContextPermissions(ctx, permission, requirements)
+	return h.hydrateTargetSiteCoverageByEvent(ctx, orgID, event)
```
rl-block: Worth having target-site hydration happen inside the permission helper? A coverage-query failure here can block Stop, AdminTerminate, or ForceRelease after envelope authorization succeeds. Maybe this helper could remain authorization-only, with read responses hydrated separately or control responses hydrated best-effort.

server/internal/handlers/curtailment/handler_test.go
```diff
@@ -76,6 +104,61 @@ func siteScopeJSON(t *testing.T, siteID int64) []byte {
...
+		selectedSiteIDs = append([]int64(nil), scope.SiteIDs...)
+	case coverageComplete && len(currentMemberSiteIDs) > 0:
+		currentMemberSiteIDs = append([]int64(nil), currentMemberSiteIDs...)
+	default:
+		minerScopeUnbounded = true
+	}
+	fanSiteIDs := append([]int64(nil), event.FacilityFanSiteIDs...)
+	fanScopeUnbounded := len(event.FacilityFanDeviceIDs) > 0 && len(fanSiteIDs) != len(event.FacilityFanDeviceIDs)
+	event.AuthorizationEnvelopeJSON = testAuthorizationEnvelopeJSON(
+		selectedSiteIDs,
```
rl-block, line 153: Should we perhaps add handler-level cases with a non-empty malformed envelope? Because this helper synthesizes missing envelopes, List and ForceRelease do not currently prove that corrupt persisted data returns Internal and prevents mutation.

server/internal/domain/curtailment/authorization_envelope.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	for _, field := range authorizationEnvelopeFields {
+		if _, ok := fields[field]; !ok {
+			return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("is missing %q", field)
+		}
+	}
+
+	var envelope models.AuthorizationEnvelope
+	if err := json.Unmarshal(raw, &envelope); err != nil {
+		return models.AuthorizationEnvelope{}, invalidAuthorizationEnvelope("contains an invalid field: %v", err)
```
rl-block, line 51: The parser says null fields fail closed, but `json.Unmarshal` accepts `null` into these non-pointer booleans as `false`. Perhaps reject raw `null` before unmarshalling and add cases for both unbounded flags?

server/internal/handlers/curtailment/handler.go
```diff
@@ -443,6 +444,84 @@ type scopeResourceContextRequirements struct {
...
+
+func requireAuthorizationEnvelopePermissions(ctx context.Context, permission string, raw []byte) error {
+	requirements, err := authorizationEnvelopeResourceContextRequirements(raw)
+	if err != nil {
+		return err
+	}
+	if err := requireResourceContextPermissions(ctx, permission, requirements.resource); err != nil {
+		return err
+	}
+	return requireResourceContextPermissions(ctx, authz.PermSiteRead, requirements.facilityFanRead)
```
rl-block, line 516: Probably worth adding a test that isolates this `site:read` check after curtailment permission succeeds? Existing cases appear to grant or deny both permissions together, so this separate fan-visibility gate could regress unnoticed; a fan-unbounded case might be worth covering too.

server/internal/handlers/curtailment/handler.go
```diff
@@ -118,7 +115,7 @@ func (h *Handler) PreviewCurtailmentPlan(ctx context.Context, req *connect.Reque
 }
 
 func (h *Handler) StartCurtailment(ctx context.Context, req *connect.Request[pb.StartCurtailmentRequest]) (*connect.Response[pb.StartCurtailmentResponse], error) {
-	info, err := middleware.RequirePermission(ctx, authz.PermCurtailmentManage, authz.ResourceContext{})
+	info, err := requireScopedPermissionCapability(ctx, authz.PermCurtailmentManage)
```
rl-block, line 118: Possible to have replay lookup happen before the following live selector and fan checks? After a miner or fan moves or is deleted, a site-scoped retry can fail even though replay later treats the persisted row and envelope as authoritative. Maybe live authorization should run only when no replay exists.

rl-block (approved):

PR #959, opened 2026-08-21
fix(driver): build the sim miner image once and keep it

mcharles-square:
## Summary

`TestDiscoverDevice_MultipleSimMiners` fails intermittently while building the sim miner container:

```
create container: build image: failed to get digest sha256:...:
open /var/lib/docker/image/overlay2/imagedb/content/sha256/...: no such file or directory
```

The digest is different on every occurrence, because it tracks the current build cache rather than a fixed image.

## Cause

`startSimMiner` built from the Dockerfile on every call. Three things combine:

1. Every build uses the same context and Dockerfile, so the build cache resolves all of them to **one image digest**.
2. testcontainers tags each build with a fresh UUID when `Repo`/`Tag` are unset (`container.go:302` `GetRepo`), so several tags point at that single digest.
3. `Terminate` removes a built image unless `KeepImage` is set, and removes it **by ID** (`docker.go:318`), so tearing down one container deletes the digest its siblings still reference.

`TestDiscoverDevice_WithSimMiner` and `TestDiscoverDevice_SchemeNegotiation` run first and each defer a `Terminate`. `TestDiscoverDevice_MultipleSimMiners` then builds twice in a loop; the builds are pure cache hits onto that same digest, and when an earlier removal is still in flight the imagedb entry disappears between build resolution and container create.

## Fix

Build the image once per test binary under a fixed tag with `KeepImage: true`, and start every container from that tag. No production code changes.

## Why now

The race is timing-dependent, so identical code passes here and fails elsewhere. On `main` the package runs ~167s on this repo's runners and passes; on a downstream repo running the same commit on slower runners it took ~185s and failed 4 out of 4 times. This repo is winning the race by margin, not by correctness. Grepping confirms no `KeepImage` anywhere in the tree today.

## Verification

`plugin/proto/internal/driver` passes locally against Docker (60s), `golangci-lint` reports 0 issues, and the built image survives the run:

```
$ docker images proto-fleet-fake-proto-rig --format "{{.Repository}}:{{.Tag}} {{.Size}}"
proto-fleet-fake-proto-rig:driver-tests 21MB
```

Note `plugin/proto/tests` fails locally for me with `read unix ->docker.sock: use of closed network connection`; that reproduces on unmodified `main` (it uses the BuildKit builder over a Docker Desktop socket) and is unrelated to this change. CI covers that package.

### DIFF
--- a/plugin/proto/internal/driver/discovery_test.go
+++ b/plugin/proto/internal/driver/discovery_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/moby/moby/client"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -49,8 +50,11 @@ var (
 // a Docker daemon (parallel packages, or a public and a downstream checkout on
 // one machine) would otherwise retag the same name and start containers from
 // each other's revision.
+//
+// The suffix is random rather than the PID, because binaries in separate PID
+// namespaces can share a daemon through a mounted socket and collide.
 func simMinerImageTag() string {
-	return fmt.Sprintf("driver-tests-%d", os.Getpid())
+	return "driver-tests-" + uuid.NewString()
 }
 
 // simMinerImage builds the fake-proto-rig image on first use and returns the
--- a/plugin/proto/internal/driver/discovery_test.go
+++ b/plugin/proto/internal/driver/discovery_test.go
@@ -38,7 +38,11 @@ type simMinerContainer struct {
 // still in flight, and container creation fails with:
 //
 //	failed to get digest sha256:...: open /var/lib/docker/image/overlay2/imagedb/content/sha256/...: no such file or directory
-const simMinerImageRepo = "proto-fleet-fake-proto-rig"
+//
+// The localhost registry qualifier marks this as a local image and prevents
+// TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX from rewriting the tag between the
+// direct BuildImage call and GenericContainer.
+const simMinerImageRepo = "localhost:5000/proto-fleet-fake-proto-rig"
 
 var (
 	simMinerImageOnce sync.Once
@@ -98,10 +102,21 @@ func TestMain(m *testing.M) {
 	code := m.Run()
 
 	if simMinerImageRef != "" {
-		if provider, err := testcontainers.NewDockerProvider(); err == nil {
-			_, _ = provider.Client().ImageRemove(context.Background(), simMinerImageRef, client.ImageRemoveOptions{})
+		// The test timeout alarm no longer protects cleanup after m.Run, and the
+		// Docker client has no default HTTP timeout. Keep cleanup best-effort: a
+		// daemon disappearing after successful tests should warn, not fail them.
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+
+		provider, err := testcontainers.NewDockerProvider()
+		if err != nil {
+			_, _ = fmt.Fprintf(os.Stderr, "warning: create Docker provider for sim miner image cleanup: %v\n", err)
+		} else {
+			if _, err := provider.Client().ImageRemove(ctx, simMinerImageRef, client.ImageRemoveOptions{}); err != nil {
+				_, _ = fmt.Fprintf(os.Stderr, "warning: remove retained sim miner image %s: %v\n", simMinerImageRef, err)
+			}
 			_ = provider.Close()
 		}
+		cancel()
 	}
 
 	os.Exit(code)
--- a/plugin/proto/internal/driver/discovery_test.go
+++ b/plugin/proto/internal/driver/discovery_test.go
@@ -2,7 +2,9 @@ package driver
 
 import (
 	"context"
+	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


plugin/proto/internal/driver/discovery_test.go
```diff
@@ -21,17 +23,73 @@ type simMinerContainer struct {
...
+			return
+		}
+		defer provider.Close()
+
+		req := testcontainers.ContainerRequest{
+			FromDockerfile: testcontainers.FromDockerfile{
+				Context:    "../../../..",
+				Dockerfile: "server/fake-proto-rig/Dockerfile",
+				Repo:       simMinerImageRepo,
+				Tag:        simMinerImageTag,
```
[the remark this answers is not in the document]
mcharles-square: Valid, fixed in f8c861d (and hardened in 1b8e7f6).

A fixed tag is global to the daemon rather than scoped to the `sync.Once`, and this is not hypothetical here: the public repo and a downstream internal fork are routinely checked out on the same machine, so the two can hold genuinely different revisions of `fake-proto-rig`.

One correction to the mechanism, since it affects how much this matters: a clobbered tag makes a container run the *wrong revision*, it does not reproduce the crash this PR fixes. A retag leaves the previous digest dangling but alive, and `KeepImage` means nothing deletes it mid-run. So this is a correctness hazard for concurrent checkouts, not a second instance of the imagedb race.

I used a unique tag rather than an immutable image ID because `BuildImage` returns a tag and testcontainers image handling is written around named references. That needs cleanup which `KeepImage` suppresses, so there is now a `TestMain` that drops the tag after `m.Run()`; that removal cannot disturb a concurrent binary, for the reason in my reply to the Copilot comment.

plugin/proto/internal/driver/discovery_test.go
```diff
@@ -21,17 +25,95 @@ type simMinerContainer struct {
...
+	simMinerImageRef  string
+	simMinerImageErr  error
+)
+
+// simMinerImageTag scopes the tag to this test binary. Two invocations sharing
+// a Docker daemon (parallel packages, or a public and a downstream checkout on
+// one machine) would otherwise retag the same name and start containers from
+// each other's revision.
+func simMinerImageTag() string {
+	return fmt.Sprintf("driver-tests-%d", os.Getpid())
```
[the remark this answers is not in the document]
mcharles-square: Valid, fixed in 1b8e7f6. PIDs are only unique within a namespace, and a dev container or DinD setup mounting the host socket can easily produce two binaries with the same PID against one daemon. The tag is now `driver-tests-<uuid>`, which also promotes `github.com/google/uuid` to a direct dependency in `plugin/proto/go.mod`.

plugin/proto/internal/driver/discovery_test.go
```diff
@@ -21,17 +23,73 @@ type simMinerContainer struct {
...
+		defer provider.Close()
+
+		req := testcontainers.ContainerRequest{
+			FromDockerfile: testcontainers.FromDockerfile{
+				Context:    "../../../..",
+				Dockerfile: "server/fake-proto-rig/Dockerfile",
+				Repo:       simMinerImageRepo,
+				Tag:        simMinerImageTag,
+				KeepImage:  true,
+			},
```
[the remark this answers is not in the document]
mcharles-square: I dont think this one holds, and I checked it against the daemon rather than reasoning about it.

The claim is that `plugin/proto/tests` terminating its container can delete the digest behind this PRs retained tag. Terminate removes the built image *by tag*, not by ID (`docker.go:358`, `ImageRemove(ctx, c.Image, ...)` where `c.Image` is `repo:tag`). Docker only deletes the underlying digest when the tag being removed is its last reference; when another tag shares the digest, a forced removal just untags. Demonstrated:

```
$ docker build -t tagtest:aaa .   # sha256:9efdc97cb814...
$ docker tag tagtest:aaa tagtest:bbb
$ docker rmi -f tagtest:bbb
$ docker image inspect tagtest:aaa   # => still there, digest intact
```

So the retained tag is exactly what makes the digest safe. If `tests` builds to the same digest, this PRs tag is a second reference, and the `tests` teardown can only drop its own throwaway tag. That is also the precise reason the bug existed before: each build got a UUID tag that was the digests *only* reference, so removing it took the digest with it.

Worth noting the inverse too: `tests` is not exposed to this race on its own, since it builds one container per test and never has a sibling depending on a digest it is tearing down. Giving it the same treatment would be churn without a failure to point at, so Id rather leave it until it actually breaks.

Marking resolved on that basis, happy to reopen if you can point at a sequence where a shared digest loses its last reference while a live tag still names it.

plugin/proto/internal/driver/discovery_test.go
```diff
@@ -21,17 +26,98 @@ type simMinerContainer struct {
...
+// Nothing else deletes it, since KeepImage suppresses Terminate's cleanup.
+// Removing by tag only untags when another tag shares the digest, so this
+// cannot pull the image out from under a concurrent test binary.
+func TestMain(m *testing.M) {
+	code := m.Run()
+
+	if simMinerImageRef != "" {
+		if provider, err := testcontainers.NewDockerProvider(); err == nil {
+			_, _ = provider.Client().ImageRemove(context.Background(), simMinerImageRef, client.ImageRemoveOptions{})
+			_ = provider.Close()
```
rl-block, line 117: Should we bound this cleanup and surface failures? It runs after `m.Run`, when Go's test timeout alarm has stopped, and the Docker client has no default HTTP timeout, so `context.Background()` can leave the process hanging on a wedged daemon. The ignored provider and removal errors also let a passing run leak its retained image; perhaps a short deadline plus logging and a nonzero exit for unexpected cleanup failures would keep this path visible.

plugin/proto/internal/driver/discovery_test.go
```diff
@@ -21,17 +26,98 @@ type simMinerContainer struct {
...
 
+	image, err := simMinerImage(ctx)
+	require.NoError(t, err, "Failed to build sim miner image")
+
 	req := testcontainers.ContainerRequest{
-		FromDockerfile: testcontainers.FromDockerfile{
-			Context:    "../../../..",
-			Dockerfile: "server/fake-proto-rig/Dockerfile",
-		},
+		Image:        image,
```
rl-block, line 135: Worth keeping this locally built reference from going through the hub-prefix substitutor? `BuildImage` creates the unprefixed tag, but `GenericContainer` applies `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX`, so configured hosts may inspect or pull a different random tag. A registry-qualified local repo, or another substitution bypass, plus a prefix-specific test could keep both paths aligned.

rl-block (approved):

plugin/proto/internal/driver/discovery_test.go
mcharles-square: Valid in part, addressed in 11c29a8.

The unbounded context is a real problem: after `m.Run()` the Go test timeout alarm is no longer protecting cleanup, and the Docker client has no default HTTP timeout. Cleanup now has a 10-second deadline, and provider/removal failures are printed to stderr rather than silently ignored.

I intentionally did **not** make cleanup failure change the test exit code. This image is test-only and uniquely tagged; if the daemon disappears after every assertion passes, the only consequence is a leaked local tag (and CI runners are ephemeral). Turning that into a failed suite would trade a visible, bounded cleanup warning for an infrastructure flake. If we later see leaks accumulating on persistent runners despite the warning, we can tighten that policy with evidence.

plugin/proto/internal/driver/discovery_test.go
mcharles-square: Valid, fixed in 11c29a8. `BuildImage` is called directly with no provider-added substitutors, while `GenericContainer` later appends the configured hub substitutor, so the two paths could indeed disagree.

The repo is now `localhost:5000/proto-fleet-fake-proto-rig`. The port is intentional: this testcontainers version does not recognize bare `localhost/` as a registry, but does recognize `localhost:5000/` and skips the hub-prefix rewrite. Docker still treats the already-present tag as local and does not contact a registry.

I exercised the prefix path directly:

```
TESTCONTAINERS_RYUK_DISABLED=true \
TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.example.invalid/mirror \
go test ./internal/driver/ -run TestDiscoverDevice_MultipleSimMiners -count=1
# ok (18.883s)
```

(`RYUK_DISABLED` only isolates this image check; otherwise the deliberately invalid prefix correctly rewrites Ryuk too.) I did not add a permanent prefix-specific test because that would retest testcontainers own documented registry detection in an expensive Docker integration test; the direct CI-equivalent exercise is sufficient for this one-line invariant.

PR #964, opened 2026-08-24
Enable fixed-kW building, rack, and group curtailments

negarn:
Reviewable diff: +44/-35 across 8 files (excludes generated, test, and story files).

## Summary

Operators can now run and test fixed-kW curtailments against Buildings, Racks, and Groups. These starts resolve the selected topology once, freeze the selected miners as concrete targets, and use the existing open-loop curtailment lifecycle. Full-fleet topology execution and topology-backed automation remain disabled while their topology-following lifecycle work continues under #909.

## How it works

The New curtailment and response-profile modals classify the terminal scope through the shared curtailment scope model. Fixed-kW topology scopes can proceed through Run or Test, while the modal still blocks full-fleet topology execution and the automation UI continues to exclude topology-scoped profiles.

The client submits the existing typed Building, Rack, or Group selectors. The handler requires organization-wide `curtailment:manage`, then the domain selector resolves eligible miners and freezes the selected identifiers. During persistence, the SQL store locks and re-resolves the topology, verifies every frozen target is still a member, creates the authorization envelope, and writes the event and targets atomically. The existing reconciler then dispatches those concrete targets without treating the event as a topology-following watcher.

## Diagrams

```mermaid
flowchart TD
  A["New curtailment or profile Test"] --> B["Canonical Building, Rack, or Group scope"]
  B --> C{"Curtailment mode"}
  C --> D["Fixed kW"]
  C --> E["Full fleet"]
  E --> F["Blocked until topology-following lifecycle lands"]
  D --> G["Start RPC and organization-wide authorization"]
  G --> H["Resolve and freeze selected miners"]
  H --> I["Transactionally revalidate topology and persist envelope, event, and targets"]
  I --> J["Existing open-loop reconciler dispatch"]
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/protoFleet/api/curtailmentScopes.ts` | Centralizes Building/Rack/Group scope classification. | Keeps manual execution and automation policy aligned on the same scope vocabulary. |
| `client/src/protoFleet/features/energy/CurtailmentStartModal.tsx` | Enables fixed-kW topology Run/Test while retaining the full-fleet gate and mode-specific copy. | Defines the operator-facing rollout boundary. |
| `client/src/protoFleet/features/settings/components/Curtailment/` | Renames profile readiness to automation readiness and keeps topology profiles out of automation. | Separates newly supported manual execution from unsupported automated execution. |
| `server/internal/domain/curtailment/service.go` | Allows topology Start for non-full-fleet modes and continues rejecting full-fleet topology starts. | Establishes frozen versus topology-following lifecycle behavior at the domain boundary. |
| `server/internal/domain/stores/sqlstores/curtailment_topology_start_integration_test.go` | Covers transactional topology-envelope persistence and membership-drift rejection. | Verifies a topology change cannot widen or persist a stale frozen start. |
| `docs/plans/2026-08-11-curtailment-building-rack-and-group-targets-plan.md` | Records the staged fixed-kW Run/Test rollout. | Keeps the implementation boundary and remaining #909 work explicit. |

- Enable only fixed-kW topology execution, which can freeze concrete targets, instead of exposing full-fleet scopes before topology-following ownership and restoration exist.
- Continue requiring organization-wide `curtailment:manage` for topology starts until dispatch-time scoped reauthorization is implemented.
- Revalidate topology inside the event-and-target transaction instead of trusting the earlier selector result, so membership drift fails closed.
- Model profile readiness specifically as automation readiness instead of using one flag for both manual and automated execution.

## Testing & validation

- Built the ProtoFleet production client with `npm run build:protoFleet`.
- Passed TypeScript typechecking and targeted ESLint/Prettier checks for the changed client code.
- Passed 266 tests across the six affected frontend suites; re-ran the 125 directly affected tests after consolidating topology classification.
- Passed the changed curtailment domain and handler topology Start tests.
- Compiled the curtailment domain, handler, reconciler, and SQL-store Go packages.
- The new SQL integration test compiles but was not executed locally because the configured test Postgres rejected the local credentials; CI will run it with the repository test database.

Refs #909

### DIFF
--- a/server/internal/domain/curtailment/reconciler/dispatch_authorization.go
+++ b/server/internal/domain/curtailment/reconciler/dispatch_authorization.go
@@ -24,6 +24,7 @@ func (r *Reconciler) authorizeTopologyCurtailDispatch(
 	ctx context.Context,
 	ev *models.Event,
 	targets []*models.Target,
+	knownUnsentDeviceIdentifiers []string,
 ) bool {
 	if ev.ScopeType != models.ScopeTypeMixed {
 		return true
@@ -39,7 +40,9 @@ func (r *Reconciler) authorizeTopologyCurtailDispatch(
 			ctx,
 			ev.OrgID,
 			ev.EventUUID,
-			interfaces.BeginRestoreTransitionParams{},
+			interfaces.BeginRestoreTransitionParams{
+				KnownUnsentDeviceIdentifiers: knownUnsentDeviceIdentifiers,
+			},
 		); restoreErr != nil {
 			slog.Error("curtailment reconciler: failed to restore after topology dispatch authorization failure",
 				"event_id", ev.ID, "event_uuid", ev.EventUUID, "error", restoreErr)
--- a/server/internal/domain/curtailment/reconciler/dispatch_authorization.go
+++ b/server/internal/domain/curtailment/reconciler/dispatch_authorization.go
@@ -31,19 +31,23 @@ func (r *Reconciler) authorizeTopologyCurtailDispatch(
 	if ev.ScopeType != models.ScopeTypeMixed {
 		return false, true
 	}
-	reject := func(reason string, cause error) bool {
-		r.topologyDispatchRejects.Store(ev.EventUUID, reason)
-		attrs := []any{"event_id", ev.ID, "event_uuid", ev.EventUUID, "reason", reason}
+	reject := func(rejection topologyDispatchRejection, cause error) bool {
+		r.topologyDispatchRejects.Store(ev.EventUUID, rejection)
+		attrs := []any{"event_id", ev.ID, "event_uuid", ev.EventUUID, "reason", rejection.reason}
 		if cause != nil {
 			attrs = append(attrs, "error", cause)
 		}
 		slog.Error("curtailment reconciler: topology dispatch authorization failed", attrs...)
+		releaseUnsent := knownUnsentDeviceIdentifiers
+		if rejection.preserveRestoreOwnership {
+			releaseUnsent = nil
+		}
 		if _, restoreErr := r.store.BeginRestoreTransition(
 			ctx,
 			ev.OrgID,
 			ev.EventUUID,
 			interfaces.BeginRestoreTransitionParams{
-				KnownUnsentDeviceIdentifiers: knownUnsentDeviceIdentifiers,
+				KnownUnsentDeviceIdentifiers: releaseUnsent,
 			},
 		); restoreErr != nil {
 			slog.Error("curtailment reconciler: failed to restore after topology dispatch authorization failure",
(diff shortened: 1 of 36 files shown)


server/internal/domain/curtailment/reconciler/dispatch_authorization.go
```diff
@@ -0,0 +1,183 @@
...
+		return reject("topology scope resolver is not configured", nil)
+	}
+	coverage, err := topologyStore.ResolveCurtailmentTopologyScope(ctx, filter)
+	if err != nil {
+		return reject("reload topology coverage", err)
+	}
+	if !topologyCoverageWithinEnvelope(coverage, envelope) {
+		return reject("current topology exceeds the persisted authorization envelope", nil)
+	}
+	candidates, err := r.store.ListCandidates(ctx, filter)
```
rl-block: Looks like this re-expands the full selector for every dispatch batch. With 10,000 targets and a 100-miner batch, the singleton reconciler can repeat this work about 100 times. Perhaps we could use a prepared query that returns aggregate coverage plus membership for only `dispatchSet`, while keeping the live permission check before each command?

server/internal/domain/curtailment/reconciler/dispatch_authorization.go
```diff
@@ -0,0 +1,183 @@
...
+		slog.Error("curtailment reconciler: topology dispatch authorization failed", attrs...)
+		if _, restoreErr := r.store.BeginRestoreTransition(
+			ctx,
+			ev.OrgID,
+			ev.EventUUID,
+			interfaces.BeginRestoreTransitionParams{},
+		); restoreErr != nil {
+			slog.Error("curtailment reconciler: failed to restore after topology dispatch authorization failure",
+				"event_id", ev.ID, "event_uuid", ev.EventUUID, "error", restoreErr)
+		}
```
rl-block, line 57: If `BeginRestoreTransition` fails, the rejection is only logged and the event stays pending; a later tick could resume Curtail if the rejected condition clears. Should we persist a sticky dispatch-rejected state or intent so later ticks only retry restoration?

server/internal/domain/curtailment/reconciler/dispatch_authorization.go
```diff
@@ -0,0 +1,183 @@
...
+		if _, ok := currentMembers[target.DeviceIdentifier]; !ok {
+			return reject("claimed miner is no longer in the persisted topology scope", nil)
+		}
+	}
+
+	if r.permissions == nil {
+		return reject("dispatch permission resolver is not configured", nil)
+	}
+	effective, err := r.permissions.LoadEffective(ctx, latest.CreatedByUserID, latest.OrgID)
```
rl-block: `coverage` and `candidates` come from separate snapshots, and membership is checked only by identifier. A group member could move to an out-of-envelope site between these reads and still be authorized against stale site coverage. Worth combining coverage and batch membership in one transactionally consistent store operation and add a barrier-based race test?

server/internal/domain/curtailment/reconciler/dispatch_authorization.go
```diff
@@ -0,0 +1,183 @@
...
+		if cause != nil {
+			attrs = append(attrs, "error", cause)
+		}
+		slog.Error("curtailment reconciler: topology dispatch authorization failed", attrs...)
+		if _, restoreErr := r.store.BeginRestoreTransition(
+			ctx,
+			ev.OrgID,
+			ev.EventUUID,
+			interfaces.BeginRestoreTransitionParams{},
+		); restoreErr != nil {
```
rl-block: 🤖 If an authorization read fails because `ctx` reached its deadline, this restore attempt reuses the already-canceled context and will fail immediately. Could we perhaps use a short bounded context derived from `context.WithoutCancel(ctx)` here and add a cancellation-path test?

server/internal/domain/curtailment/reconciler/reconciler.go
```diff
@@ -704,6 +705,9 @@ func (r *Reconciler) dispatchCurtailBatch(ctx context.Context, ev *models.Event,
 	if !r.eventStillDispatchable(ctx, ev) {
 		return false
 	}
+	if !r.authorizeTopologyCurtailDispatch(ctx, ev, dispatchSet) {
+		return false
+	}
```
rl-block: Should we perhaps call `eventStillDispatchable` again immediately before `cmd.Curtail` and cover this with a race test?

server/internal/domain/curtailment/reconciler/reconciler_test.go
```diff
@@ -899,6 +922,113 @@ func TestReconciler_PendingDispatchesCurtail(t *testing.T) {
...
+		{
+			name:       "member left scope restores without dispatch",
+			candidates: nil,
+			effective:  allowed,
+		},
+		{
+			name:       "creator permission revoked restores without dispatch",
+			candidates: []*models.Candidate{{DeviceIdentifier: "miner-1"}},
+			effective:  authz.NewEffectivePermissions(nil),
+		},
```
rl-block, line 1053: This table covers clean success, missing membership, and revoked permission, but doesn't seem to cover the new reload, parse, resolver, or permission-read errors. Since these branches are the fail-closed safety boundary, maybe add table cases that assert zero Curtail calls and a restore attempt for each dependency or malformed-data error.

rl-block (approved):

PR #967, opened 2026-08-25
Add topology curtailment store primitives (1/2)

negarn:
Reviewable diff: +896/-135 across 5 files (excludes generated, test, and story files).

## Summary

This PR adds the persistence foundation for topology-scoped FULL_FLEET curtailment. It introduces deterministic logical-scope reservations, transactional target admission, topology departure transitions, and guarded bulk readiness updates without enabling the new reconciliation behavior by itself.

**Stack:** this PR → #968 adds the reconciliation and API behavior. #968 is based directly on this branch; the shared drill-down UI and remaining frontend/E2E work stay tracked in #909.

## How it works

The store serializes curtailment scope changes per organization and resolves older logical scopes against each candidate before allowing ownership. New targets are inserted or reopened under the same event lock, while cross-event conflicts remain no-ops for a later retry.

For live topology changes, the store can move departed targets into the existing restore phase while leaving the parent watcher active. All-paired readiness changes carry the target state observed by the reconciler into expected-state guards, so a stale snapshot cannot overwrite a concurrent lifecycle transition. Device and pairing-row locks serialize departure checks with pairing updates, including the first pairing insert. Dispatch fencing resolves membership, authorization coverage, and event state in one database transaction before the caller sends physical commands.

## Diagrams

```mermaid
flowchart TD
  A["Admission request"] --> B["Lock organization scope"]
  B --> C["Resolve earlier logical reservations"]
  C --> D["Lock parent event"]
  D --> E["Insert or reopen eligible targets"]
  E --> F["Return claimed rows"]
  C --> G["Skip conflicting devices"]
```

```mermaid
flowchart LR
  A["Topology member leaves"] --> B["Lock event and topology coverage"]
  B --> C["No Curtail attempt"]
  B --> D["Curtail may have been sent"]
  C --> E["Release target"]
  D --> F["Create restore obligation"]
```

## Areas of the code involved

| --- | --- | --- |
| `server/sqlc/queries/curtailment.sql` | Adds logical reservation, admission claim, dispatch fence, departure restore, and guarded readiness queries. | Defines locking, ownership, and race behavior. |
| `server/sqlc/queries/device.sql` | Makes pairing inserts take the same device lock as topology departure checks. | Prevents a first pairing from racing a stale restore decision. |
| `server/internal/domain/stores/interfaces/curtailment.go` | Exposes topology restore and expected-state readiness contracts. | Establishes the boundary consumed by the stacked reconciler PR. |
| `server/internal/domain/stores/sqlstores/curtailment.go` | Implements the new transactional store operations and payload conversion. | Connects domain operations to the SQL invariants. |
| `server/internal/domain/curtailment/reconciler/reconciler_allpaired.go` | Supplies the observed target state to the guarded bulk readiness API. | Makes the new store primitive independently safe before the stacked behavior lands. |
| `server/generated/sqlc/` | Regenerated SQL bindings — skip. | Mechanical output from the query changes. |
| SQL store and reconciler tests | Covers topology scope ownership, dispatch fencing, all-paired guards, and claim results. | Verifies database-level concurrency and caller behavior. |

- Serialize logical reservations with an organization-scoped advisory lock instead of allowing overlapping watchers to race; persisted event order becomes deterministic at the cost of per-organization admission serialization.
- Keep ownership conflicts as no-ops instead of errors so active watchers can retry after the earlier owner resolves.
- Use expected event, target, and desired-state guards instead of unconditional updates so stale reconciliation work cannot clobber concurrent Stop or restore transitions.
- Lock device rows before optional pairing rows and make pairing inserts honor the same order, covering devices with no pairing row without weakening existing-row serialization.
- Return inserted and reopened CTE rows directly instead of rereading them through the statement snapshot, ensuring successful claims are visible to the caller.

## Testing & validation

- `go test ./internal/domain/curtailment/reconciler -run 'TestReconciler_AllPairedPolicy' -count=1`
- `golangci-lint run ./internal/domain/stores/... ./generated/sqlc/...`
- SQL store package compilation and non-database unit coverage
- Focused first-pairing concurrency test added; local execution requires the repository database harness and is delegated to CI.
- `git diff --check`
- Database-backed integration tests could not run locally because the local Postgres credentials do not match the repository test harness; they are delegated to CI.

Refs #909

### DIFF
--- a/server/internal/domain/stores/sqlstores/curtailment.go
+++ b/server/internal/domain/stores/sqlstores/curtailment.go
@@ -2266,7 +2266,7 @@ func (s *SQLCurtailmentStore) WithCurtailmentTopologyRestoreDispatchFence(
 	ctx context.Context,
 	event *models.Event,
 	dispatchDeviceIdentifiers []string,
-	command func(interfaces.CurtailmentTopologyDispatchFenceSnapshot) error,
+	command func(interfaces.CurtailmentTopologyRestoreDispatchFenceSnapshot) error,
 ) error {
 	if event == nil || command == nil {
 		return fleeterror.NewInvalidArgumentError("topology restore dispatch fence requires an event and command")
@@ -2303,9 +2303,27 @@ func (s *SQLCurtailmentStore) WithCurtailmentTopologyRestoreDispatchFence(
 		if err != nil {
 			return err
 		}
-		return command(interfaces.CurtailmentTopologyDispatchFenceSnapshot{
-			Event:    latest,
-			Topology: topology,
+		parkReturnedTargets := func(returnedDeviceIdentifiers []string) error {
+			reason := "restore paused: device returned to topology scope"
+			expectedState := models.TargetStateDispatching
+			desiredActive := models.DesiredStateActive
+			for _, deviceIdentifier := range uniqueSortedStrings(returnedDeviceIdentifiers) {
+				if err := updateCurtailmentTargetState(ctx, q, latest.ID, deviceIdentifier, interfaces.UpdateCurtailmentTargetStateParams{
+					State:                models.TargetStateRestoreFailed,
+					LastError:            &reason,
+					ExpectedEventState:   &latest.State,
+					ExpectedDesiredState: &desiredActive,
+					ExpectedState:        &expectedState,
+				}); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+		return command(interfaces.CurtailmentTopologyRestoreDispatchFenceSnapshot{
+			Event:               latest,
+			Topology:            topology,
+			ParkReturnedTargets: parkReturnedTargets,
 		})
 	})
 }
@@ -2899,7 +2917,17 @@ func (s *SQLCurtailmentStore) RecoverTerminalFanState(
 }
 
 func (s *SQLCurtailmentStore) UpdateTargetState(ctx context.Context, eventID int64, deviceIdentifier string, params interfaces.UpdateCurtailmentTargetStateParams) error {
-	rows, err := s.GetQueries(ctx).UpdateCurtailmentTargetState(ctx, sqlc.UpdateCurtailmentTargetStateParams{
+	return updateCurtailmentTargetState(ctx, s.GetQueries(ctx), eventID, deviceIdentifier, params)
+}
+
+func updateCurtailmentTargetState(
+	ctx context.Context,
+	q sqlc.Querier,
+	eventID int64,
+	deviceIdentifier string,
+	params interfaces.UpdateCurtailmentTargetStateParams,
+) error {
+	rows, err := q.UpdateCurtailmentTargetState(ctx, sqlc.UpdateCurtailmentTargetStateParams{
 		CurtailmentEventID:        eventID,
 		DeviceIdentifier:          deviceIdentifier,
 		State:                     string(params.State),
--- a/server/internal/domain/stores/sqlstores/curtailment.go
+++ b/server/internal/domain/stores/sqlstores/curtailment.go
@@ -2290,18 +2290,25 @@ func (s *SQLCurtailmentStore) WithCurtailmentTopologyRestoreDispatchFence(
 			return interfaces.ErrCurtailmentEventStateRaceLoss
 		}
 		latest := convertEventRow(locked)
-		scope, hasScope, err := domainCurtailment.ScopeFromJSON(latest.ScopeJSON)
-		if err != nil || !hasScope || !domainCurtailment.IsTopologyScope(scope) {
-			return fleeterror.NewFailedPreconditionError("persisted topology scope is no longer valid")
-		}
-		params, err := domainCurtailment.ListCandidatesParamsForScope(scope)
-		if err != nil {
-			return err
-		}
-		params.OrgID = latest.OrgID
-		topology, err := lockCurtailmentTopologyRestoreDispatch(ctx, q, params, identifiers)
-		if err != nil {
-			return err
+		topology := interfaces.CurtailmentTopologyDispatchSnapshot{}
+		if latest.State == models.EventStateRestoring {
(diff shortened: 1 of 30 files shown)


server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -3072,22 +3111,33 @@ func (s *SQLCurtailmentStore) ClaimClosedLoopFullFleetTargets(
...
+			return nil, err
+		}
+		payload, err := buildBulkTargetPayload(available)
+		if err != nil {
+			return nil, fleeterror.NewInternalErrorf("failed to encode curtailment target payload: %v", err)
+		}
 		rows, err := q.ClaimClosedLoopFullFleetTargets(ctx, sqlc.ClaimClosedLoopFullFleetTargetsParams{
 			CurtailmentEventID: eventID,
 			TargetsJsonb:       payload,
 		})
```
rl-block, line 3250: Worth making this claim atomic with reservation-relevant topology changes? Building/rack/group writers can commit between this reservation read and claim, letting a newer event take concrete ownership from an older targetless watcher. Maybe we could lock the relevant selector/device rows, or verify a topology revision, then re-read reservations immediately before the claim. A concurrency test that pauses after the reservation read and moves the device into the older scope could pin the behavior.

server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -3111,24 +3161,162 @@ func (s *SQLCurtailmentStore) ClaimAllPairedPolicyTargets(
...
+				ExpectedEventState: string(event.State),
+				DeviceIdentifiers:  departed,
+			},
+		)
+	})
+	if err != nil {
+		if errors.Is(err, interfaces.ErrCurtailmentEventStateRaceLoss) {
+			return 0, err
+		}
+		return 0, fleeterror.NewInternalErrorf("failed to begin topology target restore: %v", err)
```
rl-block: Could we preserve typed `fleeterror.FleetError`s here before wrapping? The transaction can return `FailedPrecondition`, but this converts it to `Internal`. Maybe we could mirror `ClaimClosedLoopFullFleetTargets` with `errors.As` so callers retain the intended error contract.

server/sqlc/queries/curtailment.sql
```diff
@@ -1003,37 +1356,106 @@ SELECT ((SELECT COUNT(*) FROM reopened) + (SELECT COUNT(*) FROM inserted))::BIGI
...
+          JOIN curtailment_event other_event
+            ON other_event.id = other_target.curtailment_event_id
+          WHERE other_target.device_identifier = target.device_identifier
+            AND other_target.curtailment_event_id <> target.curtailment_event_id
+            AND other_event.state IN ('pending', 'active', 'restoring')
+            AND other_target.state NOT IN ('resolved', 'restore_failed', 'released')
+      )
+  )
+  AND (
+      (target.desired_state = 'curtailed' AND target.state IN ('pending', 'unavailable'))
```
rl-block, line 1489: This guard should probably also account for older logical reservations. An older targetless watcher has no concrete row here, so a newer restore-failed target can requeue first. Maybe we could reuse the earlier-reservation check under the org lock and add an integration case with an older targetless watcher.

server/sqlc/queries/curtailment.sql
```diff
@@ -1208,6 +1630,124 @@ SET desired_state      = 'active',
...
+        ELSE 'active'
+    END,
+    state = CASE
+        WHEN target.state IN ('pending', 'unavailable')
+         AND target.last_dispatched_at IS NULL
+         AND target.curtail_dispatched_at IS NULL
+         AND target.retry_count = 0
+         AND target.restore_started_at IS NULL
+        THEN 'released'
+        ELSE 'pending'
```
rl-block, line 1696: Perhaps worth adding database coverage for both outcomes of this branch. Current coverage does not execute this SQL for an actual departed target, so the immediate-release and conservative-restore paths could regress. Perhaps test one departed undispatched target and one with dispatch history, asserting the resulting state and restore fields.

server/sqlc/queries/curtailment.sql
```diff
@@ -1003,37 +1356,106 @@ SELECT ((SELECT COUNT(*) FROM reopened) + (SELECT COUNT(*) FROM inserted))::BIGI
...
 ) ON TRUE
 WHERE target.curtailment_event_id = locked_event.id
   AND target.device_identifier = t.device_identifier
-  AND target.desired_state = 'curtailed'
-  AND target.state IN ('pending', 'unavailable')
+  AND target.state = t.expected_state
+  AND (
+      target.state <> 'restore_failed'
+      OR NOT EXISTS (
+          SELECT 1
```
rl-block: Could be worth carrying the expected `desired_state` alongside `expected_state`. Curtail and restore can share `pending`/`unavailable` while the event stays active, so a direction flip can still apply a stale opposite-phase update. Adding the desired state to the payload and guard, with race tests in both directions, could close that window.

server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -3111,24 +3161,162 @@ func (s *SQLCurtailmentStore) ClaimAllPairedPolicyTargets(
...
+			return 0, err
+		}
+		payload, err := buildBulkTargetPayload(available)
+		if err != nil {
+			return 0, fleeterror.NewInternalErrorf("failed to encode curtailment target payload: %v", err)
+		}
+		return q.ClaimAllPairedPolicyTargets(ctx, sqlc.ClaimAllPairedPolicyTargetsParams{
+			CurtailmentEventID: eventID,
+			TargetsJsonb:       payload,
+		})
```
rl-block, line 3340: This store call can receive the full all-paired target slice (up to 10,000) and holds the org lock while checking reservations. Perhaps we could reuse `curtailBatchSizeForEvent` before the call so one tick and transaction stay bounded? A max-size test could also verify that only one batch is claimed per tick.

rl-block (approved):

PR #971, opened 2026-08-26
fix(client): refocus password input after weak-password warning

rl-block:
Reviewable diff: +48/-9 across 4 files (excludes generated, test, and story files).

## Summary

When a user submits a weak password, the weak-password dialog offers "Continue anyway" or "Create a stronger password". Choosing the latter previously just closed the dialog, leaving focus nowhere — the user had to click back into the password field. This PR returns focus to the relevant password input in all four flows that show the dialog, and fixes a pre-existing bug in the miner password modal where returning from the dialog rendered the fields empty while the submit button stayed enabled with the invisible stale values.

## How it works

All four flows render the shared `WeakPasswordWarning` dialog, but they fall into two structural cases:

**Form stays mounted during the warning** (account-creation `Authentication`, temporary-password `UpdatePasswordForm`, fleet security settings `Auth`): the dialog overlays the form via a portal, and neither `Dialog` nor `PageOverlay` does focus restoration on close. Each flow now holds a `RefObject` on its password input (via the `inputRef` prop the shared `Input` already supports) and calls `.focus()` in the dialog's `onReturn` handler. The `Input` component's existing password-focus effect places the cursor at the end of the current value, so the user can immediately edit.

**Form unmounts during the warning** (`UpdateMinerPasswordModal` returns the warning *instead of* the modal for animation reasons): there is no input to focus at click time. `onReturn` instead sets a `refocusNewPassword` flag; when the modal remounts, `autoFocus` is handed to the new-password input rather than the current-password input (which keeps `autoFocus` on a fresh open). The same remount previously caused the value-loss bug: `Input` seeds its internal display state from `initValue` only on mount, so remounted fields rendered empty while component state (and the enabled Continue button) still held the typed values. The three inputs now pass their retained state back as `initValue`, making the display consistent with state and with the other three flows.

## Diagrams

```mermaid
flowchart TD
    A["User submits weak password"] --> B["WeakPasswordWarning dialog"]
    B -->|"Continue anyway"| C["Submit proceeds"]
    B -->|"Create a stronger password"| D{"Is the form still mounted?"}
    D -->|"Yes: Authentication, UpdatePasswordForm, settings Auth"| E["onReturn closes dialog and calls focus() via inputRef"]
    D -->|"No: UpdateMinerPasswordModal"| F["onReturn sets refocusNewPassword flag"]
    F --> G["Modal remounts: autoFocus goes to new-password input; initValue re-seeds all field values"]
```

```mermaid
sequenceDiagram
    participant U as User
    participant M as UpdateMinerPasswordModal
    participant W as WeakPasswordWarning
    U->>M: Continue with weak password
    M->>W: unmount modal, render warning
    U->>W: Create a stronger password
    W->>M: onReturn sets refocusNewPassword, modal remounts
    M->>M: inputs re-seed from initValue
    M->>U: new-password input focused, values visible
```

## Areas of the code involved

| --- | --- | --- |
| `client/src/shared/components/Setup/Authentication.tsx` | Ref on password input, focused in `onReturn` | Shared component used by account creation in both ProtoOS and ProtoFleet onboarding |
| `client/src/protoFleet/features/auth/components/UpdatePasswordForm.tsx` | Same ref treatment on "New password" | Temporary-password login flow |
| `client/src/protoFleet/features/settings/components/Auth.tsx` | Same ref treatment inside the settings modal | Fleet security settings password update |
| `client/.../ManageSecurity/UpdateMinerPasswordModal.tsx` | `refocusNewPassword` flag swaps `autoFocus` on remount; `initValue` re-seeds field values | Different mechanism because the modal unmounts during the warning; also fixes the stale-invisible-values bug |
| Four `*.test.tsx` files | Focus and value-persistence tests | Test-only |

- **Direct `.focus()` in `onReturn` over an effect-based approach** for the three mounted flows: the dialog does no focus management of its own, so a synchronous call is sufficient and avoids extra state.
- **`autoFocus` swap over ref-plus-effect in the miner modal**: the input does not exist when the button is clicked, so a flag consumed at remount is the simplest correct mechanism; a ref would require an effect watching a flag anyway.
- **`initValue` re-seeding over clearing state on return**: preserving values matches the behavior of the other three flows (where the form never unmounts) and keeps the current-password field intact so the user only retypes the field they came back to change.

## Testing & validation

- New vitest cases in all four component test files: focus lands on the password input after "Create a stronger password", and (miner modal) all three field values survive the warning round-trip. 86 tests across the four files pass.
- The miner modal test file's `Input` mock now forwards `initValue` as `defaultValue` to mirror the real component's seed-on-mount behavior.
- Not covered: E2E; verified via unit tests and the lefthook typecheck on push.

### DIFF
--- a/client/src/protoFleet/features/auth/components/UpdatePasswordForm.test.tsx
+++ b/client/src/protoFleet/features/auth/components/UpdatePasswordForm.test.tsx
@@ -170,4 +170,20 @@ describe("UpdatePasswordForm", () => {
       getByText("You logged in with a temporary password. Enter your new password to continue."),
     ).toBeInTheDocument();
   });
+
+  it("refocuses the new password input when the user chooses to create a stronger password", async () => {
+    const { getByLabelText, getByText, findByText, queryByText } = render(
+      <UpdatePasswordForm onSubmit={mockOnSubmit} />,
+    );
+
+    fireEvent.change(getByLabelText("New password"), { target: { value: "aaaaaaaa" } });
+    fireEvent.change(getByLabelText("Confirm password"), { target: { value: "aaaaaaaa" } });
+    fireEvent.click(getByText("Continue"));
+
+    const returnButton = await findByText("Create a stronger password");
+    fireEvent.click(returnButton);
+
+    expect(queryByText("Create a stronger password")).not.toBeInTheDocument();
+    expect(getByLabelText("New password")).toHaveFocus();
+  });
 });
--- a/client/src/protoFleet/features/auth/components/UpdatePasswordForm.tsx
+++ b/client/src/protoFleet/features/auth/components/UpdatePasswordForm.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from "react";
+import { RefObject, useCallback, useRef, useState } from "react";
 import Footer from "@/protoFleet/components/Footer";
 import { Alert, Logo } from "@/shared/assets/icons";
 import Button from "@/shared/components/Button";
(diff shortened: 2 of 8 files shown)


flesher (approved):
👍

PR #977, opened 2026-08-27
Version curtailment response profile execution

negarn:
Reviewable diff: +1748/-248 across 29 files (excludes generated, test, and story files).

## Summary

This PR prevents saved curtailment response profiles from being executed after their execution settings have changed. Profile updates, manual preview/start, save-and-run, and automation bindings now carry an opaque profile revision and fail closed when the caller, profile, or automation rule is stale. It also enables topology full-fleet execution and preserves saved priority, tolerance, cooldown, maintenance inclusion, and all-paired settings end to end. It is the response-profile execution-versioning slice of #909; merging it does not complete or close the tracking issue.

## How it works

Response profiles receive a UUID revision, and automation rules persist the exact profile revision they were bound to. Migration 144 stores those values in trigger-maintained companion tables so the previous active binary can keep reading and writing the unchanged base-table shapes while a new passive binary migrates during the standard HA rollout. It drains active profile and rule writers before canonicalizing or backfilling, then holds those locks through trigger installation so no old-binary write can miss or retain stale companion state. The migration canonicalizes existing version-zero profile scopes to schema version 1, preserving legacy site selectors, binds existing rules, and stamps fully owned live automation-event snapshots only when their persisted scope and execution fields still match the profile. Deployed curtailment data can recover the event-create/active-pointer crash window without weakening the runtime revision or scope fence; mismatched events remain unbound and fail closed. Updating execution-affecting profile fields rotates the revision, while metadata-only renames and effective tolerance changes between NULL and zero preserve it.

The client carries that revision through profile updates, profile-derived Preview and Start requests, automation create/update/enable, and the response-profile save-and-run flow. Preview and Start also require execution-contract version 1, so an already-open pre-versioning tab cannot submit copied profile values as an unbound custom plan. Saved priority, fixed-kW tolerance, post-event cooldown, independent maintenance inclusion, and all-paired settings are hydrated into those requests, including building, rack, and group full-fleet execution. The server verifies that profile-derived request values still match the referenced revision, then the SQL store locks the profile revision and automation binding in the event-creation or recurtail transaction before persisting work. A concurrent profile edit or rule rebind therefore returns a failed-precondition response instead of dispatching a mixed configuration.

Settings reloads the latest profile after an optimistic-update conflict or a stale revision prevents saving or enabling an automation rule. Automation modals replace their captured revision with the refreshed value before the operator retries. Saved profiles are paired with server-canonical persisted values, so database-normalized values and the returned revision always describe the same executable configuration; editing an externally created profile also preserves its priority, optional tolerance, cooldown, and independent maintenance inclusion.

## Diagrams

```mermaid
flowchart TD
  A["Response profile editor or automation form"] --> B["Profile ID plus expected revision"]
  B --> C["Execution-version and API validation"]
  C --> D["Profile and automation authorization"]
  D --> E["Transactional profile and rule locks"]
  E --> F["Persist event and concrete targets"]
  E --> G["Reject stale revision or binding"]
```

```mermaid
sequenceDiagram
  participant UI as ProtoFleet
  participant API as Curtailment API
  participant DB as PostgreSQL
  UI->>API: Start with profile ID and expected revision
  API->>API: Compare request values with saved profile
  API->>DB: Lock profile revision and automation binding
  alt Revision and binding still match
    DB-->>API: Event and targets committed
    API-->>UI: Curtailment event
  else Profile or rule changed
    DB-->>API: Failed precondition
    API-->>UI: Reload and retry
  end
```

## Areas of the code involved

| --- | --- | --- |
| `proto/curtailment/v1/` | Adds profile revisions, expected-revision fields, and the required Preview/Start execution-contract version. | Defines the cross-client concurrency contract and rejects stale pre-version clients. |
| `server/migrations/000144_*` and `server/sqlc/queries/` | Persists revisions in companion tables, maintains mixed-version writes with triggers, backfills compatible bindings, and adds revision rotation, optimistic updates, and execution locks. | Establishes the database source of truth and preserves passive-first HA compatibility. |
| `server/internal/domain/curtailment/` | Propagates profile/rule fences through manual execution, automation OFF/recurtail, replay validation, and event snapshots. | Ensures every profile-derived execution uses one coherent configuration. |
| `server/internal/handlers/curtailment/` | Validates revision pairs and exact profile-derived values before invoking domain execution. | Rejects stale or altered requests at the API boundary. |
| `client/src/protoFleet/api/` | Carries revisions through profile and automation APIs and reloads stale profile edits. | Keeps browser state synchronized with server revisions. |
| `client/src/protoFleet/features/energy/` and settings curtailment components | Threads the selected revision and execution fields through Preview, Start, automation forms, topology full-fleet execution, and save-and-run. | Prevents a fresh revision from being paired with stale or incomplete form values. |
| `docs/plans/2026-08-11-curtailment-building-rack-and-group-targets-plan.md` | Reconciles rollout guidance with safe profile/rule revision backfill. | Keeps the implementation contract aligned with supported deployed data. |
| Generated protobuf/sqlc/Fleet CLI files | Regenerated from the proto, SQL, and CLI manifest inputs. | Generated — skip. |

- Rotate opaque UUID revisions only for execution-affecting changes, so harmless profile renames do not force automation rebinding.
- Enforce the final fence inside the SQL transaction instead of relying only on handler reads, covering profile edits and automation rebinds that race event creation.
- Require revisions when creating, updating, or enabling automations, while allowing disable/delete recovery paths without a current profile revision.
- Require execution-contract version 1 on Preview and Start so stale clients fail closed instead of silently taking the custom-plan path.
- Keep response-profile and automation-rule base-table shapes unchanged during the HA mixed-version window; companion tables and triggers preserve old-binary reads and writes without a coordinated cutover.
- Canonicalize existing profile scopes and backfill profile revisions, automation bindings, and scope-and-field-matched replay-eligible live automation snapshots in migration 144; runtime execution remains strict and mismatched events stay unbound.
- Treat server-returned persisted values as canonical for saved profiles while retaining only UI-only duration fields in session state.

## Testing & validation

- `bin/just gen`
- `bin/buf lint`, `bin/just _lint-client`, and `bin/just _lint-server`
- 5 directly affected ProtoFleet Vitest suites after review fixes: 132 tests passed
- `cd client && npx tsc --noEmit`
- `go test ./internal/domain/curtailment ./internal/handlers/curtailment`
- `go test -tags=e2e ./e2e -run '^$'` compiles the versioned curtailment lifecycle fixtures
- 7 focused SQL-store response-profile revision and replay integration tests passed against local PostgreSQL
- Migration bridge, passive-first write compatibility, and terminal-provenance rollback coverage passed from versions 130, 142, and dirty 143
- Migration ordering test verifies active profile and rule writers are drained before canonicalization and revision backfill
- `npm run build:protoFleet`
- `cd server && ../bin/just build`

Part of #909


client/src/protoFleet/features/energy/CurtailmentStartModal.tsx
```diff
@@ -1356,6 +1356,7 @@ function CurtailmentStartModalContent({
     return {
       ...nextValues,
       responseProfileId: customResponseProfileId,
+      responseProfileRevision: undefined,
     };
```
rl-block, line 1360: `confirmForceInclusion` also reaches this helper with an empty pending patch when acknowledging an all-paired saved profile. That clears the profile ID and revision before Start, so could we preserve the binding when confirmation did not apply an edit and add a saved-profile run regression test?

proto/curtailment/v1/curtailment.proto
```diff
@@ -714,6 +724,17 @@ message StartCurtailmentRequest {
...
+    // Set together when starting values copied from a saved response profile.
+    // The server locks this revision in the event-create transaction.
+    int64 response_profile_id = 42 [(buf.validate.field).int64.gte = 0];
+    string expected_response_profile_revision = 43 [(buf.validate.field).string.max_len = 36];
+
+    option (buf.validate.message).cel = {
+        id: "response_profile_revision_pair"
+        message: "response_profile_id and expected_response_profile_revision must be set together"
+        expression: "(this.response_profile_id > 0) == (this.expected_response_profile_revision != '')"
+    };
```
rl-block, line 748: An already-open pre-change tab omits both fields, so this pair check still passes and the server treats copied profile values as custom. Should we add an explicit request capability or schema version, or a separate profile-execution path, so stale clients fail closed?

server/internal/domain/curtailment/automation.go
```diff
@@ -341,6 +369,19 @@ func (s *AutomationService) handleRuleOff(ctx context.Context, rule *models.Auto
...
 	}
+	replayEvent, err := s.curtailment.LookupStartReplay(ctx, startReq)
+	if err != nil {
+		return err
+	}
+	if replayEvent != nil {
+		if err := validateAutomationReplayEvent(replayEvent, rule); err != nil {
+			return err
+		}
+		return s.store.SetAutomationActiveEvent(ctx, rule.ID, signal.Source.ID, replayEvent.EventUUID, at)
```
rl-block, line 368: If this replay event is `restoring`, this branch only records the pointer; unlike the pointer-present path, it never calls `Recurtail`. Do we want to route restoring replays through the same max-duration, binding, and recurtail flow before setting the pointer?

server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -3155,6 +3361,15 @@ func (s *SQLCurtailmentStore) BeginRecurtailTransition(
...
+			ctx,
+			q,
+			orgID,
+			responseProfileID,
+			responseProfileRevision,
+		); err != nil {
+			return nil, err
+		}
 
 		updated, err := q.ResumeCurtailmentFromRestoring(ctx, current.ID)
```
rl-block, line 3448: This lock proves the supplied profile revision is current, but it does not verify that `current.DecisionSnapshotJsonb` is bound to that ID and revision. Since migration 144 intentionally leaves mismatched events unstamped, could we validate the automation event's snapshot binding before resuming and fail closed when it is absent or different?

server/migrations/000144_curtailment_response_profile_revision.up.sql
```diff
@@ -0,0 +1,94 @@
...
+          profile.mode = 'FIXED_KW'
+          AND (event.mode_params_jsonb->>'target_kw')::NUMERIC = profile.target_kw
+          AND (event.mode_params_jsonb->>'tolerance_kw')::NUMERIC = COALESCE(profile.tolerance_kw, 0)
+      )
+  )
+  AND NOT (event.decision_snapshot_jsonb ? 'response_profile_id')
+  AND NOT (event.decision_snapshot_jsonb ? 'response_profile_revision');
+
+ALTER TABLE curtailment_automation_rule
+    ALTER COLUMN response_profile_revision SET NOT NULL;
```
rl-block: The standard HA updater starts the new passive, which runs migrations, while the old active keeps serving. Once this column is `NOT NULL`, old rule inserts fail and old profile or event writes bypass revision binding, so should we split compatibility and enforcement across releases or require a documented coordinated cutover?

client/src/protoFleet/features/settings/components/Curtailment/CurtailmentAutomations.tsx
```diff
@@ -168,12 +174,12 @@ function getAutomationFormValuesFromRule(
...
 
+  const responseProfile = responseProfiles.find((profile) => profile.id === rule.responseProfileId);
   return {
     name: rule.name,
     sourceId: rule.sourceId ?? sources[0]?.id ?? "",
-    responseProfileId: responseProfiles.some((profile) => profile.id === rule.responseProfileId)
-      ? rule.responseProfileId
-      : "",
+    responseProfileId: responseProfile?.id ?? "",
+    responseProfileRevision: responseProfile?.revision ?? "",
```
rl-block, line 183: If this edit opens before the referenced profile loads, we replace the rule's profile ID with empty and the later dependency effect defaults to the first profile. Worth keeping `rule.responseProfileId` until that exact profile arrives and keep Save blocked if it never becomes available?

server/migrations/000144_curtailment_response_profile_revision.down.sql
```diff
@@ -0,0 +1,12 @@
+UPDATE curtailment_event
+SET decision_snapshot_jsonb = decision_snapshot_jsonb
+    - 'response_profile_id'
+    - 'response_profile_revision'
+WHERE source_actor_type = 'automation'
+  AND external_source = 'curtailment_automation';
```
rl-block: This also strips binding provenance from completed, failed, and cancelled events, while a later up migration only reconstructs live rows. We could limit this cleanup to `pending`, `active`, and `restoring` events so terminal history remains auditable.

server/sqlc/queries/curtailment_response_profile.sql
```diff
@@ -104,6 +112,30 @@ RETURNING *;
...
+    revision = CASE
+        WHEN site_id IS DISTINCT FROM sqlc.narg('site_id')
+          OR scope_json IS DISTINCT FROM sqlc.arg('scope_json')::jsonb
+          OR authorization_envelope_jsonb IS DISTINCT FROM sqlc.arg('authorization_envelope_jsonb')::jsonb
+          OR mode IS DISTINCT FROM sqlc.arg('mode')
+          OR strategy IS DISTINCT FROM sqlc.arg('strategy')
+          OR level IS DISTINCT FROM sqlc.arg('level')
+          OR priority IS DISTINCT FROM sqlc.arg('priority')
+          OR target_kw IS DISTINCT FROM sqlc.narg('target_kw')
+          OR tolerance_kw IS DISTINCT FROM sqlc.narg('tolerance_kw')
```
rl-block: The domain treats `NULL` and zero tolerance as execution-equivalent, but `IS DISTINCT FROM` rotates the revision between those representations and stales bound automations. We could probably compare effective values with `COALESCE(..., 0)`, or canonicalize before writing, and cover both transitions?

rl-block (approved):

PR #978, opened 2026-08-28
Enable topology response profiles for curtailment automation

negarn:
Reviewable diff: +377/-146 across 11 files (excludes generated, test, and story files).

## Summary

Topology-scoped response profiles can now be bound to and executed by MQTT curtailment automation. Building, rack, and group selectors remain logical scopes, so each trigger resolves current membership while every automation scope fails closed if the selector, source principal, permissions, or required admin role changed. This is the automation-enablement slice of #909; the remaining picker/UI and E2E work stays tracked there.

## How it works

When a rule is created, updated, enabled, or triggered, the response profile service strictly resolves its persisted topology selector. A new automation event enters the existing SQL execution transaction, which locks the rule, profile revision, MQTT source, current topology, effective permission assignments, and primary role before persisting any event or target reservation. If an OFF signal reasserts demand while an event is restoring, the re-curtail transaction applies the same rule, source-principal, permission, admin-role, profile, and event-ownership checks before it reclaims targets. The authorization-envelope evaluator is shared with the reconciler's pre-dispatch fence, keeping admission, re-curtail, and physical dispatch aligned.

```mermaid
flowchart LR
    A["MQTT OFF signal"] --> B["Load bound response profile"]
    B --> C{"Existing event is restoring?"}
    C -->|"No"| D["Resolve current building, rack, or group scope"]
    D --> E["Build curtailment plan"]
    E --> F["Transactional start fence"]
    C -->|"Yes"| G["Transactional re-curtail fence"]
    F --> H{"Binding and authorization still valid?"}
    G --> H
    H -->|"Yes"| I["Persist event or reclaim targets"]
    H -->|"No"| J["Reject without changing ownership"]
    I --> K["Reconciler revalidates before Curtail dispatch"]
```

## Areas of the code involved

| --- | --- | --- |
| `client/` curtailment profile types | Marks every recognized terminal scope as automation-ready | Building, rack, and group profiles are selectable without treating unknown scopes as valid |
| `server/internal/domain/curtailment` | Removes topology staging gates, adds strict live selector validation, and carries the automation execution fence through re-curtail | Binding and trigger paths preserve logical selectors and re-curtail cannot bypass current rule ownership |
| Authorization envelope | Extracts the shared current-permission coverage check | Start, re-curtail, and dispatch use the same site/org-wide authorization semantics for every scope |
| `server/internal/domain/stores/sqlstores` | Adds transactional source-principal, permission, role, topology, profile, rule, and event-ownership fences | Unauthorized automation cannot reserve or reclaim miners before dispatch |
| `server/sqlc/queries/` | Locks and verifies the current MQTT source user and primary role | Concurrent source edits or role demotions serialize with event creation |
| `server/generated/sqlc/` | Regenerated SQL bindings — skip | Generated from the reviewed query sources |

- Keep topology selectors logical and resolve them live instead of snapshotting miner IDs into the automation rule.
- Recheck authorization inside start and re-curtail transactions instead of relying only on the later dispatch fence, preventing unauthorized reservations from being persisted or reclaimed.
- Apply permission and admin-role checks to every automation scope, including site, whole-org, and explicit-miner profiles.
- Preserve idempotent replay recovery even when a profile later changes; strict current-profile validation applies only before creating or re-curtailing an event.
- Keep profile reads tolerant of stale topology for settings/history, while execution-capable paths opt into strict resolution.

## Testing & validation

- `go test ./internal/domain/curtailment/... ./internal/handlers/curtailment`
- `go test ./internal/domain/stores/sqlstores -run '^$'`
- `go build ./...` from `server/`
- `just _lint-server`
- `go test ./cmd/fleetnode -run '^TestRunCmd_RefreshFailureCannotExtendControlStreamPastExpiry$' -count=10`
- `npx vitest run` for the response-profile hook, automation settings, and curtailment settings suites (89 tests)
- `npx tsc --noEmit`
- `npm run build:protoFleet`
- `just _lint-client`
- Added SQL integration coverage for deleted selectors; topology and site permission revocation; admin demotion; changed MQTT source principals; and successful or rejected transactional re-curtail. The local integration database could not authenticate the configured `fleet` user, so those cases are compiled locally and will execute in CI.

Refs #909

### DIFF
--- a/server/internal/domain/curtailment/automation.go
+++ b/server/internal/domain/curtailment/automation.go
@@ -367,7 +367,7 @@ func (s *AutomationService) handleRuleOff(ctx context.Context, rule *models.Auto
 		}
 		return s.store.SetAutomationActiveEvent(ctx, rule.ID, signal.Source.ID, replayEvent.EventUUID, at)
 	}
-	if err := validateBoundAutomationProfile(rule, profile); err != nil {
+	if _, err := s.currentBoundAutomationProfile(ctx, rule); err != nil {
 		return err
 	}
 	plan, err := s.curtailment.Start(ctx, startReq)
@@ -536,7 +536,7 @@ func (s *AutomationService) validateAndNormalize(
 			"curtailment response profile changed before automation rule save; retry",
 		)
 	}
-	if err := validateAutomationProfileBinding(profile, canUseAdminControls); err != nil {
+	if err := s.validateAutomationProfileBinding(ctx, profile, canUseAdminControls); err != nil {
 		return models.AutomationRule{}, models.ResponseProfileFanSettings{}, err
 	}
 	if !sameResponseProfileFanSettings(responseProfileFanSettings(profile), expectedFanSettings) {
@@ -580,7 +580,7 @@ func (s *AutomationService) ensureProfileCanBeAutomated(
 			"curtailment response profile changed before automation rule save; retry",
 		)
 	}
-	if err := validateAutomationProfileBinding(profile, canUseAdminControls); err != nil {
+	if err := s.validateAutomationProfileBinding(ctx, profile, canUseAdminControls); err != nil {
 		return models.ResponseProfileFanSettings{}, err
 	}
 	if !sameResponseProfileFanSettings(responseProfileFanSettings(profile), expectedFanSettings) {
@@ -605,6 +605,9 @@ func (s *AutomationService) currentBoundAutomationProfile(
 	if err := validateBoundAutomationProfile(rule, profile); err != nil {
 		return nil, err
 	}
+	if err := s.profiles.ValidateAutomationScope(ctx, profile); err != nil {
+		return nil, err
+	}
 	return profile, nil
 }
 
@@ -622,18 +625,16 @@ func validateBoundAutomationProfile(rule *models.AutomationRule, profile *models
 	return nil
 }
 
-func validateAutomationProfileBinding(profile *models.ResponseProfile, canUseAdminControls bool) error {
+func (s *AutomationService) validateAutomationProfileBinding(
+	ctx context.Context,
+	profile *models.ResponseProfile,
+	canUseAdminControls bool,
+) error {
 	if profile == nil {
 		return nil
 	}
-	scope, err := ResponseProfileScope(*profile)
-	if err != nil {
+	if err := s.profiles.ValidateAutomationScope(ctx, profile); err != nil {
 		return err
-	}
-	if hasTopologySelectors(scope) {
-		return fleeterror.NewFailedPreconditionError(
-			"topology-scoped response profiles cannot be used by automation until topology curtailment execution is supported",
-		)
 	}
 	if canUseAdminControls || !responseProfileRequiresAdminControls(*profile) {
 		return nil
--- a/server/internal/domain/curtailment/automation.go
+++ b/server/internal/domain/curtailment/automation.go
@@ -608,9 +608,6 @@ func (s *AutomationService) currentBoundAutomationProfile(
 	if err := validateBoundAutomationProfile(rule, profile); err != nil {
 		return nil, err
 	}
-	if err := s.profiles.ValidateAutomationScope(ctx, profile); err != nil {
-		return nil, err
-	}
 	return profile, nil
 }
 
--- a/server/internal/domain/curtailment/automation.go
+++ b/server/internal/domain/curtailment/automation.go
@@ -323,7 +323,7 @@ func (s *AutomationService) handleRuleOff(ctx context.Context, rule *models.Auto
 		case event.State.IsTerminal():
 			// Stale terminal state; start a fresh event below.
 		case event.State == models.EventStateRestoring:
-			return s.recurtailAutomationEvent(ctx, rule, signal.Source.ID, event, at)
+			return s.recurtailAutomationEvent(ctx, rule, signal.Source, event, at)
 		default:
 			return nil
 		}
@@ -363,7 +363,7 @@ func (s *AutomationService) handleRuleOff(ctx context.Context, rule *models.Auto
 			return err
 		}
 		if replayEvent.State == models.EventStateRestoring {
-			return s.recurtailAutomationEvent(ctx, rule, signal.Source.ID, replayEvent, at)
+			return s.recurtailAutomationEvent(ctx, rule, signal.Source, replayEvent, at)
 		}
 		return s.store.SetAutomationActiveEvent(ctx, rule.ID, signal.Source.ID, replayEvent.EventUUID, at)
 	}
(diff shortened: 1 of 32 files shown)


server/internal/domain/curtailment/automation.go
```diff
@@ -418,11 +418,14 @@ func (s *AutomationService) recurtailAutomationEvent(
 		EventUUID:               event.EventUUID,
 		ResponseProfileID:       rule.ResponseProfileID,
 		ResponseProfileRevision: rule.ResponseProfileRevision,
+		AutomationRuleID:        rule.ID,
+		AutomationMQTTSourceID:  mqttSource.ID,
+		AutomationServiceUserID: mqttSource.ServiceUserID,
```
rl-block, line 423: Worth binding re-curtail to the event's original service user? This forwards the source's current principal, but the transactional ownership check never compares it with `event.CreatedByUserID`. After a source reassignment, the replacement principal could therefore reactivate the prior user's event. Perhaps we could compare those IDs and update the source-change test to pass the replacement user.

server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -1301,6 +1283,79 @@ func lockAutomationRuleForExecution(
...
+	}
+	effective, err := authz.LoadEffectiveForUpdate(ctx, q, serviceUserID, event.OrgID)
+	if err != nil {
+		return fleeterror.NewInternalErrorf("failed to reload automation source user permissions: %v", err)
+	}
+	envelope, err := domainCurtailment.AuthorizationEnvelopeFromJSON(event.AuthorizationEnvelopeJSON)
+	if err != nil {
+		return err
+	}
+	if !domainCurtailment.AuthorizationEnvelopeAllows(effective, envelope, nil, nil, false) {
```
rl-block: We might want to pass live topology coverage into this check for re-curtail. A building, rack, or group can move after the service preflight, but this transaction authorizes only the persisted envelope before resetting restore targets. We could lock and resolve the selector here and leave the event restoring when coverage drifts.

server/internal/domain/curtailment/automation.go
```diff
@@ -785,15 +789,21 @@ func validateAutomationReplayOwnership(event *models.Event, rule *models.Automat
...
+// ValidateAutomationEventOwnership verifies that a persisted event carries
+// every immutable ownership marker for the expected automation rule.
+func ValidateAutomationEventOwnership(event *models.Event, orgID, ruleID int64) error {
+	externalReference, idempotencyKey := automationRuleEventReference(ruleID)
+	if event == nil || event.OrgID != orgID ||
 		event.SourceActorType != models.SourceActorAutomation ||
 		event.ExternalSource == nil || *event.ExternalSource != automationExternalSource ||
 		event.ExternalReference == nil || *event.ExternalReference != externalReference ||
 		event.IdempotencyKey == nil || *event.IdempotencyKey != idempotencyKey ||
 		event.SourceActorID == nil || *event.SourceActorID != externalReference {
```
rl-block, line 801: This ownership predicate perhaps could get a table-driven test that changes each marker independently, plus one SQL re-curtail case. Current negative coverage changes only `SourceActorType`, so regressions in the external reference, idempotency key, actor ID, or organization checks could pass unnoticed.

server/internal/domain/stores/sqlstores/response_profile_revision_integration_test.go
```diff
@@ -484,6 +496,358 @@ func TestSQLCurtailmentStore_AutomationExecutionRejectsReboundRule(t *testing.T)
...
+			execution.AutomationMQTTSourceID = sourceID
+			execution.ForceIncludeAllPairedMiners = tt.requiresAdminRole
+			execution.DecisionSnapshotJSON = []byte(fmt.Sprintf(
+				`{"requires_admin_controls":%t}`,
+				tt.requiresAdminRole,
+			))
+
+			_, err = store.InsertEventWithTargets(ctx, execution, nil)
+
+			require.Error(t, err)
```
rl-block: Should this matrix include one unchanged, authorized topology case that expects success? Every topology SQL execution here currently requires an error, so the suite would still pass if the real store rejected all valid building-scoped automation starts. It might be worth asserting the persisted scope and authorization envelope on the happy path.

server/internal/domain/stores/sqlstores/curtailment.go
```diff
@@ -1301,6 +1283,79 @@ func lockAutomationRuleForExecution(
...
+) error {
+	if ruleID == 0 {
+		return nil
+	}
+	if event == nil {
+		return fleeterror.NewInternalError("automation execution authorization requires an event")
+	}
+	effective, err := authz.LoadEffectiveForUpdate(ctx, q, serviceUserID, event.OrgID)
+	if err != nil {
+		return fleeterror.NewInternalErrorf("failed to reload automation source user permissions: %v", err)
```
rl-block: This could probably wrap `err` with `%w` instead of `%v`, and do the same for the role lookup below? `WithTransaction` needs the underlying PostgreSQL error to classify retries; `%v` drops the `FleetError` cause, so deadlock or serialization failures skip the bounded transaction retry.

server/internal/domain/stores/sqlstores/response_profile_revision_integration_test.go
```diff
@@ -484,6 +496,358 @@ func TestSQLCurtailmentStore_AutomationExecutionRejectsReboundRule(t *testing.T)
...
+			execution.AutomationRuleID = rule.ID
+			execution.AutomationMQTTSourceID = sourceID
+			execution.DecisionSnapshotJSON = []byte(fmt.Sprintf(
+				`{"response_profile_id":%d,"response_profile_revision":%q,"requires_admin_controls":true}`,
+				profile.ID,
+				profile.Revision.String(),
+			))
+			inserted, err := store.InsertEventWithTargets(ctx, execution, nil)
+			require.NoError(t, err)
+			_, err = database.ExecContext(ctx, `UPDATE curtailment_event SET state = 'restoring' WHERE id = $1`, inserted.ID)
```
rl-block: Maybe we could drive this event-state change and the source-principal mutation below through the existing sqlc paths. `AGENTS.md` says all server database access goes through sqlc, and this would keep the test aligned with supported update semantics.

server/internal/domain/curtailment/automation.go
```diff
@@ -363,11 +363,11 @@ func (s *AutomationService) handleRuleOff(ctx context.Context, rule *models.Auto
...
 		if replayEvent.State == models.EventStateRestoring {
-			return s.recurtailAutomationEvent(ctx, rule, signal.Source.ID, replayEvent, at)
+			return s.recurtailAutomationEvent(ctx, rule, signal.Source, replayEvent, at)
 		}
 		return s.store.SetAutomationActiveEvent(ctx, rule.ID, signal.Source.ID, replayEvent.EventUUID, at)
 	}
-	if err := validateBoundAutomationProfile(rule, profile); err != nil {
+	if _, err := s.currentBoundAutomationProfile(ctx, rule); err != nil {
 		return err
 	}
```
rl-block, line 372: Do we need to resolve topology here before `Service.Start`? The path already loaded the profile, and `Start` resolves the same selector again before the transactional insert fence. Perhaps validating the existing profile binding here would avoid a second coverage query while preserving the live checks.

rl-block (approved):

PR #981, opened 2026-08-29
Add topology curtailment profile E2E coverage

negarn:
Reviewable diff: +192/-23 across 4 files (excludes generated, test, and story files).

## Summary

This PR adds end-to-end coverage for building-, rack-, and group-scoped curtailment response profiles across both Settings and Energy. The scenario verifies that topology scopes and the all-paired policy survive create, reload, edit, Test, profile selection, and live Run flows without widening into a multi-type union.

## How it works

The Playwright scenario provisions a site, building, rack, two assigned miners, and a group. It creates an all-paired building profile, reloads and changes the terminal scope to the rack, reloads again and changes it to the group, then tests the profile from Settings and runs the saved profile from New curtailment. Each API boundary is inspected for one typed terminal scope, schema version, profile revision, and all-paired policy; each started event is restored before fixtures are deleted in dependency order.

## Diagrams

```mermaid
flowchart TD
  A["Create site, building, rack, miners, and group"] --> B["Create building-scoped response profile"]
  B --> C["Reload and replace scope with rack"]
  C --> D["Reload and replace scope with group"]
  D --> E["Test profile from Settings"]
  E --> F["Select saved profile in New curtailment"]
  F --> G["Run and verify active event"]
  G --> H["Restore event and delete fixtures"]
```

## Areas of the code involved

| --- | --- | --- |
| `client/e2eTests/protoFleet/pages/components/curtailmentModal.ts` | Adds shared topology selection and run-confirmation behavior. | Keeps both modal entry points on the same Schedules-style drill-down interactions and makes repeated parent selection idempotent. |
| `client/e2eTests/protoFleet/pages/energy.ts` | Adds saved-profile selection, topology assertions, and shared modal execution. | Covers the New curtailment entry point and exposes typed Start request fields. |
| `client/e2eTests/protoFleet/pages/settingsCurtailment.ts` | Adds profile edit/reload/Test and topology-target actions. | Covers create/edit response profile behavior through persisted scope changes. |
| `client/e2eTests/protoFleet/pages/groups.ts` | Adds a fixture helper that creates a group from selected miner IPs and returns its persisted ID. | Lets request assertions compare the submitted group selector with the actual created resource. |
| `client/e2eTests/protoFleet/spec/curtailmentTopology.spec.ts` | Adds the full create, reload, edit, Test, Run, restore, and cleanup scenario. | Proves the issue #909 workflow across real frontend/backend boundaries in the fake-miner environment. |

- Exercise all three new terminal types in one lifecycle scenario so the test proves replacement semantics and persistence without repeating expensive fleet setup.
- Assert request IDs and policy fields at each transition instead of relying only on UI summaries, which catches accidental scope widening or profile fallback.
- Run only outside the real-miner target because the scenario intentionally starts and restores curtailment commands.
- Clean up from a fresh authenticated browser context and restore events before deleting dependent topology resources so interrupted test pages do not leave active ownership or fixtures behind.

## Testing & validation

- `cd client && ../bin/npx eslint e2eTests/protoFleet/pages/energy.ts e2eTests/protoFleet/pages/groups.ts e2eTests/protoFleet/pages/settingsCurtailment.ts e2eTests/protoFleet/pages/components/curtailmentModal.ts e2eTests/protoFleet/spec/curtailmentTopology.spec.ts`
- `cd client && ../bin/npx prettier --check ...` for the five changed files.
- `cd client && ../bin/npx tsc --noEmit`
- `cd client/e2eTests/protoFleet && ../../../bin/npx playwright test spec/curtailmentTopology.spec.ts --project=desktop --list`
- The live Playwright scenario was not run locally because the configured frontend and backend were not running; CI provides the simulator-backed execution.

Refs #909

### DIFF
--- a/client/e2eTests/protoFleet/pages/components/curtailmentModal.ts
+++ b/client/e2eTests/protoFleet/pages/components/curtailmentModal.ts
@@ -33,7 +33,16 @@ export class CurtailmentModal {
   }
 
   async confirmRun() {
-    await this.root.getByRole("button", { name: "Run curtailment", exact: true }).click();
+    const runButton = this.root.getByRole("button", { name: "Run curtailment", exact: true });
+    if (await runButton.isVisible().catch(() => false)) {
+      await runButton.click();
+    } else {
+      await this.root.getByTestId("overflow-menu-trigger").click();
+      await this.page
+        .getByTestId("modal-overflow-sheet-content")
+        .getByRole("button", { name: "Run curtailment", exact: true })
+        .click();
+    }
 
     const forceConfirmation = this.page.getByTestId("curtailment-force-inclusion-confirmation");
     const runConfirmation = this.page.getByTestId("curtailment-run-confirmation");
--- a/client/e2eTests/protoFleet/pages/components/curtailmentModal.ts
+++ b/client/e2eTests/protoFleet/pages/components/curtailmentModal.ts
@@ -1,0 +1,67 @@
+import { expect, type Locator, type Page } from "@playwright/test";
+
+type TopologyTarget = "Buildings" | "Groups" | "Racks" | "Sites";
+
+const selectionModalTitles: Record<TopologyTarget, string> = {
+  Buildings: "Select buildings",
+  Groups: "Select groups",
+  Racks: "Select racks",
+  Sites: "Select sites",
+};
+
+export class CurtailmentModal {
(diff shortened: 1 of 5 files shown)


illegalprime (commented):
neat

illegalprime (approved):

PR #985, opened 2026-08-31
Pre-bake ProtoFleet E2E runner dependencies

illegalprime:
Reviewable diff: +1552/-12 across 8 files (excludes generated, test, and story files).

## Summary

ProtoFleet E2E can now run on an opt-in GitHub-hosted custom VM image that preloads the exact client dependency tree, matching Playwright browser, and Docker Compose images. Hash and presence checks keep stale image components from being used, while the existing `ubuntu-latest`, Actions cache, and Docker artifact paths remain the default and fallback. This also checks in the timing analyzer and baseline used to measure whether the runner reduces overhead or causes regressions.

## How it works

A trusted weekly or manual workflow on the default branch prepares a Linux x64 image-generation runner, writes a manifest containing the source SHA plus `hashFiles` digests, and snapshots it as `protofleet-e2e`. Once an administrator provisions a runtime larger runner and sets `PROTOFLEET_E2E_RUNNER`, the E2E build, 32 functional shards, and two visual jobs land on that image.

After checkout, a local action compares the current server and lockfile hashes with the manifest. Matching client dependencies are copied into the workspace, Playwright is pointed at the baked Chromium tree, and Docker images are accepted only if every recorded image is still present in the host daemon. Each component falls back independently; scheduled E2E runs always rebuild Docker images against fresh upstream bases, and the build job still publishes an image artifact so older runner instances can fall back safely during image rollouts.

The timing analyzer reads completed PR Gate runs and latest-attempt job data from GitHub, groups the newest successful runs per PR, classifies step time as test, validation/build, or overhead, and emits Markdown/JSON plus a regression comparison against the checked-in baseline.

## Diagrams

```mermaid
flowchart LR
  A["Trusted default-branch image workflow"] --> B["Image-generation larger runner"]
  B --> C["Install Node dependencies and Chromium"]
  B --> D["Pull and build Compose images"]
  C --> E["Write hash manifest"]
  D --> E
  E --> F["Snapshot protofleet-e2e"]
  F --> G["Runtime larger runner"]
  H["PR E2E job"] --> I["Hash-aware restore action"]
  G --> I
  I --> J{"Component matches and exists?"}
  J -->|"Yes"| K["Use pre-baked component"]
  J -->|"No"| L["Use existing cache or artifact path"]
  K --> M["Build, functional shards, and visual tests"]
  L --> M
```

```mermaid
sequenceDiagram
  participant Admin as "CI administrator"
  participant Builder as "Image workflow"
  participant GitHub as "Custom image store"
  participant E2E as "PR E2E job"
  Admin->>Builder: "Configure builder variable and run on main"
  Builder->>GitHub: "Snapshot tools, dependencies, Docker images, and manifest"
  Admin->>E2E: "Configure runtime runner variable"
  E2E->>E2E: "Compare current hashes and inspect Docker images"
  alt "Valid pre-baked component"
    E2E->>E2E: "Restore or reuse local component"
  else "Missing or stale component"
    E2E->>E2E: "Restore Actions cache/artifact or rebuild"
  end
```

## Areas of the code involved

| --- | --- | --- |
| `.github/workflows/protofleet-e2e-runner-image.yml` | Adds trusted weekly/manual custom-image generation with GitHub `snapshot` | Verify the image lifecycle, default-branch restriction, and organization runner dependency |
| `.github/actions/protofleet-e2e-runner/` | Exposes independently validated client, Playwright, and Docker hits | This is the boundary that decides whether setup steps may be skipped |
| `.github/scripts/prepare_protofleet_e2e_runner.sh` | Installs exact dependencies and Compose images, then writes the root-owned manifest | Review image contents, hash inputs, and snapshot safety |
| `.github/scripts/restore_protofleet_e2e_runner.sh` | Validates the manifest, restores dependencies, and inspects host Docker images | Review fail-safe behavior for stale, missing, and partially available content |
| `.github/workflows/protofleet-e2e-tests.yml` | Selects the configured runner and conditionally skips redundant setup | Verify the normal runner path is unchanged on misses and scheduled rebuilds stay fresh |
| `.github/scripts/analyze_ci_timings.py` and baseline | Adds repeatable historical step timing, Markdown/JSON reports, caching, and regression thresholds | Review sampling semantics and name-based timing classification |
| `docs/plans/2026-08-31-protofleet-e2e-prebaked-runner-plan.md` | Documents administrator activation, canary measurement, and rollback | Organization-level runners cannot be provisioned by repository code |
| Python tests | Covers run selection/classification and runner hit/miss/fallback behavior | Test-only — focus on edge-case coverage |

- Use a custom runner VM rather than a job container, because the measured hotspot is hydrating the host Docker daemon used by Compose.
- Validate server and client inputs independently rather than invalidating the entire image, so client-only and server-only changes retain safe partial hits.
- Keep publishing the Docker artifact on a pre-baked hit rather than maximizing the single build-job saving, so mixed image versions can fall back during rollout.
- Force scheduled runs down the Docker rebuild path rather than using a weekly snapshot, preserving the existing fresh-upstream-base behavior.
- Keep `ubuntu-latest` as the variable fallback rather than requiring organization infrastructure at merge time; activation and rollback are configuration-only.
- Classify timing by normalized step name rather than workflow-specific IDs, which supports reusable workflows but surfaces unmatched steps explicitly for review.

## Testing & validation

- `server/sdk/v1/python/.venv/bin/python -B .github/scripts/protofleet_e2e_runner_test.py` — 5 tests passed.
- `server/sdk/v1/python/.venv/bin/python -B .github/scripts/analyze_ci_timings_test.py` — 7 tests passed.
- `server/sdk/v1/python/.venv/bin/python -B .github/scripts/evaluate_review_policy_test.py` — 72 tests passed.
- Ruff check/format, shell syntax, YAML parsing, Prettier, and `git diff --check` passed.
- Pre-commit hooks passed Ruff and protected-branch checks; pre-push hooks passed client typecheck and server lint.
- Not covered yet: generating the real GitHub custom image and running the E2E canary. Those require an organization administrator to provision the image-generation/runtime larger runners and set the two documented variables.

### DIFF
--- a/.github/actions/protofleet-e2e-runner/action.yml
+++ b/.github/actions/protofleet-e2e-runner/action.yml
@@ -1,0 +1,40 @@
+name: Restore ProtoFleet E2E runner content
+description: Use hash-matched dependencies and Docker images from the custom runner image
+
+inputs:
+  server-hash:
+    description: GitHub hashFiles digest for server/**
+    required: true
+  client-lock-hash:
+    description: GitHub hashFiles digest for client/package-lock.json
+    required: true
+  allow-docker-images:
+    description: Whether pre-baked Docker images may be used for this run
+    required: false
+    default: "true"
+
+outputs:
+  image-version:
+    description: Source commit recorded in the runner image
+    value: ${{ steps.restore.outputs.image-version }}
+  client-dependencies:
+    description: Whether client/node_modules was restored from the runner image
+    value: ${{ steps.restore.outputs.client-dependencies }}
+  playwright-browsers:
+    description: Whether Playwright will use browsers from the runner image
+    value: ${{ steps.restore.outputs.playwright-browsers }}
+  docker-images:
+    description: Whether every required Docker image is present and hash-matched
+    value: ${{ steps.restore.outputs.docker-images }}
+
+runs:
+  using: composite
+  steps:
+    - name: Inspect and restore pre-baked content
+      id: restore
+      shell: bash
+      env:
+        PROTOFLEET_CURRENT_SERVER_HASH: ${{ inputs.server-hash }}
+        PROTOFLEET_CURRENT_CLIENT_LOCK_HASH: ${{ inputs.client-lock-hash }}
+        PROTOFLEET_ALLOW_PREBAKED_DOCKER: ${{ inputs.allow-docker-images }}
+      run: "$GITHUB_ACTION_PATH/../../scripts/restore_protofleet_e2e_runner.sh"
--- a/.github/scripts/analyze_ci_timings.py
+++ b/.github/scripts/analyze_ci_timings.py
@@ -1,0 +1,1065 @@
+#!/usr/bin/env python3
+r"""Analyze GitHub Actions runner time by step category.
+
+The default selection reproduces the CI investigation performed on 2026-08-31:
+the latest 100 completed PR Gate runs, with detailed job data for up to the
+newest three successful runs per pull request. GitHub's API returns rerun jobs
+separately, so the analyzer requests only the latest attempt for each run.
+
+Examples:
+
+    GITHUB_TOKEN=... python3 .github/scripts/analyze_ci_timings.py \
+      --json-out /tmp/proto-fleet-ci-baseline.json
+
+    GITHUB_TOKEN=... python3 .github/scripts/analyze_ci_timings.py \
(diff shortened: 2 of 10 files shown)


ankitgoswami (commented):
seems like a good idea if we can get a custom runner

PR #989, opened 2026-09-01
make Fleet Node enrollment one command

ankitgoswami:
Reviewable diff: +84/-8 across 5 files (excludes generated, test, and story files).

## Summary

Fleet Node enrollment is now a single copy/paste command: `sudo fleetnode-enroll --server-url=...`. The packaged helper owns the service-account, state-directory, and systemd details, so operators no longer need the verbose internal command or a separate service-start step.

**Stack:** [#990 Make Fleet Node startup readiness explicit](https://github.com/block/proto-fleet/pull/990) → [#991 Complete the Fleet Node installer lifecycle](https://github.com/block/proto-fleet/pull/991) → [#992 Qualify Fleet Node release artifacts](https://github.com/block/proto-fleet/pull/992) → **#989 (this PR)**. This diff is relative to #992. The upstream PRs establish the service-readiness, transactional installer, and clean-host qualification contracts that this helper extends; broader release/distribution work remains outside this PR.

## How it works

The release artifact includes `fleetnode-enroll`, and the installer places a managed copy at `/usr/local/bin/fleetnode-enroll`. The operator UI builds the server URL exactly as before, but now displays the short helper command. The helper runs the installed Fleet Node binary as the `fleetnode` service account with `/var/lib/fleetnode` as its state directory, then enables and starts `fleet-node.service` only after enrollment succeeds.

Installer upgrades treat the helper like the binary and unit: an existing managed copy is backed up and restored if installation fails, while an unrelated file at the global path is rejected rather than overwritten. Uninstall and purge remove the managed helper.

## Diagrams

```mermaid
flowchart LR
    A["Operator copies sudo fleetnode-enroll command"] --> B["Managed helper in /usr/local/bin"]
    B --> C["Fleet Node runs as fleetnode user"]
    C --> D["Enrollment state is written to /var/lib/fleetnode"]
    D --> E["systemd enables and starts fleet-node.service"]
```

## Areas of the code involved

| --- | --- | --- |
| `deployment-files/fleetnode/fleetnode-enroll` | Adds the operator-facing enrollment entry point | Confirm privilege boundaries, argument forwarding, and start-after-success ordering |
| `deployment-files/fleetnode/install-fleet-node.sh` | Installs, validates, rolls back, and removes the helper | Confirm upgrades remain transactional and do not overwrite unmanaged files |
| `.github/workflows/` | Packages the helper and syntax-checks it in deployment CI | Confirms every published Fleet Node artifact contains the command shown by the UI |
| `client/.../Nodes/enrollNodeCommand.ts` | Produces the concise command while retaining existing URL and HTTP transport rules | Confirm UI output matches the installed package contract |
| Installer, qualification, and client tests | Cover install/upgrade/removal behavior and exact UI output | Prevents packaging and operator guidance from drifting apart |

- Install a dedicated `fleetnode-enroll` helper instead of placing the raw `fleetnode` binary on `PATH`; this keeps the full CLI privilege model explicit while making the supported enrollment workflow concise.
- Keep the helper root-owned and switch only the enrollment process to the `fleetnode` account; systemd enablement still has the privilege it requires.
- Start the service only after successful enrollment; a canceled or failed enrollment leaves lifecycle state unchanged.
- Reject an unrelated `/usr/local/bin/fleetnode-enroll` instead of overwriting it, trading automatic replacement for safe host ownership boundaries.

## Testing & validation

- `bash deployment-files/fleetnode/tests/test-install-fleetnode.sh`
- `bash -n` for the helper, installer, installer tests, and package qualification script
- `vitest --run` for `enrollNodeCommand.test.ts` and `EnrollNodeModal.test.tsx`: 13 tests passed
- Client ESLint passed with zero warnings; targeted Prettier check passed
- Workflow YAML parsed successfully; `git diff --check` passed
- Not manually exercised against a live Fleet server; clean-host package qualification verifies installation and non-mutating helper help, while CI performs the architecture-specific artifact checks

### DIFF
--- a/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
+++ b/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
@@ -44,7 +44,11 @@ describe("EnrollNodeModal", () => {
     const { getByText } = render(<EnrollNodeModal open onDismiss={mockOnDismiss} onUpdated={mockOnUpdated} />);
 
     await waitFor(() => {
-      expect(getByText(/fleetnode enroll --server-url=/)).toBeInTheDocument();
+      expect(
+        getByText(
+          "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll --server-url=http://localhost:4000 && sudo systemctl enable --now fleet-node.service",
+        ),
+      ).toBeInTheDocument();
       expect(getByText("pf_code_123")).toBeInTheDocument();
       expect(getByText("Waiting for the node to register…")).toBeInTheDocument();
     });
--- a/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
+++ b/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
@@ -47,10 +47,8 @@ describe("EnrollNodeModal", () => {
 
     await waitFor(() => {
       const command = getByRole("button", { name: "Copy command" }).parentElement;
-      expect(command).toHaveTextContent("sudo -u fleetnode /opt/fleetnode/fleetnode");
-      expect(command).toHaveTextContent("--state-dir /var/lib/fleetnode");
-      expect(command).toHaveTextContent("enroll --server-url=http://localhost:4000");
-      expect(command).toHaveTextContent("sudo systemctl enable --now fleet-node.service");
+      expect(command).toHaveTextContent("sudo fleetnode-enroll");
+      expect(command).toHaveTextContent("--server-url=http://localhost:4000");
       expect(getByText("pf_code_123")).toBeInTheDocument();
       expect(getByText("Waiting for the node to register…")).toBeInTheDocument();
     });
--- a/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
+++ b/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
@@ -50,7 +50,7 @@ describe("EnrollNodeModal", () => {
       expect(command).toHaveTextContent("sudo -u fleetnode /opt/fleetnode/fleetnode");
       expect(command).toHaveTextContent("--state-dir /var/lib/fleetnode");
       expect(command).toHaveTextContent("enroll --server-url=http://localhost:4000");
-      expect(command).toHaveTextContent("sudo systemctl enable fleet-node.service");
+      expect(command).toHaveTextContent("sudo systemctl enable --now fleet-node.service");
       expect(getByText("pf_code_123")).toBeInTheDocument();
       expect(getByText("Waiting for the node to register…")).toBeInTheDocument();
     });
--- a/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
+++ b/client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
@@ -41,14 +41,16 @@ describe("EnrollNodeModal", () => {
     mockListFleetNodes.mockResolvedValue([]);
     mockCreateEnrollmentCode.mockResolvedValue({ code: "pf_code_123", pendingEnrollmentId: "11", expiresAt: null });
 
-    const { getByText } = render(<EnrollNodeModal open onDismiss={mockOnDismiss} onUpdated={mockOnUpdated} />);
+    const { getByRole, getByText } = render(
+      <EnrollNodeModal open onDismiss={mockOnDismiss} onUpdated={mockOnUpdated} />,
+    );
 
     await waitFor(() => {
-      expect(
-        getByText(
-          "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll --server-url=http://localhost:4000 && sudo systemctl enable --now fleet-node.service",
-        ),
-      ).toBeInTheDocument();
+      const command = getByRole("button", { name: "Copy command" }).parentElement;
+      expect(command).toHaveTextContent("sudo -u fleetnode /opt/fleetnode/fleetnode");
+      expect(command).toHaveTextContent("--state-dir /var/lib/fleetnode");
+      expect(command).toHaveTextContent("enroll --server-url=http://localhost:4000");
+      expect(command).toHaveTextContent("sudo systemctl enable fleet-node.service");
       expect(getByText("pf_code_123")).toBeInTheDocument();
       expect(getByText("Waiting for the node to register…")).toBeInTheDocument();
     });
--- a/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
+++ b/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
@@ -2,8 +2,7 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
-const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
-const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable --now fleet-node.service";
+const FLEET_NODE_ENROLL_COMMAND = "sudo fleetnode-enroll";
 
 const isLoopbackHostname = (hostname: string) => {
   const normalized = hostname.toLowerCase().replace(/^\[|\]$/g, "");
@@ -27,5 +26,5 @@ export const buildFleetNodeEnrollCommand = (location: Pick<Location, "origin" | 
     ? `${enrollCommand} ${INSECURE_TRANSPORT_FLAG}`
     : enrollCommand;
 
-  return `${command} && ${FLEET_NODE_ENABLE_COMMAND}`;
+  return command;
 };
--- a/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
+++ b/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
@@ -3,7 +3,7 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
 const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
-const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable --now fleet-node.service";
+const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable fleet-node.service";
 
 const isLoopbackHostname = (hostname: string) => {
   const normalized = hostname.toLowerCase().replace(/^\[|\]$/g, "");
--- a/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
+++ b/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
@@ -3,7 +3,7 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
 const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
-const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable fleet-node.service";
+const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable --now fleet-node.service";
 
 const isLoopbackHostname = (hostname: string) => {
   const normalized = hostname.toLowerCase().replace(/^\[|\]$/g, "");
--- a/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
+++ b/client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
@@ -2,6 +2,8 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
+const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
+const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable --now fleet-node.service";
 
 const isLoopbackHostname = (hostname: string) => {
   const normalized = hostname.toLowerCase().replace(/^\[|\]$/g, "");
@@ -20,6 +22,10 @@ const getHttpFleetApiOrigin = (origin: string) => {
 export const buildFleetNodeEnrollCommand = (location: Pick<Location, "origin" | "protocol" | "hostname">) => {
   const serverUrl =
     location.protocol === "http:" ? getHttpFleetApiOrigin(location.origin) : `${location.origin}${API_PROXY_BASE}`;
-  const command = `fleetnode enroll --server-url=${serverUrl}`;
-  return shouldAppendInsecureTransportFlag(location) ? `${command} ${INSECURE_TRANSPORT_FLAG}` : command;
+  const enrollCommand = `${FLEET_NODE_ENROLL_COMMAND} --server-url=${serverUrl}`;
+  const command = shouldAppendInsecureTransportFlag(location)
+    ? `${enrollCommand} ${INSECURE_TRANSPORT_FLAG}`
+    : enrollCommand;
+
+  return `${command} && ${FLEET_NODE_ENABLE_COMMAND}`;
 };
--- a/deployment-files/fleetnode/fleetnode-enroll
+++ b/deployment-files/fleetnode/fleetnode-enroll
@@ -4,35 +4,83 @@ set -euo pipefail
 PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 export PATH
 
+FLEETNODE_BIN="/opt/fleetnode/fleetnode"
+STATE_DIR="/var/lib/fleetnode"
+STATE_PATH="$STATE_DIR/state.yaml"
+SERVICE_NAME="fleet-node.service"
+
 usage() {
   cat <<'EOF'
 Usage: sudo fleetnode-enroll --server-url=URL [--allow-insecure-transport]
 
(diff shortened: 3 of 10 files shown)


client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
```diff
@@ -2,6 +2,8 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
+const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
+const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable --now fleet-node.service";
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> `sudo systemctl enable --now fleet-node.service` doesn’t match the packaged installer guidance in `deployment-files/fleetnode/install-fleet-node.sh` (it prints `sudo systemctl enable fleet-node.service` without `--now`). If the UI is intended to mirror the installer, consider switching to `enable` (and optionally an explicit `start` if you need to guarantee the daemon is running) to avoid relying on `--now` support and to keep docs/UX consistent.

Addressed: the generated enrollment command now enables the service without redundantly starting it, matching the installer lifecycle.

client/src/protoFleet/features/settings/components/Nodes/EnrollNodeModal.test.tsx
```diff
@@ -44,7 +44,11 @@ describe("EnrollNodeModal", () => {
     const { getByText } = render(<EnrollNodeModal open onDismiss={mockOnDismiss} onUpdated={mockOnUpdated} />);
 
     await waitFor(() => {
-      expect(getByText(/fleetnode enroll --server-url=/)).toBeInTheDocument();
+      expect(
+        getByText(
+          "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll --server-url=http://localhost:4000 && sudo systemctl enable --now fleet-node.service",
+        ),
+      ).toBeInTheDocument();
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> This test asserts the entire enrollment command as an exact string. Since `buildFleetNodeEnrollCommand` is already covered by `enrollNodeCommand.test.ts`, this makes the modal test unnecessarily brittle (any harmless formatting/ordering change will break it twice). Consider asserting key substrings (or using a predicate matcher) so this test focuses on rendering the command rather than duplicating the command-builder contract.

Addressed: the modal test now verifies the rendered command contains the key packaged binary, state directory, enrollment URL, and service lifecycle fragments without duplicating the full command.

client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
```diff
@@ -2,6 +2,8 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
+const FLEET_NODE_ENROLL_COMMAND = "sudo -u fleetnode /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll";
+const FLEET_NODE_ENABLE_COMMAND = "sudo systemctl enable fleet-node.service";
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> If the host reboots between installation and enrollment, or the service has otherwise stopped, this command enrolls successfully but leaves the Fleet Node offline: the installer deliberately disables a fresh unit and only starts it for the current boot (`deployment-files/fleetnode/install-fleet-node.sh` lines 318-323), while `systemctl enable` does not activate it.

Addressed: the one-step enrollment command retains `systemctl enable --now`, so a successfully enrolled node is enabled for future boots and is brought online immediately even if the installer-started service had stopped.

client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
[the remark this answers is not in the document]
ankitgoswami: 🤖

Follow-up correction: a later review identified the reboot/stopped-service case. The final command retains `systemctl enable --now fleet-node.service`; unlike the installer's immediate post-install guidance, enrollment must also guarantee that the newly enrolled node comes online. The installer does not keep the unit enabled before enrollment.

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -359,6 +381,8 @@ cleanup() {
 }
 trap cleanup EXIT
 
+validate_existing_enroll_helper
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Restrict this replacement check to installs; on uninstall, remove the global helper only when it matches the managed copy and otherwise preserve it while continuing to remove Fleet Node.

Addressed in `23c5e2b76`: helper ownership validation now applies only during installation. Uninstall removes `fleetnode-enroll` only when it is a regular non-symlink file matching the packaged helper; unrelated files and symlinks are preserved while Fleet Node is removed. Added regression coverage for damaged payloads and unmanaged helper paths.

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+for command in runuser systemctl; do
+  command -v "$command" >/dev/null 2>&1 || { echo "required command not found: $command" >&2; exit 1; }
+done
+[[ -x /opt/fleetnode/fleetnode ]] || { echo "Fleet Node is not installed at /opt/fleetnode/fleetnode" >&2; exit 1; }
+
+runuser -u fleetnode -- \
+  /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode enroll "$@"
+systemctl enable --now fleet-node.service
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Detect partial versus complete state and resume with refresh or service activation so the supported command remains retryable.

Addressed in `23c5e2b76`: the helper now checks persisted state through `fleetnode status`. Partial enrollment resumes with `fleetnode refresh`, completed enrollment retries service activation, and `--force` still explicitly re-enrolls. Enrollment or refresh failures do not call systemd. Added hermetic coverage for each path, including retry after a `systemctl` failure.

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,86 @@
...
+  local arg
+  for arg in "$@"; do
+    [[ "$arg" == "--force" ]] && return 0
+  done
+  return 1
+}
+
+enroll_or_resume() {
+  if has_force_flag "$@" || [[ ! -f "$STATE_PATH" ]]; then
+    run_fleetnode enroll "$@"
```
rl-block, line 62: The `--force` path goes straight to `enroll` while the service may still be running. The daemon holds the state lock for its lifetime (`run.go` wraps `runLocked` in `bootstrap.WithStateLock`) and `enroll` takes the same lock with `LOCK_EX|LOCK_NB`, so on an enrolled, running node this fails with the "stop it (kill N)" error -- and `--force` is exactly what the helper's own hint at :43 tells operators to try.

Should we stop the unit first when `--force` is present (and restart it if enroll then fails), or at least check `systemctl is-active` and fail fast with "stop fleet-node.service before re-enrolling"? It might also be worth listing `[--force]` in `usage()`.

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,86 @@
...
+  fleet_node_id=$(awk '$1 == "fleet_node_id:" { print $2 }' <<<"$status")
+  api_key_present=$(awk '$1 == "api_key_present:" { print $2 }' <<<"$status")
+  if [[ ! "$fleet_node_id" =~ ^[1-9][0-9]*$ ]]; then
+    echo "Fleet Node state has no fleet_node_id; pass --force to re-enroll" >&2
+    return 1
+  fi
+
+  case "$api_key_present" in
+    false) run_fleetnode refresh ;;
+    true) return 0 ;;
```
rl-block: On this branch `"$@"` is never consulted, so an operator who re-runs the UI command with a *different* `--server-url` (say, after moving servers) gets a silent exit 0 and the service is enabled against the old server. `fleetnode status` already prints `server_url:`, so a comparison is cheap.

Maybe parse `server_url` from the status output, compare it to the requested `--server-url` when one is given, and fail with "already enrolled with X; pass --force to re-enroll with Y"? A one-line "already enrolled (fleet_node_id=N); enabling service" notice here would also make the no-op visible. A mock-status case with a differing `server_url:` in `test-fleetnode-enroll.sh` would cover it.

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,86 @@
...
+      ;;
+  esac
+}
+
+main() {
+  case "${1:-}" in
+    -h|--help)
+      usage
+      return 0
+      ;;
```
rl-block: Only `$1` is checked for help, and `main` enables the unit on *any* zero exit from the child. kong runs the `--help` hook before validation and exits 0, so `sudo fleetnode-enroll --server-url=X --help` prints the Go CLI's help and then runs `systemctl enable --now` on an unenrolled host (which then exits 78 and lands in `failed`).

Two small hardenings we could consider: scan all of `"$@"` for `-h`/`--help` (like `has_force_flag`), and gate the `systemctl enable --now` at :81 on evidence rather than exit status -- e.g. re-run `status` and require `api_key_present: true`, or at minimum `[[ -f "$STATE_PATH" ]]`. A test where the mock exits 0 without writing state would pin this down.

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -270,6 +274,18 @@ validate_preserved_access() {
   done
 }
 
+validate_existing_enroll_helper() {
+  [[ -e "$ENROLL_HELPER_PATH" || -L "$ENROLL_HELPER_PATH" ]] || return 0
+  [[ -f "$ENROLL_HELPER_PATH" && ! -L "$ENROLL_HELPER_PATH" ]] || {
+    echo "enrollment helper path is not a regular file: $ENROLL_HELPER_PATH" >&2
+    return 1
+  }
+  [[ -f "$PROGRAM_DIR/fleetnode-enroll" ]] && cmp -s "$PROGRAM_DIR/fleetnode-enroll" "$ENROLL_HELPER_PATH" || {
```
rl-block: Because "managed" is defined only as byte-equal to `$PROGRAM_DIR/fleetnode-enroll`, a helper whose `/opt` twin is gone (damaged uninstall, failed rollback restore at :354, or a manual `rm -rf /opt/fleetnode`) is refused as "unmanaged" on every future install, and uninstall leaves it behind -- with no hint about what to do. The installer tests currently work around this with `rm -f` (~:681, :695) rather than asserting the behaviour.

Perhaps add a remediation hint to this message (`... remove it with: sudo rm $ENROLL_HELPER_PATH`)? If we want to be more forgiving, a fixed marker comment in `fleetnode-enroll` (e.g. `# managed-by: proto-fleet install-fleet-node.sh`) would let both this check and `remove_fleet_node` (:413-417) fall back to marker matching when the `/opt` copy is missing.

deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -75,7 +80,7 @@ EOF
 sudo chown fleetnode:fleetnode /var/lib/fleetnode/state.yaml
 sudo chmod 0600 /var/lib/fleetnode/state.yaml
 
-sudo systemctl enable --now fleet-node.service
+sudo fleetnode-enroll --server-url=http://127.0.0.1:1 --allow-insecure-transport
```
rl-block: This is the only real-binary run of the helper, and the synthetic state above has `fleet_node_id: 1` plus `api_key: qualification`, so it can only take the `true) return 0` branch. `enroll`, `refresh`, `--force`, and the malformed-status branch are exercised solely against the bash mock in `test-fleetnode-enroll.sh`, which hardcodes today's `fleet_node_id:` / `api_key_present:` keys -- if `status.go` ever renames them, PR CI stays green and only the nightly would notice.

Could we add a couple of real-binary negative cases here that need neither a server nor stdin? For example: a run with `--server-url=http://127.0.0.1:2` expecting non-zero (once the mismatch check lands), a `--force </dev/null` run expecting the enrollment-code prompt failure rather than a lock error, and an unparseable `state.yaml` expecting the helper to fail and leave the unit disabled. A keep-in-sync comment between `status.go:29-31` and the awk at `fleetnode-enroll:40-41` would help too.

client/src/protoFleet/features/settings/components/Nodes/enrollNodeCommand.ts
```diff
@@ -2,6 +2,7 @@ import { API_PROXY_BASE } from "@/protoFleet/api/constants";
 
 const INSECURE_TRANSPORT_FLAG = "--allow-insecure-transport";
 const HTTP_FLEET_API_PORT = "4000";
+const FLEET_NODE_ENROLL_COMMAND = "sudo fleetnode-enroll";
```
rl-block: The helper is installed only at `/usr/local/bin`, and the RHEL-family default sudoers `secure_path` (`/sbin:/bin:/usr/sbin:/usr/bin`) doesn't include it, so on those hosts `sudo fleetnode-enroll` is "command not found"; CI only qualifies on Ubuntu, where it works. The installer's support statement is a generic "Linux only".

If we don't want to restrict to Debian-family, maybe render the absolute path everywhere the command is shown (`sudo /usr/local/bin/fleetnode-enroll` here, in the installer banner at `install-fleet-node.sh:627`, and in the helper's `usage()`), and update the vitest expectations and the qualify grep accordingly?

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,86 @@
...
+  if has_force_flag "$@" || [[ ! -f "$STATE_PATH" ]]; then
+    run_fleetnode enroll "$@"
+    return
+  fi
+
+  local status fleet_node_id api_key_present
+  status=$(run_fleetnode status) || return
+  fleet_node_id=$(awk '$1 == "fleet_node_id:" { print $2 }' <<<"$status")
+  api_key_present=$(awk '$1 == "api_key_present:" { print $2 }' <<<"$status")
+  if [[ ! "$fleet_node_id" =~ ^[1-9][0-9]*$ ]]; then
```
rl-block: None of the fail-closed branches in `enroll_or_resume` / `main` are exercised by `test-fleetnode-enroll.sh`: `status` failing (:39), a zero/missing `fleet_node_id` (here), a malformed `api_key_present` (:50-53), no-args (:63-66), and the non-root refusal (:69-72). The mock already has a `MOCK_FLEET_NODE_ID` knob but nothing sets it.

It might be worth a handful of cases asserting non-zero exit and no `systemctl` call: `MOCK_FLEET_NODE_ID=0`, `MOCK_API_KEY_PRESENT=maybe`, a `MOCK_STATUS_EXIT=1` knob, `main` with no args, and an `id()` override returning 1000.

deployment-files/fleetnode/fleetnode-enroll
```diff
@@ -0,0 +1,86 @@
...
+    echo "Fleet Node state has no fleet_node_id; pass --force to re-enroll" >&2
+    return 1
+  fi
+
+  case "$api_key_present" in
+    false) run_fleetnode refresh ;;
```
rl-block: When the operator rejects a pending node, the UI says "Start over with a new enrollment code" -- but `enroll` persists `fleet_node_id` before the api_key prompt (`enroll.go:79-81`), so a re-run of the pasted command lands here and `refresh` asks for an *api_key* the rejected node will never get. The failure text says "retry", and `--force` (the only way out) isn't in `usage()` or surfaced on this path. The same dead-end appears after a Ctrl-C at the api_key prompt once the UI session expires.

Perhaps print a short explanation before `refresh` ("resuming interrupted enrollment for fleet_node_id=N; if you rejected that node or started over, re-run with --force"), and repeat the hint if `refresh` exits non-zero?

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -331,6 +349,13 @@ cleanup() {
     if [[ "$service_stopped" == "1" ]]; then
       as_root "$SYSTEMCTL" stop fleet-node.service || rollback_error "stop candidate Fleet Node service"
     fi
+    if [[ "$enroll_helper_replaced" == "1" ]]; then
+      if [[ -f "$enroll_helper_backup" ]]; then
+        as_root install -m 0755 "$enroll_helper_backup" "$ENROLL_HELPER_PATH" || rollback_error "restore previous enrollment helper"
+      else
+        as_root rm -f "$ENROLL_HELPER_PATH" || rollback_error "remove candidate enrollment helper"
```
rl-block: This `rm -f` arm (a fresh install or pre-helper upgrade that fails after the helper is placed) is never executed in `test-install-fleetnode.sh`: the fresh-install failure cases abort before :599, and the failed v1.3.0 upgrades have a v1.1.0 helper so they hit the restore arm above. A regression here would leave `/usr/local/bin/fleetnode-enroll` orphaned and, per the check at :283, lock out reinstalls.

Maybe a `FAKE_SYSTEMCTL_FAIL_DAEMON_RELOAD` knob in the fake systemctl, then after an uninstall: a failing fresh install asserting neither the helper nor `/opt/fleetnode` exists, followed by a plain install that succeeds?

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -581,9 +624,7 @@ echo "Configuration: $CONFIG_DIR"
 echo "Protected state: $STATE_DIR"
 echo
 echo "Enroll:"
-echo "  sudo -u fleetnode $PROGRAM_DIR/fleetnode --state-dir $STATE_DIR enroll --server-url=https://YOUR-FLEET-SERVER"
-echo "Then enable and start the service:"
-echo "  sudo systemctl enable --now fleet-node.service"
+echo "  sudo fleetnode-enroll --server-url=https://YOUR-FLEET-SERVER"
```
rl-block: nit: while updating the banner here, the `usage()` heredoc at :35-36 still says "enroll the node and then enable and start the service". With the helper owning enable/start, maybe point it at `sudo fleetnode-enroll` too so `--help`, this banner, and the UI all agree?

mcharles-square (approved):

rl-block (approved):

deployment-files/fleetnode/fleetnode-enroll
ankitgoswami: 🤖

> The `--force` path goes straight to `enroll` while the service may still be running.

Addressed in `803f68913`: `--force` now stops `fleet-node.service` before enrollment, propagates a stop failure, and only enables/starts the service after successful enrollment. The usage text now documents `--force`; focused tests cover stop-before-enroll ordering and stop failure.

deployment-files/fleetnode/fleetnode-enroll
ankitgoswami: 🤖

> An operator who re-runs the UI command with a different `--server-url` gets a silent exit 0 and the service is enabled against the old server.

Addressed in `803f68913`: existing-state retries now compare the requested URL with `server_url` from `fleetnode status`. A mismatch reports both URLs, directs the operator to `--force`, and prevents refresh, enrollment, or service activation. Matching completed state reports the fleet node ID before enabling the service.

PR #990, opened 2026-09-01
Make Fleet Node startup readiness explicit

ankitgoswami:
Reviewable diff: +91/-41 across 6 files (excludes generated, test, and story files).

## Summary

Fleet Node now exposes a real systemd readiness boundary and stops restart loops when operator action is required. Fresh installs remain stopped until enrollment, while upgrades only succeed after the replacement process finishes local initialization and reports ready.

**Stack**

1. **[#990](https://github.com/block/proto-fleet/pull/990) — this PR: startup readiness and safe upgrade boundary**
2. [#991](https://github.com/block/proto-fleet/pull/991) — installer lifecycle: preflight, uninstall, purge, and idempotency
3. [#992](https://github.com/block/proto-fleet/pull/992) — native package qualification before artifact upload

The later PRs build on this readiness contract. Installer lifecycle breadth and release-artifact qualification are intentionally deferred to [#991](https://github.com/block/proto-fleet/pull/991) and [#992](https://github.com/block/proto-fleet/pull/992); runbooks and broader distribution work remain out of this stack.

## How it works

1. `fleetnode run` validates local state, initializes plugins, and makes its first gateway attempt.
2. Once local startup is usable, it sends `READY=1` to systemd. A temporarily unavailable Fleet server does not prevent the local daemon from becoming ready and retrying.
3. Invalid local state, revoked enrollment, missing server-side identity, or permanent plugin/bootstrap failures exit with code 78.
4. The unit uses `Type=notify`, waits up to 300 seconds for readiness, and does not restart exit code 78.
5. During an upgrade, the installer starts the replacement synchronously. If it never becomes ready, the prior payload and unit are restored and restarted.
6. A fresh install is left stopped and disabled so the operator can enroll before enabling it.

```mermaid
flowchart TD
  A["systemd starts Fleet Node"] --> B["Validate state and initialize plugins"]
  B --> C{"Local startup usable?"}
  C -->|"Yes"| D["Send READY=1"]
  D --> E["Run heartbeat and control loops"]
  C -->|"Operator action required"| F["Exit 78"]
  C -->|"Other startup failure"| G["Exit non-zero"]
  F --> H["systemd leaves service stopped"]
  G --> I["systemd may restart service"]
```

```mermaid
sequenceDiagram
  participant Installer
  participant Systemd
  participant Candidate as Fleet Node candidate
  participant Previous as Previous payload
  Installer->>Systemd: start fleet-node.service
  Systemd->>Candidate: launch
  Candidate-->>Systemd: READY=1 after local initialization
  Systemd-->>Installer: start succeeded
  alt candidate cannot become ready
    Systemd-->>Installer: start failed or timed out
    Installer->>Previous: restore payload and unit
    Installer->>Systemd: start previous service
  end
```

## Areas of the code involved

| --- | --- | --- |
| `server/cmd/fleetnode/` | Added readiness notification and an operator-action exit classification | Defines when Fleet Node is safe to consider started and when retries must stop |
| `deployment-files/fleetnode/fleet-node.service` | Switched to `Type=notify`, added startup timeout and exit-78 restart prevention | Establishes the service-manager contract |
| `deployment-files/fleetnode/install-fleet-node.sh` | Leaves fresh installs stopped and blocks upgrades on systemd readiness | Makes installer success mean the replacement is locally runnable |
| `server/go.mod` | Promoted the existing systemd module to a direct dependency | Supports native readiness notification |
| Tests | Added readiness, permanent-error, fresh-install, and rollback coverage | Guards the lifecycle boundary; review-light |

- Readiness means local initialization completed, not that the Fleet server is currently reachable; transient network outages continue retrying after startup.
- Exit code 78 is reserved for states that need enrollment or operator repair, preventing an unbounded restart loop while preserving ordinary restart-on-failure behavior.
- Fresh installation does not auto-start because an unenrolled node cannot run successfully; upgrades still start synchronously so rollback remains possible.

## Testing & validation

- `go test ./server/cmd/fleetnode`
- `bash deployment-files/fleetnode/tests/test-install-fleetnode.sh`
- Shell syntax and diff checks
- Native amd64/arm64 package installation is intentionally covered by PR 3, not this PR.

### DIFF
--- a/deployment-files/fleetnode/fleet-node.service
+++ b/deployment-files/fleetnode/fleet-node.service
@@ -5,7 +5,8 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=simple
+Type=notify
+NotifyAccess=main
 User=fleetnode
 Group=fleetnode
 Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
@@ -13,6 +14,8 @@ EnvironmentFile=-/etc/fleetnode/fleetnode.env
 ExecStart=/usr/bin/env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode run
 Restart=on-failure
 RestartSec=5s
+RestartPreventExitStatus=78
+TimeoutStartSec=120s
 TimeoutStopSec=90s
 NoNewPrivileges=true
 PrivateTmp=true
--- a/deployment-files/fleetnode/fleet-node.service
+++ b/deployment-files/fleetnode/fleet-node.service
@@ -15,7 +15,7 @@ ExecStart=/usr/bin/env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/s
 Restart=on-failure
 RestartSec=5s
 RestartPreventExitStatus=78
-TimeoutStartSec=120s
+TimeoutStartSec=180s
 TimeoutStopSec=90s
 NoNewPrivileges=true
 PrivateTmp=true
--- a/deployment-files/fleetnode/fleet-node.service
+++ b/deployment-files/fleetnode/fleet-node.service
@@ -15,7 +15,7 @@ ExecStart=/usr/bin/env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/s
 Restart=on-failure
 RestartSec=5s
 RestartPreventExitStatus=78
-TimeoutStartSec=180s
+TimeoutStartSec=300s
 TimeoutStopSec=90s
 NoNewPrivileges=true
 PrivateTmp=true
--- a/deployment-files/fleetnode/install-fleet-node.sh
+++ b/deployment-files/fleetnode/install-fleet-node.sh
@@ -174,6 +174,9 @@ cleanup() {
     rollback_failed=1
   }
   if [[ "$install_complete" != "1" ]]; then
+    if [[ "$service_stopped" == "1" ]]; then
+      as_root "$SYSTEMCTL" stop fleet-node.service || rollback_error "stop candidate Fleet Node service"
+    fi
     if [[ "$unit_replaced" == "1" ]]; then
       if [[ -f "$unit_backup" ]]; then
         as_root install -m 0644 "$unit_backup" "$UNIT_PATH" || rollback_error "restore previous systemd unit"
@@ -275,8 +278,10 @@ if ! grep -Fxq "version: $VERSION" "$source_dir/version.txt"; then
 fi
 
 if [[ "$fresh_install" == "0" ]]; then
-  as_root "$SYSTEMCTL" stop fleet-node.service
-  service_stopped=1
+  if as_root "$SYSTEMCTL" is-active --quiet fleet-node.service; then
+    as_root "$SYSTEMCTL" stop fleet-node.service
+    service_stopped=1
+  fi
 fi
 
 as_root install -d -m 0755 "${PROGRAM_DIR%/*}" "${UNIT_PATH%/*}"
@@ -332,7 +337,9 @@ else
   # Type=notify keeps this call blocked until Fleet Node finishes local
   # initialization and reports READY=1, so the previous payload remains
   # available to the EXIT trap until the candidate is actually runnable.
-  as_root "$SYSTEMCTL" start fleet-node.service
+  if [[ "$service_stopped" == "1" ]]; then
+    as_root "$SYSTEMCTL" start fleet-node.service
+  fi
 fi
 
 install_complete=1
--- a/deployment-files/fleetnode/install-fleet-node.sh
+++ b/deployment-files/fleetnode/install-fleet-node.sh
@@ -21,8 +21,8 @@ Usage: install-fleet-node.sh VERSION
 Install one exact Proto Fleet Node release on Linux. VERSION must be a
 release-specific tag such as v1.2.3 or nightly-20260825-0123456789ab.
 
-The installer starts the service but leaves a fresh installation disabled at
-boot. After installation, enroll the node and then enable the service.
+The installer leaves a fresh installation stopped and disabled at boot. After
+installation, enroll the node and then enable and start the service.
 EOF
 }
 
@@ -167,23 +167,29 @@ install_complete=0
 
 cleanup() {
   status=$?
+  set +e
+  rollback_failed=0
+  rollback_error() {
+    echo "rollback failed: $*" >&2
+    rollback_failed=1
+  }
   if [[ "$install_complete" != "1" ]]; then
     if [[ "$unit_replaced" == "1" ]]; then
       if [[ -f "$unit_backup" ]]; then
-        as_root install -m 0644 "$unit_backup" "$UNIT_PATH"
+        as_root install -m 0644 "$unit_backup" "$UNIT_PATH" || rollback_error "restore previous systemd unit"
       else
-        as_root rm -f "$UNIT_PATH"
+        as_root rm -f "$UNIT_PATH" || rollback_error "remove candidate systemd unit"
       fi
-      as_root "$SYSTEMCTL" daemon-reload >/dev/null 2>&1 || true
+      as_root "$SYSTEMCTL" daemon-reload || rollback_error "reload systemd after restoring previous unit"
     fi
     if [[ "$program_replaced" == "1" ]]; then
-      as_root rm -rf "$PROGRAM_DIR"
+      as_root rm -rf "$PROGRAM_DIR" || rollback_error "remove candidate Fleet Node payload"
     fi
     if [[ "$previous_saved" == "1" && -d "$previous" ]]; then
-      as_root mv "$previous" "$PROGRAM_DIR"
+      as_root mv "$previous" "$PROGRAM_DIR" || rollback_error "restore previous Fleet Node payload"
     fi
     if [[ "$service_stopped" == "1" ]]; then
-      as_root "$SYSTEMCTL" start fleet-node.service >/dev/null 2>&1 || true
+      as_root "$SYSTEMCTL" start fleet-node.service || rollback_error "restart previous Fleet Node service"
     fi
   fi
   if [[ -n "$INSTALL_LOCK_PID" ]]; then
@@ -194,6 +200,9 @@ cleanup() {
   as_root rm -rf "$work_dir" "$incoming"
   if [[ "$install_complete" == "1" ]]; then
     as_root rm -rf "$previous"
+  fi
+  if [[ "$rollback_failed" == "1" ]]; then
+    status=1
   fi
   exit "$status"
 }
@@ -319,8 +328,12 @@ if [[ "$fresh_install" == "1" ]]; then
   if as_root "$SYSTEMCTL" is-enabled --quiet fleet-node.service; then
     as_root "$SYSTEMCTL" disable fleet-node.service
   fi
+else
+  # Type=notify keeps this call blocked until Fleet Node finishes local
+  # initialization and reports READY=1, so the previous payload remains
+  # available to the EXIT trap until the candidate is actually runnable.
+  as_root "$SYSTEMCTL" start fleet-node.service
 fi
-as_root "$SYSTEMCTL" start fleet-node.service
 
 install_complete=1
 echo "Fleet Node $VERSION installed."
@@ -329,8 +342,8 @@ echo "Protected state: $STATE_DIR"
 echo
 echo "Enroll:"
 echo "  sudo -u fleetnode $PROGRAM_DIR/fleetnode --state-dir $STATE_DIR enroll --server-url=https://YOUR-FLEET-SERVER"
-echo "Then enable at boot:"
-echo "  sudo systemctl enable fleet-node.service"
+echo "Then enable and start the service:"
+echo "  sudo systemctl enable --now fleet-node.service"
 echo "Inspect:"
 echo "  systemctl status fleet-node.service"
 echo "  journalctl -u fleet-node.service"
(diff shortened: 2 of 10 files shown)


server/cmd/fleetnode/run.go
```diff
@@ -102,20 +110,20 @@ func (r *RunCmd) run(c *Context, logOutput io.Writer) error {
...
-			return resolveErr
+			return operatorActionRequired(resolveErr)
 		}
 		resolvedPluginsDir = resolved
 	}
 	path := bootstrap.StatePath(c.StateDir)
 	st, exists, err := bootstrap.LoadState(path)
 	if err != nil {
-		return err
+		return operatorActionRequired(err)
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > Classify the outer lock error with `operatorActionRequired` as well.

Addressed in `654e723ed`: state-lock preparation and acquisition errors now exit 78, while errors returned by `runLocked` retain their original classification. Added regression coverage for both paths.

deployment-files/fleetnode/fleet-node.service
```diff
@@ -5,14 +5,17 @@ Wants=network-online.target
...
+NotifyAccess=main
 User=fleetnode
 Group=fleetnode
 Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 EnvironmentFile=-/etc/fleetnode/fleetnode.env
 ExecStart=/usr/bin/env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /opt/fleetnode/fleetnode --state-dir /var/lib/fleetnode run
 Restart=on-failure
 RestartSec=5s
+RestartPreventExitStatus=78
+TimeoutStartSec=120s
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > Set `TimeoutStartSec` above the complete worst-case path or move the readiness boundary before the network attempt.

Addressed in `654e723ed`: raised `TimeoutStartSec` to 180 seconds to cover plugin initialization, both possible refresh calls, and the initial heartbeat. Updated the installer assertion accordingly.

deployment-files/fleetnode/fleet-node.service
```diff
@@ -5,14 +5,17 @@ Wants=network-online.target
...
 TimeoutStopSec=90s
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 > Consider increasing `TimeoutStartSec` so the systemd contract matches the worst-case startup path.

Addressed in `654e723ed`: raised `TimeoutStartSec` to 180 seconds to cover plugin initialization, both possible refresh calls, and the initial heartbeat. Updated the installer assertion accordingly.

server/cmd/fleetnode/run.go
```diff
@@ -342,14 +355,25 @@ func (r *RunCmd) refreshAndSave(ctx context.Context, st *bootstrap.State, path s
 	return nil
 }
 
+func isRetryableRefreshError(err error) bool {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var connectErr *connect.Error
+	return errors.As(err, &connectErr)
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Restrict retries to transient Connect codes; permanent codes such as Unauthenticated and InvalidArgument must not be retried.

Addressed in `7ad9b6a10`: `isRetryableRefreshError` now retries only Canceled, DeadlineExceeded, ResourceExhausted, Aborted, Internal, and Unavailable Connect codes while preserving direct context cancellation and deadline handling. Focused table tests cover transient, permanent, wrapped, and non-Connect errors.

server/cmd/fleetnode/run.go
```diff
@@ -342,14 +355,34 @@ func (r *RunCmd) refreshAndSave(ctx context.Context, st *bootstrap.State, path s
...
+		return false
+	}
+	code := connectErr.Code()
+	return code == connect.CodeCanceled ||
+		code == connect.CodeDeadlineExceeded ||
+		code == connect.CodeResourceExhausted ||
+		code == connect.CodeAborted ||
+		code == connect.CodeInternal ||
+		code == connect.CodeUnavailable
+}
```
rl-block, line 382: Since this is now an allowlist, anything outside it becomes exit 78 and systemd stops retrying. Two codes connect-go produces for non-Connect HTTP responses look infra-transient rather than operator-fixable: `CodeUnknown` (any HTTP status outside 429/502/503/504 with a non-Connect body, e.g. an nginx/LB 500 error page) and `CodeUnimplemented` (HTTP 404, e.g. a proxy misrouting mid-deploy).

Should we add `CodeUnknown` (and perhaps `CodeUnimplemented`) here, or flip this to a denylist of codes we're confident are permanent (`Unauthenticated`, `PermissionDenied`, `InvalidArgument`, `NotFound`, `FailedPrecondition`)? The mapping is `httpToCode` in connect-go's `protocol.go` if useful.

server/cmd/fleetnode/run.go
```diff
@@ -342,14 +355,34 @@ func (r *RunCmd) refreshAndSave(ctx context.Context, st *bootstrap.State, path s
...
 func (r *RunCmd) tick(ctx context.Context, client gatewayClient, st *bootstrap.State, path string, logger *slog.Logger) error {
 	if r.sessionNeedsRefresh(st) {
 		if err := r.refreshAndSave(ctx, st, path, logger); err != nil {
 			if errors.Is(err, bootstrap.ErrBeginAuthRejected) {
-				return fmt.Errorf("%w. The server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Exiting; re-enroll once the operator-side cause is resolved", bootstrap.ErrBeginAuthRejected)
+				return operatorActionRequired(fmt.Errorf("%w. The server returns Unauthenticated for any of: revoked api_key, identity_pubkey mismatch, expired challenge, or server clock drift. Exiting; local credentials are preserved, re-enroll once the operator-side cause is resolved", bootstrap.ErrBeginAuthRejected))
+			}
+			if !isRetryableRefreshError(err) {
+				return operatorActionRequired(fmt.Errorf("session refresh requires operator action: %w", err))
 			}
```
rl-block, line 395: One transient case that lands here: `CompleteAuthHandshake` returns `Unauthenticated` for "challenge expired or not found", and the server's challenge TTL (30s) equals the per-step handshake timeout, so a slow second RPC can trip it once. Before this PR that was retried on the next tick; now it exits 78 and the node stays down until someone runs `systemctl start`.

Maybe worth a small escalation threshold (e.g. only exit 78 after N consecutive non-retryable refresh failures), so a one-off blip doesn't strand a remote node while a real key mismatch still stops the loop quickly?

server/cmd/fleetnode/run.go
```diff
@@ -133,29 +144,34 @@ func (r *RunCmd) run(c *Context, logOutput io.Writer) error {
...
-		return r.runLocked(ctx, c, resolvedPluginsDir, logger)
-	})
+	var runErr error
+	if err := bootstrap.WithStateLock(c.StateDir, func() error {
+		runErr = r.runLocked(ctx, c, resolvedPluginsDir, logger)
+		return nil
+	}); err != nil {
+		return operatorActionRequired(err)
+	}
+	return runErr
```
rl-block, line 155: Nit: the `return nil` inside the callback is doing real work (keeping `runLocked`'s classification separate from lock-acquisition errors), but it reads like a dropped error at first glance. A one-line comment might save the next reader a detour.

server/cmd/fleetnode/exitcode.go
```diff
@@ -0,0 +1,24 @@
+package main
+
+import "errors"
+
+const operatorActionExitCode = 78
```
rl-block, line 6: Nit: it might be worth a short comment noting this is `EX_CONFIG` from `sysexits.h` and that `fleet-node.service` pins it via `RestartPreventExitStatus=78`, so the two stay in sync if anyone touches either side.

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -167,23 +167,29 @@ install_complete=0
...
     fi
     if [[ "$previous_saved" == "1" && -d "$previous" ]]; then
-      as_root mv "$previous" "$PROGRAM_DIR"
+      as_root mv "$previous" "$PROGRAM_DIR" || rollback_error "restore previous Fleet Node payload"
     fi
     if [[ "$service_stopped" == "1" ]]; then
-      as_root "$SYSTEMCTL" start fleet-node.service >/dev/null 2>&1 || true
+      as_root "$SYSTEMCTL" start fleet-node.service || rollback_error "restart previous Fleet Node service"
     fi
   fi
```
rl-block, line 197: +1 to the open Codex thread on the pending auto-restart. Since a failed `start` leaves the unit in `activating (auto-restart)` for non-78 exits and timeouts, maybe the rollback branch could begin with `as_root "$SYSTEMCTL" stop fleet-node.service` when `service_stopped == 1`. That cancels the pending restart and also makes the `rm -rf "$PROGRAM_DIR"` below safe against a still-running candidate. The fake `systemctl` in the test could then assert the `stop -> start` ordering.

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -319,8 +328,12 @@ if [[ "$fresh_install" == "1" ]]; then
   if as_root "$SYSTEMCTL" is-enabled --quiet fleet-node.service; then
     as_root "$SYSTEMCTL" disable fleet-node.service
   fi
+else
+  # Type=notify keeps this call blocked until Fleet Node finishes local
+  # initialization and reports READY=1, so the previous payload remains
+  # available to the EXIT trap until the candidate is actually runnable.
+  as_root "$SYSTEMCTL" start fleet-node.service
 fi
```
rl-block, line 343: Do we want upgrades to be blocked when the node is already stopped or sitting in an exit-78 state (e.g. revoked, not yet re-enrolled)? In that case the candidate exits 78, `start` fails, rollback runs, and the rollback `start` also fails, so the installer exits 1 with "rollback failed" even though the payload swap itself was fine. An operator would have to re-enroll on the old binary before they can upgrade.

If that's the intended contract, perhaps just call it out in the usage text. Otherwise we could gate on readiness only when the service was active before the `stop` (the test currently forbids `is-active`, so that would need revisiting).

rl-block (approved):

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -319,8 +328,12 @@ if [[ "$fresh_install" == "1" ]]; then
   if as_root "$SYSTEMCTL" is-enabled --quiet fleet-node.service; then
     as_root "$SYSTEMCTL" disable fleet-node.service
   fi
+else
+  # Type=notify keeps this call blocked until Fleet Node finishes local
+  # initialization and reports READY=1, so the previous payload remains
+  # available to the EXIT trap until the candidate is actually runnable.
+  as_root "$SYSTEMCTL" start fleet-node.service
```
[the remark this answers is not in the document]
ankitgoswami: 🤖 AI agent response

> Stop the unit to cancel any pending restart before restoring the previous unit and payload.

Addressed in da5ee72c9: rollback now stops the candidate before restoring the previous unit and payload, cancelling any pending `Restart=on-failure`.

deployment-files/fleetnode/install-fleet-node.sh
ankitgoswami: 🤖 AI agent response

> The fake `systemctl` in the test could then assert the `stop -> start` ordering.

Addressed in da5ee72c9: the installer test now verifies that the failed candidate is stopped before the previous service is restarted.

deployment-files/fleetnode/install-fleet-node.sh
ankitgoswami: 🤖 AI agent response

> Otherwise we could gate on readiness only when the service was active before the `stop`.

Addressed in da5ee72c9: upgrades now check the existing service's active state and only stop/start it when it was active. An inactive or exit-78 node upgrades successfully and remains inactive.

server/cmd/fleetnode/run.go
ankitgoswami: 🤖 AI agent response

> Should we add `CodeUnknown` (and perhaps `CodeUnimplemented`) here, or flip this to a denylist?

Addressed in da5ee72c9: added `CodeUnknown` and `CodeUnimplemented` to the retryable allowlist. The explicit allowlist remains intact rather than broadening behavior with a denylist.

server/cmd/fleetnode/run.go
ankitgoswami: 🤖 AI agent response

> Maybe worth a small escalation threshold (e.g. only exit 78 after N consecutive non-retryable refresh failures)?

Addressed differently in da5ee72c9: the daemon retries the full refresh exactly once only when `CompleteAuthHandshake` returns `Unauthenticated`. `BeginAuthHandshake` rejection still exits immediately, and a second completion rejection preserves exit 78. This handles the demonstrated expired-challenge race without persistent counters or a state machine.

server/cmd/fleetnode/run.go
ankitgoswami: 🤖 AI agent response

> A one-line comment might save the next reader a detour.

Addressed in da5ee72c9: added a comment explaining that returning `nil` keeps `runLocked`'s exit classification separate from lock-acquisition errors.

server/cmd/fleetnode/exitcode.go
ankitgoswami: 🤖 AI agent response

> It might be worth a short comment noting this is `EX_CONFIG` from `sysexits.h` and that `fleet-node.service` pins it via `RestartPreventExitStatus=78`.

Addressed in da5ee72c9: documented exit 78 as `EX_CONFIG` and noted that it must stay synchronized with `RestartPreventExitStatus` in the systemd unit.

PR #992, opened 2026-09-01
Qualify Fleet Node release artifacts

ankitgoswami:
Reviewable diff: +36/-24 across 3 files (excludes generated, test, and story files).

## Summary

Fleet Node release artifacts are installed and exercised on native amd64 and arm64 systemd runners before upload. The customer archive contains only production plugins, and each executable must be a valid, statically linked ELF. The qualification stays intentionally narrow: it proves local startup, failed-upgrade rollback, and state-preserving uninstall without duplicating the full installer test suite.

**Stack**

1. [#990](https://github.com/block/proto-fleet/pull/990) — startup readiness and safe upgrade boundary (merged)
2. [#991](https://github.com/block/proto-fleet/pull/991) — installer lifecycle (merged)
3. **[#992](https://github.com/block/proto-fleet/pull/992) — this PR: native release-artifact qualification**
4. [#989](https://github.com/block/proto-fleet/pull/989) — one-command enrollment helper

With #990 and #991 merged, this PR's current diff is against `main`. It exercises their explicit readiness and rollback contracts using the packaged artifact. The enrollment helper remains isolated in #989; signing, release publication policy, and runbooks remain out of scope here.

## How it works

1. The artifact matrix builds directly on native `ubuntu-latest` amd64 and `ubuntu-24.04-arm` runners.
2. CI verifies that Fleet Node and all production plugins are valid ELF executables without dynamic interpreters or shared-library dependencies.
3. Packaging omits the virtual test plugin while retaining Fleet Node, Proto, Antminer, and asicrs binaries plus required configuration and unit files.
4. Before upload, the runner invokes the real installer as a non-root user. Only installer mutations cross `sudo`, matching the customer privilege boundary.
5. Qualification verifies that a fresh installation stays stopped, starts the service using synthetic local state, injects a broken plugin into an upgrade, and requires the previous running payload to be restored byte-for-byte.
6. A final uninstall must remove the program and unit while preserving Fleet Node state and its service account.
7. The exact archive and checksum are uploaded only after qualification succeeds.

```mermaid
flowchart LR
  A["Native amd64 or arm64 runner"] --> B["Build Fleet Node and production plugins"]
  B --> C["Validate ELF and static linkage"]
  C --> D["Create exact-version archive and checksum"]
  D --> E["Install through the customer installer"]
  E --> F["Start with real systemd"]
  F --> G["Inject failed upgrade and verify rollback"]
  G --> H["Uninstall while preserving state"]
  H --> I["Upload qualified artifact"]
```

```mermaid
sequenceDiagram
  participant CI
  participant Installer
  participant Systemd
  CI->>Installer: fresh install exact archive
  Installer-->>CI: stopped and disabled
  CI->>Systemd: start with synthetic local state
  Systemd-->>CI: active after local readiness
  CI->>Installer: upgrade with broken plugin
  Installer->>Systemd: candidate start fails
  Installer->>Systemd: restore and start previous payload
  Systemd-->>CI: previous service active
  CI->>Installer: uninstall
  Installer-->>CI: program removed and state preserved
```

## Areas of the code involved

| --- | --- | --- |
| `.github/workflows/proto-fleet-artifact-build.yml` | Builds on native runners, validates static executables, packages production contents, and qualifies before upload | Defines the release gate and exact customer artifact |
| `.github/workflows/deployment-config-checks.yml` | Syntax-checks the qualification script during PR CI | Catches shell parse failures before nightly or tagged builds |
| `deployment-files/fleetnode/install-fleet-node.sh` | Adds a constrained local-file systemd qualification mode and removes the virtual plugin from the accepted manifest | Reuses the customer installer without exposing production override hooks |
| `deployment-files/fleetnode/tests/qualify-package.sh` | Adds the focused native systemd lifecycle scenario | Proves the packaged binaries, unit, installer, and rollback work together; review-light |
| Installer tests | Updates the package fixture and qualification-mode guards | Prevents test-only content or broader override behavior from returning; review-light |

- Native execution replaces explicit ELF machine-string matching: wrong-architecture binaries fail by running, while CI avoids maintaining architecture-specific output text.
- Static-link validation remains explicit because a dynamically linked executable may work on the runner but fail on a customer host missing that library.
- The qualification seam accepts only a local `file://` source and rejects root-path, architecture, and systemctl overrides.
- The native scenario covers the high-value systemd and rollback boundary; successful upgrade variants and reinstall permutations remain in the fast installer suite.
- Synthetic state proves local daemon and plugin readiness, not real Fleet enrollment or server connectivity.

## Testing & validation

- `bash -n deployment-files/fleetnode/install-fleet-node.sh deployment-files/fleetnode/tests/qualify-package.sh deployment-files/fleetnode/tests/test-install-fleetnode.sh`
- `./deployment-files/fleetnode/tests/test-install-fleetnode.sh`
- `source ./bin/activate-hermit && just lint`
- Restacked #989: installer suite plus 13 focused enrollment-command client tests
- `git diff --check` for both stack layers
- Native amd64/arm64 systemd execution runs in nightly and tagged artifact workflows; it cannot be reproduced on this macOS workstation.

### DIFF
--- a/.github/workflows/proto-fleet-artifact-build.yml
+++ b/.github/workflows/proto-fleet-artifact-build.yml
@@ -267,19 +267,6 @@ jobs:
           go build -v -o server/fleetnode ./server/cmd/fleetnode
           go build -o server/proto-plugin ./plugin/proto
           go build -o server/antminer-plugin ./plugin/antminer
-          go build -o server/virtual-plugin ./plugin/virtual
-          cp plugin/virtual/config.json server/virtual-plugin.json
-
-          for binary in fleetnode proto-plugin antminer-plugin virtual-plugin; do
-            if ! readelf -h "server/$binary" >/dev/null; then
-              echo "::error::Unable to inspect Fleet Node binary $binary"
-              exit 1
-            fi
-            if readelf -l "server/$binary" | grep INTERP >/dev/null || readelf -d "server/$binary" | grep NEEDED >/dev/null; then
-              echo "::error::Fleet Node binary $binary must be statically linked"
-              exit 1
-            fi
-          done
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
@@ -293,6 +280,33 @@ jobs:
           cp /tmp/asicrs/asicrs-plugin server/asicrs-plugin
           cp /tmp/asicrs/asicrs-config.yaml server/asicrs-config.yaml
 
+      - name: Validate Fleet Node executables
+        env:
+          EXPECTED_ARCH: ${{ matrix.arch }}
+        run: |
+          case "$EXPECTED_ARCH" in
+            amd64) expected_machine="Advanced Micro Devices X86-64" ;;
+            arm64) expected_machine="AArch64" ;;
+            *) echo "::error::Unsupported architecture $EXPECTED_ARCH"; exit 1 ;;
+          esac
+
+          for binary in fleetnode proto-plugin antminer-plugin asicrs-plugin; do
+            path="server/$binary"
+            if ! header=$(readelf -h "$path"); then
+              echo "::error::Unable to inspect Fleet Node executable $binary"
+              exit 1
+            fi
+            machine=$(awk -F: '/Machine:/{sub(/^[[:space:]]+/, "", $2); print $2}' <<< "$header")
+            if [[ "$machine" != "$expected_machine" ]]; then
+              echo "::error::$binary targets '$machine', expected '$expected_machine'"
+              exit 1
+            fi
+            if readelf -l "$path" | grep INTERP >/dev/null || readelf -d "$path" | grep NEEDED >/dev/null; then
+              echo "::error::Fleet Node executable $binary must be statically linked"
+              exit 1
+            fi
+          done
+
       - name: Package Fleet Node
         working-directory: ./server
         run: |
@@ -304,8 +318,6 @@ jobs:
           install -m 0644 ../deployment-files/fleetnode/fleet-node.service "$package_dir/fleet-node.service"
           install -m 0755 proto-plugin "$package_dir/plugins/proto-plugin"
           install -m 0755 antminer-plugin "$package_dir/plugins/antminer-plugin"
-          install -m 0755 virtual-plugin "$package_dir/plugins/virtual-plugin"
-          install -m 0644 virtual-plugin.json "$package_dir/plugins/virtual-plugin.json"
           install -m 0755 asicrs-plugin "$package_dir/plugins/asicrs-plugin"
           install -m 0644 asicrs-config.yaml "$package_dir/plugins/asicrs-config.yaml"
           printf 'version: %s\n' "$VERSION" > "$package_dir/version.txt"
@@ -313,6 +325,14 @@ jobs:
           tar -czf "$archive" "$package_dir"
           sha256sum "$archive" > "$archive.sha256"
           sha256sum -c "$archive.sha256"
+
+      - name: Qualify Fleet Node package on systemd
+        env:
+          PACKAGE_DIR: ${{ github.workspace }}/server
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes nmap
+          deployment-files/fleetnode/tests/qualify-package.sh "$VERSION" "$PACKAGE_DIR"
 
       - name: Upload Fleet Node artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/proto-fleet-artifact-build.yml
+++ b/.github/workflows/proto-fleet-artifact-build.yml
@@ -281,24 +281,11 @@ jobs:
           cp /tmp/asicrs/asicrs-config.yaml server/asicrs-config.yaml
 
       - name: Validate Fleet Node executables
-        env:
-          EXPECTED_ARCH: ${{ matrix.arch }}
         run: |
-          case "$EXPECTED_ARCH" in
-            amd64) expected_machine="Advanced Micro Devices X86-64" ;;
-            arm64) expected_machine="AArch64" ;;
-            *) echo "::error::Unsupported architecture $EXPECTED_ARCH"; exit 1 ;;
-          esac
-
           for binary in fleetnode proto-plugin antminer-plugin asicrs-plugin; do
             path="server/$binary"
-            if ! header=$(readelf -h "$path"); then
+            if ! readelf -h "$path" >/dev/null; then
               echo "::error::Unable to inspect Fleet Node executable $binary"
-              exit 1
-            fi
-            machine=$(awk -F: '/Machine:/{sub(/^[[:space:]]+/, "", $2); print $2}' <<< "$header")
-            if [[ "$machine" != "$expected_machine" ]]; then
-              echo "::error::$binary targets '$machine', expected '$expected_machine'"
               exit 1
             fi
             if readelf -l "$path" | grep INTERP >/dev/null || readelf -d "$path" | grep NEEDED >/dev/null; then
@@ -324,7 +311,6 @@ jobs:
 
           tar -czf "$archive" "$package_dir"
           sha256sum "$archive" > "$archive.sha256"
-          sha256sum -c "$archive.sha256"
 
       - name: Qualify Fleet Node package on systemd
         env:
--- a/deployment-files/fleetnode/install-fleet-node.sh
+++ b/deployment-files/fleetnode/install-fleet-node.sh
@@ -11,6 +11,11 @@ INSTALL_LOCK_RELEASE_PATH=""
 ACTION=install
 VERSION=""
 
+case "$TEST_MODE" in
+  0|1|systemd) ;;
+  *) echo "invalid FLEETNODE_TEST_MODE" >&2; exit 1 ;;
+esac
+
 if [[ "$TEST_MODE" != "1" ]]; then
   PATH="$LINUX_SERVICE_PATH"
   export PATH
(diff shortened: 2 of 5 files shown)


deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -0,0 +1,114 @@
...
+
+run_installer uninstall
+[[ ! -e /opt/fleetnode ]] || fail "uninstall retained the program"
+[[ ! -e /etc/systemd/system/fleet-node.service ]] || fail "uninstall retained the unit"
+[[ -e /var/lib/fleetnode/state.yaml ]] || fail "uninstall removed state"
+getent passwd fleetnode >/dev/null || fail "uninstall removed the service account"
+
+run_installer "$VERSION"
+sudo systemctl is-active --quiet fleet-node.service && fail "reinstall started the service"
+run_installer uninstall
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Add the intended purge path and checks, or remove that claim from the PR description.

Not addressing: the purge claim has been removed from the current PR description, and the native qualifier now ends with one state-preserving uninstall. The reviewed line is outdated.

deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -0,0 +1,106 @@
...
+tar -xzf "$PACKAGE_DIR/$ARCHIVE" -C "$WORK_DIR"
+printf '#!/usr/bin/env bash\nexit 1\n' > "$WORK_DIR/${ARCHIVE%.tar.gz}/plugins/antminer-plugin"
+(
+  cd "$WORK_DIR"
+  tar -czf "$ARCHIVE" "${ARCHIVE%.tar.gz}"
+  sha256sum "$ARCHIVE" > "$ARCHIVE.sha256"
+)
+if DOWNLOAD_URL="file://$WORK_DIR" run_installer "$VERSION"; then
+  fail "upgrade accepted a plugin that exited during startup"
+fi
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Run a successful same-version upgrade from the original archive before injecting the broken plugin.

Addressed in `a221e7a01`: the native qualifier now completes a known-good same-version upgrade and confirms the service remains active before testing the deliberately broken plugin upgrade.

deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -0,0 +1,109 @@
...
+sudo systemctl is-active --quiet fleet-node.service || fail "failed upgrade did not restore the running service"
+restored_plugin_hash=$(sha256sum /opt/fleetnode/plugins/antminer-plugin | awk '{print $1}')
+[[ "$restored_plugin_hash" == "$original_plugin_hash" ]] || fail "failed upgrade did not restore the previous payload"
+rm -rf "$WORK_DIR"
+WORK_DIR=""
+
+run_installer uninstall
+[[ ! -e /opt/fleetnode ]] || fail "uninstall retained the program"
+[[ ! -e /etc/systemd/system/fleet-node.service ]] || fail "uninstall retained the unit"
+[[ -e /var/lib/fleetnode/state.yaml ]] || fail "uninstall removed state"
```
rl-block: This `-e` test runs as the unprivileged runner user, but `/var/lib/fleetnode` is created `0700 fleetnode:fleetnode` (`install-fleet-node.sh:536`) and uninstall leaves it in place. `stat()` on a path inside a directory we can't traverse returns EACCES, so `-e` is false and this line fails with "uninstall removed state" even though the state was preserved. As written I think the gate goes red on its first nightly run.

Maybe `sudo test -f /var/lib/fleetnode/state.yaml || fail "uninstall removed state"`? (Or `sudo runuser -u fleetnode -- test -r ...` if we also want to prove the service account can still read it.)

.github/workflows/proto-fleet-artifact-build.yml
```diff
@@ -304,15 +305,20 @@ jobs:
...
           sha256sum "$archive" > "$archive.sha256"
-          sha256sum -c "$archive.sha256"
+
+      - name: Qualify Fleet Node package on systemd
+        env:
+          PACKAGE_DIR: ${{ github.workspace }}/server
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes nmap
+          deployment-files/fleetnode/tests/qualify-package.sh "$VERSION" "$PACKAGE_DIR"
```
rl-block, line 321: Since this workflow is `workflow_call`-only (nightly + tag) and PR CI only runs `bash -n` on the qualifier, the gate has never actually executed on this branch; `gh run list` for `nightly-builds.yml` / `release.yml` on `ankitg/fleetnode-package-qualification` is empty. Given the `-e` issue in `qualify-package.sh:105`, the first run would probably fail.

Would it be worth dispatching `nightly-builds.yml` against this branch before merge (`gh workflow run nightly-builds.yml --ref ankitg/fleetnode-package-qualification`) and linking the green amd64/arm64 jobs here? Longer term we could consider a PR-path job in `deployment-config-checks.yml` that builds, packages, and runs `qualify-package.sh` whenever `deployment-files/fleetnode/**` or this workflow changes.

deployment-files/fleetnode/install-fleet-node.sh
```diff
@@ -51,11 +56,16 @@ case "${1:-}" in
...
     ;;
 esac
-if [[ "$TEST_MODE" != "1" ]]; then
+if [[ "$TEST_MODE" == "0" ]]; then
   if [[ -n "$ROOT_PREFIX" || -n "${FLEETNODE_ARCH:-}" || -n "$DOWNLOAD_BASE_URL" || -n "${FLEETNODE_SYSTEMCTL:-}" ]]; then
     echo "Fleet Node installer overrides are restricted to automated tests" >&2
     exit 1
   fi
+elif [[ "$TEST_MODE" == "systemd" ]]; then
+  if [[ -n "$ROOT_PREFIX" || -n "${FLEETNODE_ARCH:-}" || -n "${FLEETNODE_SYSTEMCTL:-}" || "$DOWNLOAD_BASE_URL" != file://* ]]; then
```
rl-block: The PR body's "Installer tests" row mentions qualification-mode guards, but `test-install-fleetnode.sh` at this head has no `FLEETNODE_TEST_MODE=systemd` case (the only mode it references is `=1`), so the rejections here and the `invalid FLEETNODE_TEST_MODE` case at :14-17 are untested.

These guards exit before any `sudo`/`systemctl` call, so they should slot in next to the existing override test (~:245-249) without new fakes: `FLEETNODE_TEST_MODE=systemd` with an `https://` URL; with a `file://` URL plus each of `FLEETNODE_ROOT_PREFIX` / `FLEETNODE_ARCH` / `FLEETNODE_SYSTEMCTL`; and `FLEETNODE_TEST_MODE=bogus`, each asserting the exact stderr text. Alternatively we could just reword the PR body row to match what the diff covers.

deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -0,0 +1,109 @@
...
+
+WORK_DIR=$(mktemp -d)
+tar -xzf "$PACKAGE_DIR/$ARCHIVE" -C "$WORK_DIR"
+printf '#!/usr/bin/env bash\nexit 1\n' > "$WORK_DIR/${ARCHIVE%.tar.gz}/plugins/antminer-plugin"
+(
+  cd "$WORK_DIR"
+  tar -czf "$ARCHIVE" "${ARCHIVE%.tar.gz}"
+  sha256sum "$ARCHIVE" > "$ARCHIVE.sha256"
+)
+if DOWNLOAD_URL="file://$WORK_DIR" run_installer "$VERSION"; then
```
rl-block: This asserts only that the installer exited non-zero plus post-state (`is-active`, one plugin hash). All of the installer's pre-flight rejections (:423-475) exit before the `systemctl stop` at :508 that arms rollback, so if the tampered archive were ever rejected early, every assertion here would still hold and the rollback path itself would go unproven.

One cheap discriminator: capture `invocation_before=$(sudo systemctl show -p InvocationID --value fleet-node.service)` before this run and assert it changed afterwards, which proves stop -> candidate start -> rollback start actually happened. The same check around the known-good upgrade at :82-83 would also show that a same-version upgrade really restarted the daemon rather than being indistinguishable from a no-op.

.github/workflows/proto-fleet-artifact-build.yml
```diff
@@ -304,15 +305,20 @@ jobs:
...
 
           tar -czf "$archive" "$package_dir"
           sha256sum "$archive" > "$archive.sha256"
-          sha256sum -c "$archive.sha256"
+
+      - name: Qualify Fleet Node package on systemd
+        env:
+          PACKAGE_DIR: ${{ github.workspace }}/server
+        run: |
+          sudo apt-get update
```
rl-block: nit: these two `apt-get` calls sit on the release/nightly path with no retry, so a transient mirror hiccup fails the artifact build. Perhaps `-o Acquire::Retries=3` on both?

deployment-files/fleetnode/tests/qualify-package.sh
```diff
@@ -0,0 +1,109 @@
...
+trap cleanup EXIT
+
+for path in /opt/fleetnode /etc/fleetnode /var/lib/fleetnode /etc/systemd/system/fleet-node.service; do
+  [[ ! -e "$path" ]] || fail "runner is not clean: $path already exists"
+done
+getent passwd fleetnode >/dev/null 2>&1 && fail "runner already has a fleetnode account"
+
+CLEANUP_ALLOWED=1
+run_installer "$VERSION"
+sudo systemctl is-active --quiet fleet-node.service && fail "fresh install started the unenrolled service"
```
rl-block: nit: `is-active --quiet` is also non-zero for `failed` and `activating (auto-restart)`, so this passes for a unit that was started and died (with no state file `fleetnode run` exits 78 -> `failed`) exactly like one that was never started. If the intent is "never started", maybe `[[ "$(sudo systemctl show -p ActiveState --value fleet-node.service)" == inactive ]]`?

mcharles-square (approved):

rl-block (approved):

deployment-files/fleetnode/tests/qualify-package.sh
ankitgoswami: 🤖

> The unprivileged `-e` test cannot traverse the `0700` state directory and will falsely fail; use a sudo test.

Addressed in `b5cb03aab`: the qualifier now checks preserved state with `sudo test -f /var/lib/fleetnode/state.yaml`.

.github/workflows/proto-fleet-artifact-build.yml
ankitgoswami: 🤖

> Would it be worth dispatching `nightly-builds.yml` against this branch before merge? Longer term we could consider a PR-path job.

Not changing CI for this: `nightly-builds.yml` deliberately rejects non-default branches before artifact jobs run, then publishes the resulting nightly release. Dispatching it against this branch would therefore fail at the guard; weakening that guard would couple branch qualification to the publishing workflow. A second path-triggered full artifact build would duplicate the existing nightly/tag gate and its cost. Keeping this YAGNI: the concrete privilege bug is fixed in `b5cb03aab`, and native qualification stays on the existing nightly/tag paths.

PR #995, opened 2026-09-01
handle Fleet Node outages as miner unavailability

ankitgoswami:
Reviewable diff: +396/-83 across 22 files (excludes generated, test, and story files).

## Summary

Miners assigned to a previously connected but unreachable Fleet Node now show **Unavailable** with **No connection** instead of presenting stale healthy data. The same outage classification is used by fleet lists, filters, counts, rack slots, CSV exports, and persisted uptime history. Standalone and HA deployments receive a critical Fleet Node outage alert using five non-secret database columns.

## How it works

1. Fleet Nodes send heartbeats at most 30 seconds apart. A confirmed node is considered unavailable after its last heartbeat is more than two minutes old; a node that has never connected remains **Never connected** and does not trigger outage behavior.
2. Miner reads derive an effective `OFFLINE` status for devices assigned to that node without changing the stored miner status. List, count, filter, selector, rack-slot, CSV, and historical-snapshot paths apply the same rule.
3. The API returns the existing `OFFLINE` status plus `DEVICE_OFFLINE_REASON_FLEET_NODE_UNAVAILABLE`. The client maps that reason to `Unavailable` and `No connection` while retaining existing offline action and measurement behavior.
4. Both standalone and HA Grafana rule files query `fleet_node` through a five-column `SELECT` grant. After the node has been unavailable for another three minutes, Grafana fires a generic critical alert without exposing node names or key material. In HA, both local Grafanas evaluate the rule, while only the active Fleet API accepts webhook delivery.
5. When heartbeats resume, the derived reason disappears, miners return to their last stored status, and Grafana resolves the alert.

```mermaid
flowchart LR
    FN["Fleet Node heartbeat"] --> STATUS["Effective miner status"]
    STATUS --> API["OFFLINE plus reason enum"]
    API --> UI["Unavailable and No connection"]
    API --> VIEWS["Lists, racks, CSV, counts, and selectors"]
    STATUS --> HISTORY["Persisted uptime snapshot"]
    FN --> RULES["Standalone and HA Grafana rules"]
    RULES --> ALERT["Generic outage alert and recovery"]
```

```mermaid
stateDiagram-v2
    [*] --> NeverConnected
    NeverConnected --> Available: first heartbeat
    Available --> Unavailable: last heartbeat older than two minutes
    Unavailable --> Alerting: unavailable for three more minutes
    Unavailable --> Available: heartbeat resumes
    Alerting --> Available: heartbeat resumes and alert resolves
```

## Areas of the code involved

| --- | --- | --- |
| `proto/fleetmanagement/` | Adds `DeviceOfflineReason` and `MinerStateSnapshot.offline_reason` | Review the additive API contract and reason semantics |
| `server/internal/domain/stores/sqlstores/` and `server/sqlc/queries/` | Derives Fleet Node-aware effective status across reads, filters, counts, selectors, rack slots, and persisted snapshots | Review consistency, first-heartbeat semantics, and recovery behavior |
| `server/internal/domain/fleetmanagement/` | Maps the derived reason into API snapshots and CSV copy | Review the boundary between stored and derived state |
| `client/src/protoFleet/features/` | Renders `Unavailable`, `No connection`, and the matching two-minute Nodes-page threshold | Review operator-visible behavior |
| `server/migrations/000145_*` and `deployment-files/run-fleet.sh` | Grants Grafana only `org_id`, `id`, `last_seen_at`, `enrollment_status`, and `deleted_at` | Review standalone and HA least-privilege provisioning |
| `server/monitoring/grafana/` and `server/internal/domain/alerts/` | Provisions the outage rule for standalone and HA and adds generic firing/recovery copy | Review the five-minute alert lifecycle and HA delivery behavior |
| `server/internal/handlers/alerts/` | Treats the generic Fleet Node template as safe for alert-only readers | Review the permission boundary |
| `server/cmd/fleetnode/` | Enforces a maximum 30-second heartbeat interval | Review compatibility for custom launch flags |
| `**/generated/**`, `*.pb.go`, `*.pb.ts` | Regenerated protobuf and sqlc output | Generated — skip |

- Keep `DeviceStatus.OFFLINE` and add a reason enum instead of creating a Fleet Node-specific status, preserving existing offline behavior while allowing precise copy.
- Derive outages at read and snapshot time instead of persisting a replacement device status, so recovery restores the miner's last reported state without a repair write.
- Reuse the existing effective-status fragment in dynamic Go SQL, while keeping static sqlc and Grafana predicates explicit instead of introducing a database function or configuration abstraction.
- Require a real `last_seen_at`; confirmation alone does not turn a never-connected node into an outage.
- Grant Grafana five lifecycle columns instead of table access or a new database view/function; names and key material remain denied.
- Provision the rule directly in both standalone and HA rule files rather than mounting the full standalone rule set into HA.
- Fail fast when a custom heartbeat interval exceeds 30 seconds, keeping the two-minute outage threshold bounded instead of adding clamping or another runtime setting.
- Apply the migration grant only when the HA Grafana role already exists, and repeat the grant in `run-fleet.sh` for roles created after migrations.

## Testing & validation

- `just gen` is stable after regenerating protobuf and sqlc output.
- `just lint` passes: buf, ESLint, and golangci-lint report no issues.
- Focused Go tests pass for CSV copy, Fleet Node rule provisioning, exact grant/revoke columns, dynamic/static staleness SQL, and heartbeat validation.
- `bash -n deployment-files/run-fleet.sh` and `git diff --check` pass.
- The existing database integration test now covers live outage/recovery, persisted offline history, rack offline counts, and rack-slot status, but was not run locally; CI owns the database-backed validation.
- No live Grafana firing/recovery cycle, role-level privilege integration test, or production-scale query-plan benchmark was run.

- **Window / owner:** The deployer or Fleet on-call watches the first 30 minutes after rollout and the first controlled Fleet Node outage/recovery in standalone and HA.
- **Logs:** Search Fleet API and Grafana logs for `protofleet-fleet-node-unavailable`, `permission denied for table fleet_node`, query errors, and alert evaluation errors.
- **Expected behavior:** A node that has heartbeated should move its miners to Unavailable after two minutes and fire one alert after five minutes total. A confirmed node that has never heartbeated should remain Never connected without an outage alert.
- **Healthy recovery:** After the node resumes heartbeats, live views and subsequent history snapshots return to the stored miner status and Grafana resolves the alert.
- **Failure / mitigation trigger:** Roll back or disable the rule if healthy nodes are marked unavailable, counts disagree with rack/list/history views, HA duplicates delivered notifications, Grafana reports permission errors, the alert flaps, or recovery does not clear the reason.


server/internal/domain/stores/sqlstores/device.go
```diff
@@ -1496,7 +1505,12 @@ JOIN device_set dc ON dcm.device_set_id = dc.id
 JOIN device d ON dcm.device_id = d.id
 JOIN discovered_device dd ON d.discovered_device_id = dd.id
 JOIN device_pairing dp ON d.id = dp.device_id
-LEFT JOIN device_status ds ON d.id = ds.device_id
+LEFT JOIN device_status ON d.id = device_status.device_id
+LEFT JOIN fleet_node_device fleet_node_assignment ON fleet_node_assignment.device_id = d.id
+    AND fleet_node_assignment.org_id = d.org_id
+LEFT JOIN fleet_node assigned_fleet_node ON assigned_fleet_node.id = fleet_node_assignment.fleet_node_id
+    AND assigned_fleet_node.org_id = fleet_node_assignment.org_id
+LEFT JOIN LATERAL (SELECT `+effectiveDeviceStatusExpr+` AS status) effective_status ON TRUE
```
rl-block, line 1513: Now that the rack counts derive the effective status here, `GetRackSlotStatuses` in `collection.go` (~L1009, still `ds.status AS device_status`) will disagree with them inside the same `GetCollectionStats` response during a node outage: the rack summary counts the miner as offline while its grid slot stays HEALTHY, so the affected miners can't be picked out of the rack view.

Should we give the `slot_devices` CTE the same two `fleet_node` joins plus `effectiveDeviceStatusExpr`? A DB assertion that a stale-node ACTIVE miner yields slot status OFFLINE (4) and `OfflineCount == 1` for the same rack would pin the parity. (+1 to the Codex note; still applies at 8e75ce9.)

server/sqlc/queries/device.sql
```diff
@@ -376,6 +376,18 @@ FROM device d
...
+LEFT JOIN LATERAL (
+    SELECT CASE
+        WHEN fn.id IS NOT NULL
+             AND fn.deleted_at IS NULL
+             AND fn.enrollment_status = 'CONFIRMED'
+             AND COALESCE(fn.last_seen_at, fn.updated_at, fn.created_at) < NOW() - INTERVAL '2 minutes'
+        THEN 'OFFLINE'
+        ELSE ds.status::text
+    END AS status
+) effective_status ON TRUE
```
rl-block, line 391: `InsertMinerStateSnapshot` (`miner_state_snapshots.sql:15-31`) says its CASE must match this query, but it still reads `ds.status`. During a node outage the live legend now shows these miners offline while the persisted history records them as hashing, and that history can't be repaired afterwards.

Maybe worth adding the same joins + LATERAL there and switching its CASE to `effective_status.status` (then `just gen`)? Extending `TestFleetNodeOutageDerivesOfflineStatusAndRecovers` to insert a snapshot mid-outage and expect state 0 would cover it. (+1 to the Codex comment; still open at head.)

server/migrations/000145_grant_fleet_node_alert_access.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'grafana_ha_ro') THEN
+        GRANT SELECT (
+            org_id,
+            id,
+            last_seen_at,
+            updated_at,
+            created_at,
+            enrollment_status,
+            deleted_at
+        ) ON fleet_node TO grafana_ha_ro;
```
rl-block, line 12: Heads-up on the HA side: `deployment-files/ha/fleet-compose.yaml` mounts only `server/monitoring/grafana/ha/proto-fleet-ha-rules.yaml` for alerting (and `ha/tests/test-profile.sh` asserts `proto-fleet-rules.yaml` is *not* mounted), and that file carries only the readiness rule. As written, HA installs get this grant but never the Fleet Node Unavailable rule.

Do we want to add the rule block to the HA rules file (and have `fleet_node_rule_test.go` read that file too), or is HA meant to stay readiness-only? In the latter case this migration could probably go, leaving the `run-fleet.sh` grant for standalone.

server/internal/domain/fleetmanagement/service.go
```diff
@@ -855,7 +855,10 @@ func (s *Service) buildSnapshotsFromUnifiedQuery(
 			if row.SerialNumber.Valid {
 				snapshot.SerialNumber = row.SerialNumber.String
 			}
-			if row.DeviceStatus.Valid {
+			if row.FleetNodeUnavailable {
+				snapshot.DeviceStatus = pb.DeviceStatus_DEVICE_STATUS_OFFLINE
+				snapshot.OfflineReason = pb.DeviceOfflineReason_DEVICE_OFFLINE_REASON_FLEET_NODE_UNAVAILABLE
```
rl-block, line 860: Small consistency thing: the CSV export (`export_csv.go`, `minerStatusCSVValue` / `minerIssuesCSVValue`) consumes these same snapshots but only switches on `DeviceStatus`, so an exported row for one of these miners reads `Offline` with an empty Issues cell while the list shows `Unavailable` / `No connection`.

If the CSV is meant to mirror the list copy, we could add an `OfflineReason == FLEET_NODE_UNAVAILABLE` guard at the top of both helpers plus a case in `export_csv_test.go`.

server/internal/infrastructure/timescaledb/fleet_node_rule_test.go
```diff
@@ -0,0 +1,82 @@
...
+	runFleet, err := os.ReadFile("../../../../deployment-files/run-fleet.sh")
+	require.NoError(t, err)
+
+	for _, sql := range []string{string(upSQL), string(runFleet)} {
+		require.Contains(t, sql, "GRANT SELECT (")
+		for _, column := range []string{
+			"org_id", "id", "last_seen_at", "updated_at", "created_at", "enrollment_status", "deleted_at",
+		} {
+			require.Contains(t, sql, column)
+		}
```
rl-block: These `Contains` checks run against the whole file, and `run-fleet.sh` also names all seven columns in its smoke `SELECT` (tokens like `id` appear throughout the script anyway), so dropping a column from the `GRANT` wouldn't turn the test red.

Perhaps extract the `GRANT SELECT (...) ON fleet_node` column list with a regex and `ElementsMatch` it against the expected seven (and do the same for the `REVOKE` list in the down migration)?

server/internal/domain/stores/sqlstores/device_query_fragments.go
```diff
@@ -24,12 +24,24 @@ const embeddedWebViewAvailableExpr = `(
...
+    AND fleet_node_assignment.device_id IS NULL
 )`
 
+// Four missed 30-second heartbeats marks a Fleet Node unavailable.
+const fleetNodeUnavailableExpr = `(
+    assigned_fleet_node.id IS NOT NULL
+    AND assigned_fleet_node.deleted_at IS NULL
+    AND assigned_fleet_node.enrollment_status = 'CONFIRMED'
+    AND COALESCE(assigned_fleet_node.last_seen_at, assigned_fleet_node.updated_at, assigned_fleet_node.created_at) < NOW() - INTERVAL '2 minutes'
+)`
```
rl-block, line 37: The staleness predicate now lives here plus six literal copies in `device.sql` and one in the Grafana rule. Fine for this PR, but a future threshold tweak has to land in all of them or list/count/alert classification drifts -- which is essentially what happened with the snapshot and rack-slot queries.

Would a small SQL function in a migration (e.g. `fleet_node_is_unavailable(fleet_node)`) that both sqlc and this constant call be worth it? If not, maybe a comment here listing the sibling copies so the next person finds them.

server/monitoring/grafana/provisioning/alerting/proto-fleet-rules.yaml
```diff
@@ -65,6 +65,47 @@ groups:
...
+              format: table
+              rawSql: |
+                SELECT
+                    org_id::text AS organization_id,
+                    id::text AS fleet_node_id,
+                    1 AS value
+                FROM fleet_node
+                WHERE enrollment_status = 'CONFIRMED'
+                  AND deleted_at IS NULL
+                  AND COALESCE(last_seen_at, updated_at, created_at) < NOW() - INTERVAL '2 minutes'
```
rl-block: Because of the `COALESCE` fallback (same in `fleetNodeUnavailableExpr`), a CONFIRMED node that has never sent a heartbeat flips its miners to Unavailable two minutes after confirmation and pages critical three minutes later, while the Nodes page badge shows it as "Never connected".

Is that the intended behaviour for the enroll -> confirm -> start-daemon flow? If we'd rather wait for first contact, `last_seen_at IS NOT NULL AND last_seen_at < ...` in both places would do it.

server/cmd/fleetnode/run.go
```diff
@@ -138,6 +138,19 @@ func (r *RunCmd) run(c *Context, logOutput io.Writer) error {
...
+func (r *RunCmd) validateHeartbeatInterval() error {
+	if r.HeartbeatInterval == 0 {
+		r.HeartbeatInterval = defaultHeartbeatInterval
+	}
+	if r.HeartbeatInterval < 0 {
+		return fmt.Errorf("heartbeat interval must be positive")
+	}
+	if r.HeartbeatInterval > defaultHeartbeatInterval {
+		return fmt.Errorf("heartbeat interval must be no greater than %s", defaultHeartbeatInterval)
+	}
```
rl-block, line 150: Previously a non-positive interval fell back to the default; now anything above 30s is a hard error, so a node started with a custom `--heartbeat-interval 60s` from an older unit file would fail to start after upgrading.

Do we want to clamp to 30s with a warning instead, or is fail-fast intended here (in which case maybe a release-note line)?

rl-block (approved):

server/sqlc/queries/device.sql
```diff
@@ -338,26 +338,26 @@ WHERE d.deleted_at IS NULL
 SELECT
     -- Offline
     COALESCE(SUM(CASE
-        WHEN ds.status = 'OFFLINE'
-             OR (ds.status IS NULL AND dp.pairing_status NOT IN ('AUTHENTICATION_NEEDED'))
+        WHEN effective_status.status = 'OFFLINE'
+             OR (effective_status.status IS NULL AND dp.pairing_status NOT IN ('AUTHENTICATION_NEEDED'))
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> Update the snapshot insertion classifier to use the same Fleet Node-aware effective status.

Addressed in `14f7f68d2`: snapshot insertion now uses the effective status, and the existing outage integration test asserts that the persisted history bucket is offline rather than hashing.

server/internal/domain/stores/sqlstores/device.go
```diff
@@ -1464,29 +1473,29 @@ func (s *SQLDeviceStore) GetMinerStateCountsByCollections(ctx context.Context, o
 	query := fmt.Sprintf(`SELECT dcm.device_set_id,
     -- Offline
     COALESCE(SUM(CASE
-        WHEN ds.status = 'OFFLINE'
-             OR (ds.status IS NULL AND dp.pairing_status NOT IN ('AUTHENTICATION_NEEDED'))
+        WHEN effective_status.status = 'OFFLINE'
+             OR (effective_status.status IS NULL AND dp.pairing_status NOT IN ('AUTHENTICATION_NEEDED'))
```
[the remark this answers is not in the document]
ankitgoswami: 🤖

> use the same effective-status expression in the slot query.

Addressed in `14f7f68d2`: rack slots now reuse `effectiveDeviceStatusExpr`, and the existing outage integration test pins both the rack offline count and the slot's `OFFLINE` status.

server/internal/domain/stores/sqlstores/device.go
ankitgoswami: 🤖

> Should we give the `slot_devices` CTE the same two `fleet_node` joins plus `effectiveDeviceStatusExpr`?

Addressed in `14f7f68d2`: `slot_devices` now joins the assigned Fleet Node and reuses `effectiveDeviceStatusExpr`. The existing outage integration test verifies `OfflineCount == 1` and an `OFFLINE` slot for the same rack.

server/sqlc/queries/device.sql
ankitgoswami: 🤖

> switching its CASE to `effective_status.status` ... and expect state 0 would cover it.

Addressed in `14f7f68d2`: `InsertMinerStateSnapshot` now classifies with `effective_status.status`, generated sqlc is updated, and the existing outage integration test verifies an offline persisted snapshot.

server/migrations/000145_grant_fleet_node_alert_access.up.sql
ankitgoswami: 🤖

> HA installs get this grant but never the Fleet Node Unavailable rule.

Addressed in `14f7f68d2`: the HA-specific rule file now provisions Fleet Node Unavailable without mounting the standalone default-rule file. The semantic rule test validates both standalone and HA provisioning.

server/internal/domain/fleetmanagement/service.go
ankitgoswami: 🤖

> an exported row for one of these miners reads `Offline` with an empty Issues cell while the list shows `Unavailable` / `No connection`.

Addressed in `14f7f68d2`: CSV status and issue helpers now map the Fleet Node unavailable reason to `Unavailable` and `No connection`, with focused table-test coverage.

server/internal/infrastructure/timescaledb/fleet_node_rule_test.go
ankitgoswami: 🤖

> extract the `GRANT SELECT (...) ON fleet_node` column list with a regex and `ElementsMatch` it ... and do the same for the `REVOKE` list.

Addressed in `14f7f68d2`: the test now extracts each Fleet Node grant/revoke column list and checks the exact five-column set. The tests remain outside the migration folder.

server/internal/domain/stores/sqlstores/device_query_fragments.go
ankitgoswami: 🤖

> Would a small SQL function in a migration ... be worth it? If not, maybe a comment here listing the sibling copies.

Declined on YAGNI grounds: a database function or copy inventory would add another abstraction and synchronization surface. Dynamic Go SQL reuses the existing effective-status fragment; static sqlc and Grafana predicates stay explicit and are covered by source-level contract tests. The missing snapshot and rack consumers were fixed directly.

server/monitoring/grafana/provisioning/alerting/proto-fleet-rules.yaml
ankitgoswami: 🤖

> If we'd rather wait for first contact, `last_seen_at IS NOT NULL AND last_seen_at < ...` in both places would do it.

Addressed in `14f7f68d2`: outage detection now requires a real, stale `last_seen_at` everywhere. Never-connected nodes remain in that state without paging, and removing the timestamp fallback reduced Grafana's grant from seven columns to five.

server/cmd/fleetnode/run.go
ankitgoswami: 🤖

> Do we want to clamp to 30s with a warning instead, or is fail-fast intended here?

Fail-fast is intentional: the two-minute availability contract depends on heartbeats being no more than 30 seconds apart. Keeping the explicit validation avoids silent clamping, warning behavior, and another runtime policy; the constraint is documented in the flag help and PR description.

ankitgoswami:
🤖

> Persisted uptime ignores Fleet Node outages
>
> Rack slots remain healthy while rack counts report offline

Addressed in `14f7f68d2`: persisted snapshots now use effective Fleet Node status, and rack slots reuse the same effective-status expression. The existing outage integration test now covers persisted offline history, rack offline counts, and the corresponding offline slot.

PR #1008, opened 2026-09-03
feat(rollout): firmware release channels server foundation

rl-block:
## Summary

First of three stacked PRs porting the firmware rollout prototype to production, per the release channels design note added in this PR (`docs/plans/2026-09-03-firmware-release-channels-design.md`). Server only: the `RolloutService` API and enforcement loop land here; the UI follows in the next two PRs.

- **Release channels** replace the prototype's explicit "lanes": a scope of sites / buildings / racks / groups / individual miners resolved live against fleet placement (`release_channel_match` / `release_channel_member` views, same placement rules as `fleet_device_placement`), one firmware assignment per model, and an update behavior copied onto each rollout at start. Overlapping scopes are rejected on save under an org-level advisory lock; runtime overlaps (a miner moved later) resolve to the most specific selector and are flagged.
- **Update behavior** follows the UX design in #881: method `ALL_AT_ONCE` / `BATCHED` / `PILOT_THEN_CONTINUE`, order (least efficient first / random), batch and pilot size, wait between batches, review after each batch, auto-continue with hashrate / efficiency / temperature / new-error thresholds and a stabilization period, max miners offline at once.
- **Failure and cancel semantics**: a miner that does not verify after 3 update commands (~30 min) is `FAILED`; rollouts finish as `COMPLETED_WITH_FAILURES` instead of blocking. `CancelRollout` stops remaining work and keeps updated miners; `RollbackReleaseChannelFirmware` restores the previous version. Both use a per-device halt that drift correction honours until `RetryFailedRolloutDevices`.
- **Robustness**: every multi-statement operation runs in one transaction (`interfaces.Transactor`); stage transitions are compare-and-swap; the loop isolates errors per rollout; rollout target sets are snapshotted so history stays stable when membership changes.
- Wiring: `rollout-enforcement` runtime job (15s tick), `ActorRolloutEnforcement` session for dispatch attribution, `miner:firmware_update` permission on every RPC, migration `000146`.

## Test plan

- [x] `go test ./internal/domain/rollout/` — 17 DB-backed scenarios on `dbtest` (pilot/batch gating, wait between batches, auto-continue + every threshold, failure after 3 attempts + completed-with-failures + retry, cancel remaining not restarted by drift, rollback, scope resolution across site/building/rack/group/miner incl. overlap rejection, dynamic membership, late joiners, exclusion, runtime-overlap specificity, ordering, max concurrent offline, behavior validation)
- [x] `go test ./internal/handlers/rollout/` — every RPC gated on the permission; proto translation round-trips
- [x] `go test ./migrations/` — `000146` down/up round-trip; `internal/infrastructure/db` bridge tests now assert the latest embedded migration version instead of a hardcoded one
- [x] Full server suite, `golangci-lint run ./...`, `buf lint`, client `tsc --noEmit`
- [ ] Deploy to a dev fleet and run a pilot update end to end once PR 2/3 land the UI

## Follow-ups (deferred by design)

Scheduled start; pacing on the Fleet bulk "Update firmware" modal; caching the `ListRollouts` read path at fleet scale.


rl-block:
Superseded by the smaller stack #1013 → #1014 → #1015 → #1016 → #1017 → #1018 → #1019 → #1020 → #1021 → #1022 → #1023 (same code, split for review and rebased onto current main).

PR #1043, opened 2026-09-14
feat(maintenance): search tickets and inventory parts

mcharles-square:
Follows #973 (merged): builds on the collapsible miner search that landed there. Now targets `main` directly.

## Summary

Brings the collapsible list search from the miner list to the Maintenance **Queue** and **Inventory** tabs.

- **Client:** `MinerSearchInput` is generalized into `ListSearchInput` (`label`/`id` props, same 250 ms debounce, collapse-when-empty, and code-point cap). `MinerSearchInput` stays as a thin `"Search miners"` wrapper so miner pickers are unchanged.
- **Toolbars regrouped** as filters-left / actions-right, with the search as the last filter so expanding it grows into the gap before the actions and nothing else moves:
  - Queue: `[List/Board] [My tickets] [Add Filter] [search] … [Refresh] [Create ticket]`
  - Inventory: `[Low stock] [Site] [Type] [search] … [Refresh] [Add part] [Import CSV]`
- **Inventory (new):** `InventoryFilter.search_query` (max 255), matched case-insensitively against part name, type, manufacturer, part number, bin location, and site name. `CountInventoryParts` gains the same predicate (and the `site` join it needs) so the pager total agrees with the rows.
- **Tickets (fix):** the existing `search_query` didn't escape LIKE wildcards, so searching `%` or `_` matched every ticket. All ticket/stat queries now use `ESCAPE '\'`.
- **Server consolidation:** one `likeSearchPattern` helper (trim → escape → wrap → `NULL` when blank) now backs activity, miner, ticket, and inventory search; the sqlc queries take the prebuilt pattern (`search_pattern IS NULL OR col ILIKE search_pattern ESCAPE '\'`) instead of concatenating `'%' || … || '%'` per row. `nullStringFromSearch` and `minerSearchPattern` are gone.

**Not in this PR:** the History tab (completed tickets) uses a different request type with no search field; it needs its own proto/query work and is left as a follow-up.

## Test plan

- [x] `go test` — `sqlstores` (new: inventory search per field + List/Count agreement, literal `%`/`_`, blank query; ticket `%`/`_` literal across List/Count/Stats; `likeSearchPattern` table), `handlers/inventory`, `handlers/maintenance`, `handlers/activity`, `domain/inventory`, `domain/activity` — all green against Postgres
- [x] `golangci-lint`, `gofmt`, `buf lint` clean
- [x] vitest — maintenance feature, `MinerSearchInput`, `Search`, `MinerList` suites; new tests: Queue search reaches `setFilter` with `excludeCompleted` intact, Inventory search reaches `setFilter` alongside the other filters
- [x] `tsc --noEmit`, eslint, prettier clean
- [x] Browser (Vite → rebuilt Docker fleetd): Queue search for `fan 2` finds TK-0001, `zzz-nomatch` shows "No tickets"; Inventory search `bitmain` → 3 parts, `bitmain%` → none (literal `%`), `d-0` → the two D-bin parts
- [x] Browser layout: expanding the search moves no other control on either tab (Add Filter / Refresh / Create ticket / table header positions identical before and after); at 375 px the field wraps onto its own full-width line with no horizontal overflow


client/src/protoFleet/features/maintenance/components/TicketQueue/TicketQueue.tsx
```diff
@@ -147,6 +150,14 @@ const TicketQueue = ({ initialViewMode = "list" }: TicketQueueProps) => {
     setMyTicketsActive(next);
     applyFilters(filters, next);
   }, [applyFilters, filters, myTicketsActive, options.currentAssignee]);
+  const handleSearch = useCallback(
+    (query: string) => {
+      setSearchQuery(query);
+      applyFilters(filters, myTicketsActive, query);
+      setSelectedTicketIds([]);
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Agreed, the queue only reset the ids handed to the bulk-close modal, not the List's own selection. Fixed in f4eff9840: TicketQueue now owns the List selection (`customSelectedItems`/`customSetSelectedItems`), clears it synchronously from `onQueryInput` on the first keystroke, and also on every applied filter, "My tickets" toggle, view switch, and page change. Added a test that selects a row, types a search, and asserts the action bar is gone before the debounced filter is applied.

client/src/protoFleet/components/ListSearchInput.tsx
```diff
@@ -0,0 +1,152 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+
+import { Search as SearchIcon } from "@/shared/assets/icons";
+import Button, { sizes, variants } from "@/shared/components/Button";
+import Search from "@/shared/components/Search";
+
+const SEARCH_DEBOUNCE_MS = 250;
+
+const MAX_SEARCH_QUERY_CODE_POINTS = 255;
```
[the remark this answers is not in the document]
mcharles-square: 🤖 Fixed in f4eff9840 by aligning the server instead of the client: `TicketFilter.search_query` now has `max_len = 255`, the same as the inventory, miner, and activity search fields, so the shared input's 255 code-point cap matches every list search validator. Regenerated bindings are included.

mcharles-square:
🤖 Copilot's Escape finding was valid and is fixed in 70fcf3a6a. Input blurs itself on Escape before forwarding the key, so the collapse-on-blur check still saw the typed query and kept the field open while Search then emptied it. `Search` now exposes `onEscape`, fired after the clear, and `ListSearchInput` collapses on it. Regression tests added in both suites; the clear button's stay-open-for-refinement behaviour is unchanged.

client/src/protoFleet/features/maintenance/components/TicketInventory/InventoryTab.tsx
```diff
@@ -170,22 +164,46 @@ const InventoryTab = () => {
...
+        <ListSearchInput
+          id="inventory-search"
+          label="Search parts"
+          initialValue={searchQuery}
+          onQueryChange={(query) => {
+            setSearchQuery(query);
+            applyFilters(siteIds, types, lowStockOnly, query);
+          }}
+          collapsible
+        />
```
rl-block, line 181: On phone widths this toolbar still uses `flex-wrap`, but the expanded search has no `phone:w-full` wrapper. The inner `w-full` only fills its auto-sized flex item, so the field can share a row with filters instead of taking the promised full-width row.

Perhaps wrap it like the miner-list search (or expose a wrapper class) and add a phone-layout assertion?

proto/maintenance/v1/maintenance.proto
```diff
@@ -232,7 +232,7 @@ message TicketFilter {
     repeated string         group_labels = 6;
     optional int64          assignee_user_id = 7;
     bool                    urgent_only  = 8;
-    string                  search_query = 9 [(buf.validate.field).string.max_len = 256];
+    string                  search_query = 9 [(buf.validate.field).string.max_len = 255];
```
rl-block: This tightens a released request contract from 256 to 255 characters, so a non-UI caller that previously sent a valid 256-character query will now receive a validation error. We could probably keep the API bound at 256 and make the shared input limit configurable, or otherwise confirm that rejecting the formerly valid value is safe.

rl-block (approved):

client/src/protoFleet/features/maintenance/components/TicketInventory/InventoryTab.tsx
mcharles-square: 🤖 Confirmed at 375 px: the expanded field measured 231 px sharing a row with a filter pill. Fixed in 7bba919da: ListSearchInput takes a `className` applied to the expanded landmark only, and the Inventory tab passes `phone:w-full` so the field claims its own row. Covered by an assertion in InventoryTab.test.tsx.

proto/maintenance/v1/maintenance.proto
mcharles-square: 🤖 Agreed, 256 already shipped in v0.3.1-rc.5 and v0.3.2-alpha.3. Reverted in e035ba0d8: the proto and generated files are back to `max_len = 256`. The client-side cap of 255 code points stays, since a UI limit one character stricter than the API is harmless.

client/src/protoFleet/components/ListSearchInput.tsx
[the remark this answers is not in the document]
mcharles-square: 🤖 Follow-up: the 255 proto change was reverted in e035ba0d8 after review noted 256 already shipped in tagged releases. The client caps at 255 code points instead, which keeps the input at or below every server bound without changing the API contract.

mcharles-square:
🤖 Applied Copilot's suggestion: the `role="search"` landmark is now named `"<label> controls"` (31f82a6bd), distinct from the input's own label so neither resolves to two elements.

